PR #19, opened 2015-01-02
Document inline-list and add inline-list--divided modifier

davidrapson:
Minor thing but `inline-list` is used in quite a few places so thought it worth adding to the patterns. Also added an `inline-list--divided` modifier which adds divider bullets between each list item (this was previously being done manually)

screen shot 2015-01-02 at 17 02 29


wpf500:
:+1:

PR #23, opened 2015-01-05
Log who the exception affected if they are logged in

wpf500:
It would be useful to know which user was affected by the exception.

Ideally this would be associated with the log entry for the exception itself, but often it is triggered in a context where we do not know about the request. In `Global.onError` the exception has been reconciled with the request.


jennysivapalan:
:+1: :deciduous_tree: :leaves:

PR #24, opened 2015-01-05
Increase time between MC data refresh to preserve CAPI quota

wpf500:
We are currently hitting our API quota limit at around 9-10pm each day and there is no real need to refresh the masterclass data that often.


jennysivapalan:
:+1: :hourglass:

PR #25, opened 2015-01-05
Remove tierOpt from sign in page

davidrapson:
Remove tierOpt from sign in page


davidrapson:
// @wpf500

wpf500:
:+1:

wpf500:
Oh actually also just noticed can you remove the `Benefits` import

davidrapson:
Nice, done.

PR #26, opened 2015-01-06
Fix bug with booking button visited states

davidrapson:
Spotted a minor issue with the colour of booking buttons. Only happens when the link has been visited before. Styles were a little overzealous.

**Before**
screen shot 2015-01-06 at 10 10 34 1

**After**
screen shot 2015-01-06 at 10 10 45


wpf500:
:+1:

PR #27, opened 2015-01-06
Remove dead code

davidrapson:
Remove a fragment and a bunch of old styles and a that are no longer being used. Should make @wpf500 happy.


wpf500:
:+1:

PR #29, opened 2015-01-06
Move homepage slideshow images to The Grid

davidrapson:
- Move homepage slideshow images to The Grid
- Don't use imager for homepage feature boxes as it let's us remove the noAssetHash folder and process. Not enough benefit for the effort.
- Remove noAssetHash from Gruntfile

screen shot 2015-01-06 at 11 53 56


rtyley:
Yay! :fireworks: :+1: 

Almost makes me want to run the BFG again, but I'll refrain - we already cut 40% or so off the repo size at Christmas.

PR #30, opened 2015-01-06
Don't print sensitive details in the logs

wpf500:


rtyley:
Yep, :+1:

These detials being logged out are Zuora connection credentials, not personal-user information, but we still don't want this ever going to a log file. In the future, we could maybe look at automatically replacing any credential text found in the secure properties.

PR #31, opened 2015-01-06
AuthenticatedException to wrap user and exception

wpf500:
PR #23 introduced an extra log message to help associate a user with an exception. With a great suggestion from @rtyley, this introduces an `AuthenticatedException` which wraps the actual exception and contains details about the user.

One problem with this approach is that it suppresses the original stack trace on Play's default error page (see below), while this doesn't matter in production, it can be useful for devs. However the stack trace is still in the logs so we'll see how annoying this is for a while.

image

### DIFF
--- a/frontend/app/Global.scala
+++ b/frontend/app/Global.scala
@@ -1,4 +1,4 @@
-import play.api.{UsefulException, Logger, Application}
+import play.api.Application
 import play.api.mvc.Results.{InternalServerError, NotFound}
 import play.api.mvc.{RequestHeader, Result, WithFilters}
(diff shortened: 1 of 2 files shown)


rtyley:
LGTM - with Play 2.4, we might want to take advantage of the `DefaultHttpErrorHandler` to make this change less intrusive, and just override the logged message (where we have the Request available):

https://github.com/playframework/playframework/blob/518163/framework/src/play/src/main/java/play/http/DefaultHttpErrorHandler.java#L120-L123

https://www.playframework.com/documentation/2.4.0-M2/Migration24

PR #32, opened 2015-01-06
Remove rems from borders / rem(1px)

davidrapson:
Remove rems from borders / `border: rem(1px)` as it's not a value that should be remified


wpf500:
:+1:

davidrapson:
Deploy completed, confirmed on prod. Build took quite a while, prout still thinks it's overdue for some reason.

davidrapson:
There we go, prout caught up

PR #34, opened 2015-01-06
Simplify About/package feature  fragment

davidrapson:
Simplify About/package feature fragment.

- Use `leadin` from model
- Build button from Tier, rather than passing HTML in

// @wpf500


wpf500:
:+1:

PR #35, opened 2015-01-06
Remove dead showChooseDetails logic from benefits fragment

davidrapson:
Remove dead showChooseDetails logic from benefits fragment. Simplifying this mega fragment one piece at a time…


wpf500:
:+1:

PR #38, opened 2015-01-06
Refactor footer styles to use global layout

davidrapson:
Refactor footer styles to use global layout: This helps towards being able to control our max-widths in a single place so we can go to 1300px and also remove specific scoping of `gs-container`.

Removes `nav/brand-bar` as the only place we use it is in the footer and we're not using it in the same way as next-gen so we don't really need the styles.

// @feedmypixel


davidrapson:
@feedmypixel Mind giving this a once over when you've got a minute?

feedmypixel:
As another pull request the global footer could be split apart more is it me or does it feel like it has a few to many bem children?

:+1: good stuff nice to moving towards 1300px

davidrapson:
Yep, that's a good task for another PR we can probably reduce the nesting here.

PR #39, opened 2015-01-07
Increase the timeout for fetching Masterclasses

mattandrews:
We're seeing errors fetching Masterclasses locally as it looks like they added a bunch (there's ~140 now, but it was only ~110 or so a month or so back) and now it times out. We haven't seen this in prod yet but just in case, this ups the timeout to give them space to return.

cc @wpf500 / @jamesoram  / @rtyley


rtyley:
:+1:

I'll have a think bout how we improve this beauty.

mattandrews:
Okay to go out @jamesoram?

jamesoram:
:+1: yes

PR #40, opened 2015-01-07
Track eventbrite iframe impressions

mattandrews:
It turns out that the radiator app only shows when users are sent to EventBrite when they're already signed in – if a user signs up for Membership while trying to buy tickets, we don't run the code which redirects them to EventBrite – instead they see this screen:

screen shot 2015-01-07 at 11 58 59

This update adds the same CloudWatch tracking string (`user-sent-to-eventbrite`) to this use case, so our radiator totals will now include this flow too.

### DIFF
--- a/frontend/app/controllers/Joiner.scala
+++ b/frontend/app/controllers/Joiner.scala
@@ -138,6 +138,7 @@ trait Joiner extends Controller {
         eventId <- PreMembershipJoiningEventFromSessionExtractor.eventIdFrom(request)
         event <- EventbriteService.getBookableEvent(eventId)
       } yield {
+        EventbriteService.getService(event).wsMetrics.put("user-sent-to-eventbrite", 1)
         MemberService.createDiscountForMember(request.member, event).map { discountOpt =>
           (event, (Config.eventbriteApiIframeUrl ? ("eid" -> event.id) & ("discount" -> discountOpt.map(_.code))).toString)
         }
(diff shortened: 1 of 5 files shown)


mattandrews:
cc @jennysivapalan / @wpf500 (whose code this basically is)

frontend/app/controllers/Joiner.scala
```diff
@@ -138,6 +138,7 @@ trait Joiner extends Controller {
         eventId <- PreMembershipJoiningEventFromSessionExtractor.eventIdFrom(request)
         event <- EventbriteService.getBookableEvent(eventId)
       } yield {
+        EventbriteService.getService(event).wsMetrics.put("user-sent-to-eventbrite", 1)
```
jennysivapalan: Should this be stored under another metric (so we can tell the difference between user thats are sent to Eventbrite and users that can purchase via the iframe? We can aggregate them later.

mattandrews: Not a bad idea – a followup task I'm trying to have a go at now is tracking when a user signs up (which we do already) but also recording whether they signed up for a Guardian Live event, a Masterclass, or just a plain old member. 

The main compelling argument for using the same metric name is that both instances still require another click from the user (eg. the big green Order Now button) so there's not much functional difference between this experience and being on EventBrite.

mattandrews: okay @jennysivapalan – have renamed the metric, and also added the tier (to both eventbrite metrics) on @wpf500's advice. This means the radiator will stop showing new data once this goes out, but we can do a corresponding update there too. And also break down signups by tier/intent!

jennysivapalan: cool :+1:

jennysivapalan:
:+1: :ticket:

mattandrews:
Do you need to test this @jamesoram or are you happy for it to go out? Just tweaking some metric/tracking names...

jamesoram:
:+1: ship it

PR #41, opened 2015-01-07
Move Modernizr to lib

davidrapson:
- Moves Modernizr to `lib` as it's the only piece of thirdparty code in modules, and moving to `lib` means we can more easily exclude from jshint
- Did a fresh build of Modernizr from http://modernizr.com/download/#-flexbox-flexboxlegacy-cssclasses-testprop-testallprops-domprefixes as we'd edited the file to exclude from jshint


mattandrews:
LGTM :+1:

PR #42, opened 2015-01-07
Clean up scss _vars

davidrapson:
- Removed any variables that we are not using. Mostly copied from next-gen, we're a bit behind how they handle colours anyway so if we're not using them they should go
- Also cleaned up a couple of places that variables were being used accidentally (e.g., using `newsRed` instead of a dedicated error colour for form errors)


feedmypixel:
:+1:

PR #45, opened 2015-01-08
Escape ampersands in urls

mattandrews:
#### Before:
screen shot 2015-01-08 at 12 44 11

#### After:
screen shot 2015-01-08 at 12 44 17


mattandrews:
(replacing them with `&amp;` didn't do anything)

davidrapson:
:+1:

mattandrews:
Okay with you @jamesoram?

jamesoram:
it looks good, but are there any other special characters we need to encode? Apostrophes?

rrees:
Java lib [URLEncoder](http://docs.oracle.com/javase/7/docs/api/java/net/URLEncoder.html) does all URL encoding and the unicode encoding and should be trivial to call from Scala.

PR #46, opened 2015-01-08
Use the Grid instead of Events images tool

jennysivapalan:
Please note this PR is not ready to be merge until we've completed a few more steps (see below). However code is ready for review.

This PR removes the reliance on the Events images tool and uses the Grid instead. This only affects Guardian live events and not Masterclasses.

Given a image from the Grid is supplied as a comment `"<!-- main-image: https://media-tool/images/sdf8u8sdf898hnsdcvs89dc?crop=0_3_480_288 -->`in the Eventbrite description a will call to the Grid API is made to fetch the different crops of the image and store it in the Event.

Some notes:
- Current implementation will call the Grid for every event that has a main-image. We probably want to optimise this (as a separate PR) to only call the Grid if the image url has changed.

Before we can merge we need to
- ~~Decide how to append the description to the image (I need to chat to Ben W)~~(Done)
- ~~Upload the new property files~~ (Done)
- ~~Upload all the existing images to the Grid and update existing events with the correct tag. As agreed with @rtyley we don't want to run the Events image tool and Grid service in parallel~~(Done)


frontend/app/views/fragments/event/item.scala.html
```diff
@@ -7,7 +7,7 @@
     <div class="event-item__container">
         <div class="event-item__media">
             <a href="/event/@event.id" itemprop="image" content="@event.socialImgUrl">
-                <div class="delayed-image-load" data-src="@event.imgUrl" data-alt="@event.name.text" data-class="responsive-img"></div>
+                @fragments.event.image(event)
```
davidrapson: Nice, I've got some work about to go up that splits the this item fragment up into two versions so it's good to have this as a standalone fragment.

frontend/assets/javascripts/src/modules/images.js
```diff
@@ -1,16 +1,20 @@
...
-                availableWidths: guardian.membership.eventImages.widths,
-                availablePixelRatios: guardian.membership.eventImages.ratios,
-                lazyload: true,
-                lazyloadOffset: 100
+        var imagerImages = utilsHelper.toArray(document.querySelectorAll(IMAGE_LOADED_IMAGE));
+
+        if (imagerImages.length) {
+            imagerImages.map(function(image) {
+                var imageAvailableWidths = image.getAttribute('data-available-widths');
+                new Imager([image], {
```
davidrapson: Does this need to be an array for imager? Does it not let us pass `item` instead of `[item]`?

frontend/assets/javascripts/src/modules/images.js
```diff
@@ -1,16 +1,20 @@
...
+                    availableWidths: imageAvailableWidths ? imageAvailableWidths.split(',') : [],
+                    lazyload: true,
```
davidrapson: Can we put this into a data-attribute, so we can pass a lazyload boolean through to the event image fragment? Then we can opt-in to lazyloading?

frontend/assets/javascripts/src/modules/images.js
feedmypixel: Yeah it needs to be an array or sting (selector) otherwise it tries to treat it as  the opt object. @jennysivapalan it may be worth dropping a comment in here as it feels a bit wrong maping through the array to place it back into the array

frontend/assets/javascripts/src/modules/images.js
```diff
@@ -1,16 +1,20 @@
-define(['lib/bower-components/imager.js/Imager'], function(Imager) {
+define(['lib/bower-components/imager.js/Imager', 'src/utils/helper'], function(Imager, utilsHelper) {
 
-    var IMAGES_EVENTS = '.delayed-image-load';
+    var IMAGE_LOADED_IMAGE = '.imager-loaded-image';
```
davidrapson: Should probably be `.js-` prefixed class.

davidrapson:
I'll let Will or Roberto review the Scala side, but the approach for the JS feels really nice. Couple of comments but nothing major.

frontend/app/model/Eventbrite.scala
```diff
@@ -171,6 +185,9 @@ object Eventbrite {
     val imgUrl = data.flatMap(_.images.headOption).flatMap(_.file)
       .getOrElse(views.support.Asset.at("images/event-placeholder.gif"))
       .replace("http://static", "https://static-secure")
+
+    val availableWidths = ""
```
jennysivapalan: We can get available widths for Masterclasses too as CAPI provides this. I'll do this as a separate PR as Masterclasses doesn't use the Grid.

mattandrews:
So to actually add an image someone had to add `<!-- main-image: https://media-tool/images/sdf8u8sdf898hnsdcvs89dc?crop=0_3_480_288 -->` to the Eventbrite page? :dizzy_face:

jennysivapalan:
@mattandrews unfortunately yes. We are getting to the point where we are asking the team to put a few things in the html. I think we could make a lightweight admin tool to deal with some of these things. Maybe we could have a team chat about it. @rtyley?

wpf500:
@rtyley loves lightweight admin tools

mattandrews:
A lightweight admin tool you say...

frontend/app/model/Eventbrite.scala
```diff
@@ -147,18 +147,34 @@ object Eventbrite {
...
+
+    private val widths = image.fold(List.empty[Int])(_.assets.map(_.dimensions.width))
+
+    val availableWidths = widths.mkString(",")
+
+    val imageMetadata = image.map(_.metadata)
+
     val socialImgUrl = {
       // get the largest available image (for social media etc)
       // this mirrors the URL structure generated by Imager.js
```
wpf500: These comments aren't relevant any more

frontend/app/model/Eventbrite.scala
```diff
@@ -147,18 +147,34 @@ object Eventbrite {
...
       // so any changes there need to be reflected here
-      val largestWidth = Config.eventImageWidths.toArray.last
-      val largestRatio = Config.eventImageRatios.toArray.last
-      Config.eventImageUrlPath(event.id) + "/" + largestWidth + "-" + largestRatio + "x.jpg"
+      image.fold(Config.gridConfig.fallbackImageUrl) { img =>
+        val asset = img.assets.find(_.dimensions.width == widths.max)
+        asset.map(_.file).getOrElse(Config.gridConfig.fallbackImageUrl)
```
wpf500: I think we could remove the duplicate `Config.gridConfig.fallbackImageUrl` here
```
val socialImgUrl = image.flatMap(_.assets.find(_.dimensions.width == widths.max)).fold(Config.gridConfig.fallbackImageUrl)(_.file)
```

Or something like that

jennysivapalan: Nice - that worked so I'll change it.

jennysivapalan: Nice - that worked so I'll change it.

frontend/assets/javascripts/src/modules/images.js
jennysivapalan: Can we do that when we decide to opt-out of lazy-loading? Feels a bit strange to change it when nothing requires lazyload to be false?

frontend/assets/javascripts/src/modules/images.js
jennysivapalan: done!

frontend/assets/javascripts/src/modules/images.js
davidrapson: Am I right in thinking you using the image fragment on the event detail page? We don't need lazy loading there.

frontend/app/views/event/page.scala.html
```diff
@@ -107,7 +107,7 @@ <h1 class="event-masthead__name">@event.name.text</h1>
             case _: MasterclassEvent => {}
             case _ => {
                 <div class="event-content__image" itemprop="image" content="@event.socialImgUrl">
-                    <div class="delayed-image-load" data-src="@event.imgUrl" data-alt="Please upload an event image" data-class="responsive-img"></div>
```
davidrapson: Hah. Just realised this means that all the alt images for the event detail images have been "Please upload an event image"

jennysivapalan:
This is looking good on my local dev machine. @jamesoram would you be able to pull this branch and give the events flow a test?

jamesoram:
I'll take a look

jamesoram:
:+1:

jennysivapalan:
oooh - I'm going to merge!

mattandrews:
Here is a grid of thumbs-up:

| :+1: |  |  |
|-------|------|-----|
| :+1: | |  |
| :+1: | :+1: | :+1: |
| :+1: | :+1: | :+1: |
| :+1: | :+1: | :+1: |

PR #48, opened 2015-01-08
Add diagnostics for the ID-API IdentityServiceErrors we saw today

rtyley:
There was a burst of these at lunchtime today:

```
services.IdentityServiceError: Couldn't find user with ID 1280xxxx
```

The logged error implies that the user doesn't exist in Identity, but that is not the case. I think we're expecting there to be fields on this user that do not exist - this logging should make it clearer how the parsing fails.

### DIFF
--- a/frontend/app/services/IdentityService.scala
+++ b/frontend/app/services/IdentityService.scala
@@ -1,22 +1,20 @@
 package services
 
-import scala.concurrent.Future
-import scala.concurrent.ExecutionContext.Implicits.global
-
-import play.api.Logger
-import play.api.Play.current
-import play.api.libs.json._
-import play.api.libs.ws.{WS, WSResponse}
-
 import com.gu.membership.util.Timing
 import com.gu.membership.zuora.Address
-
 import configuration.Config
 import controllers.IdentityRequest
 import forms.MemberForm._
-import model.{IdMinimalUser, IdUser}
 import model.UserDeserializer._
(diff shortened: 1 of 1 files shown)


frontend/app/services/IdentityService.scala
```diff
@@ -109,9 +107,13 @@ trait IdentityApi {
 
   def get(endpoint: String, headers:List[(String, String)], parameters: List[(String, String)]) : Future[Option[IdUser]] = {
     Timing.record(IdentityApiMetrics, "get-user") {
-      WS.url(s"${Config.idApiUrl}/$endpoint").withHeaders(headers: _*).withQueryString(parameters: _*).withRequestTimeout(1000).get().map { response =>
+      val url = s"${Config.idApiUrl}/$endpoint"
+      WS.url(url).withHeaders(headers: _*).withQueryString(parameters: _*).withRequestTimeout(1000).get().map { response =>
+        println("AX"+response.body)
```
jennysivapalan: The println can be removed or logged instead?

rtyley:
Have removed the println- and guess what, the logging shows what I suspected...

https://app.getsentry.com/the-guardian/membership/group/49685617/

```
Id Api response on https://idapi.theguardian.com/user/1280xxxx : JsError(List((/statusFields,List(ValidationError(error.path.missing,WrappedArray())))))
```

...the user definitely is missing the `statusFields` - so #50 makes the field optional.

PR #50, opened 2015-01-08
Further to #48, here's a fix for a user missing their status-fields

rtyley:
Users like this really exist, and I've verified that test json reflects genuine responses from the Identity API.

See #48 and emails from the day on membership.dev.


jennysivapalan:
:+1: :mag:

rtyley:
Logging showed an example of a bad user with build 4123:

https://app.getsentry.com/the-guardian/membership/group/49685617/

...shouldn't see this error with any build after 4128.

PR #52, opened 2015-01-09
Remove need for isCurrent flag in slideshow

davidrapson:
Spotted an issue with the slideshow that was causing the second slide to flash in before showing the first one. This was due to the `isCurrent` flag being passed on the second item, we can actually remove the flag.


wpf500:
:+1:

PR #53, opened 2015-01-09
Fix iframe tracking

mattandrews:
This prevents duplicate tracking when people refresh the thank you page.

Note that we've changed the name of the metric (before it was `user-sent-to-eventbrite-$tier`) but the dashboard's broken anyway so this shouldn't be a huge deal.

Hopefully after this, the metrics for user signups and this new one (eg. signups for an event) will correspond.


mattandrews:
This okay @jamesoram / @wpf500 ?

wpf500:
I think you should revert the `user-sent-to-eventbrite-$tier` change as well in the `Event` controller

mattandrews:
Why?

mattandrews:
:question:

wpf500:
I dunno, ignore me

mattandrews:
obviously

jennysivapalan:
:+1: :tractor:

PR #54, opened 2015-01-09
It seems sentry doesn't like Exceptions with suppressed stack traces

wpf500:


jennysivapalan:
:+1: :bird:

mattandrews:
:guardsman:

PR #55, opened 2015-01-09
Form layout

feedmypixel:
This is a refactor of the form markup, layout and styles across the site. To improve the markup, have all forms using the same styles and to apply a functional layout to the forms.
 
The following forms have been worked on:
- feedback
- friend signup
- partner & patron signup
- staff signup
- tier upgrade

We now have a form functional layout which is taken from the page functional layout with the only difference being the form does not exceed ``700px`` in width and is centred on the page between ``tablet`` and ``desktop`` break points.

**Note create password is shown only for the below screenshots**

## Payment page - Desktop
payment partner the guardian membership 3

## Payment page - Desktop - with errors
payment partner the guardian membership

## Payment page - Mobile
payment partner the guardian membership 1

## Payment page - Mobile - with errors
payment partner the guardian membership 2

## Feedback page
feedback the guardian membership 1

## Feedback page - Mobile
feedback the guardian membership


frontend/assets/stylesheets/pages/DEPRECATED.txt
```diff
@@ -1 +0,0 @@
-Do not add any new styles to pages/. All new styles should be built as stand-alone components, not tied to page specific usage.
```
davidrapson: /pages is gone! :tada:

frontend/assets/stylesheets/style.scss
```diff
@@ -76,6 +76,7 @@
 @import 'components/nav/brand-bar';
 @import 'components/action';
 @import 'components/page';
+@import 'components/page-form';
```
davidrapson: Not sure about the naming for this. Perhaps:

~~~ scss
@import 'components/form'; // Core form styles
@import 'components/form-layout'; // Form layouts
~~~

frontend/assets/stylesheets/layout/_layout.scss
```diff
@@ -76,6 +76,10 @@ body {
     padding-bottom: rem($gs-gutter * 2);
 }
 
+.l-align-middle {
```
davidrapson: Think this should be `.u-align-middle`. I feel `.l-classes` should be reserved for larger structural helpers (e.g., applied to containers / sections). Element level helpers should be utilities.

frontend/app/views/fragments/joiner/signedInAs.scala.html
```diff
@@ -2,6 +2,6 @@
 
 @import configuration.Config
 
-<div class="content__container">
-    <p>You're signed in as <strong class="js-user-displayname"></strong>. <a href="@Config.idWebAppSignOutThenInUrl(returnUri)">Sign in with another account</a>.</p>
+<div>
```
davidrapson: Naked `div` here feels a little odd. Should this be a paragraph?

frontend/app/views/info/feedback.scala.html
```diff
@@ -6,27 +6,41 @@
 
     <main role="main" class="page-content">
 
-        @fragments.page.pageHeader("Feedback", Some("Help us improve your membership experience"))
+        <section class="form-header">
```
davidrapson: Is a pageHeader displayed differently to a form header? If so we might want to make this a fragment in the same style.

feedmypixel: another example of this can be found [here](https://github.com/guardian/membership-frontend/pull/55/files#diff-c322c0d7e984f78f8e49bfc21c3e4b6eR14). It feels with this example that a fragment doesn't really give us much here

frontend/app/views/fragments/form/cardDetail.scala.html
```diff
@@ -6,49 +6,64 @@
...
-                    Payments will taken on or shortly after the <span class="js-card-note-payment-taken" data-annual="@Dates.dayInMonthWithSuffixAndMonth()" data-month="@Dates.dayInMonthWithSuffix()">@Dates.dayInMonthWithSuffixAndMonth()</span> every <span class="js-card-note-pricing-period" data-annual="year" data-month="month">year</span>.
+                    <p>Your card will be charged £<span class="js-card-note-pricing-charge" data-annual="@pricing.yearly" data-month="@pricing.monthly">@pricing.yearly</span> every <span class="js-card-note-pricing-period" data-annual="year" data-month="month">year</span>.
+                    Payments will taken on or shortly after the <span class="js-card-note-payment-taken" data-annual="@Dates.dayInMonthWithSuffixAndMonth()" data-month="@Dates.dayInMonthWithSuffix()">@Dates.dayInMonthWithSuffixAndMonth()</span> every <span class="js-card-note-pricing-period" data-annual="year" data-month="month">year</span>.</p>
```
mattandrews: Bit of an epic-length line, maybe split it apart?

frontend/assets/javascripts/src/utils/form/Password.js
```diff
@@ -79,9 +70,9 @@ function ($, bean, component) {
 
         require(['js!zxcvbn'], function() {
 
-            var $passwordStrengthIndicatorTemplate = self.$passwordStrengthIndicatorTemplate();
-
-            $passwordStrengthIndicatorTemplate.insertAfter(self.getElem('PASSWORD_STRENGTH_INPUT'));
+            utilsHelper.toArray(document.querySelectorAll('.' + self.classes.STRENGTH_INDICATOR)).map(function (elem) {
+                elem.classList.toggle('is-hidden');
+            });
```
mattandrews: Am I being dense or can you just do: `$('.' + self.classes.STRENGTH_INDICATOR).toggleClass('is-hidden');`?

mattandrews:
LGTM, few absolutely minor things. Great work as usual, really cleans up and improves these tools. Nice! :+1:

jamesoram:
there is a small issue with input widths on iPhone Safari

frontend/assets/javascripts/src/utils/form/Password.js
feedmypixel: No your right that would be the bonzo way, we are aiming to make the JS vanilla where possible.

mattandrews: Ah – well, if that's the plan then maybe we should have a card to remove all usages of bonzo/bean etc. and do things natively? Otherwise if we're already loading them IMO we might as well use them – for looping through arrays I think it's nicer to use the syntactic sugar versions...

feedmypixel: Yup agreed, been speaking with @davidrapson about a gradual move to this was along with an aligning of how we do our JavaScript - psuedo constants, removing componant.js, changing un-needed constructor modules to object literals etc. 
But yes why not we could do this as one big task. Perhaps a little get together to discus this approach and the best way forward at some point?

davidrapson: Probably worth the discussion. Where I'm coming from is that as we are IE9+ we already have enough stuff natively that we probably don't need bean/bonzo. I was happy with just moving towards that one component at a time (even if that means loading bonzo/bean as well for a while), but if it looks it's a small enough task to do and we're all happy with the vanilla JS approach then that's great.

That said if we think that bean/bonzo are still adding value then we should keep them. As Matt says the syntactic sugar can be nicer.

frontend/app/views/info/feedback.scala.html
davidrapson: Yeah, you're right a fragment wouldn't really help here.

jamesoram:
:+1: nice

frontend/assets/javascripts/src/utils/form/Password.js
wpf500: I think bonzo is one of the libraries that actually does add value, especially for this sort of thing. Anything that hides loops and provides a more functional approach is great in my opinion.

feedmypixel: Yup spoke with @mattandrews about this - agreed we don't want code looping through things all over the codebase, we could write our own methods to abstract this out but this is essentially re-inventing the wheel.
Lets plan a CS-dev chat about where we want to head with the JS and what we use from these libs and if we need them or want to look at other solutions. As @davidrapson we have enough native stuff to think about this.

wpf500:
I think there are some redundant styles now. I did a grep for one, `address-fieldset__country`, and it doesn't seem to be needed any more. Would be nice to have a quick search for some of the classes you've removed.

feedmypixel:
> I think there are some redundant styles now. I did a grep for one, address-fieldset__country, and it doesn't seem to be needed any more. Would be nice to have a quick search for some of the classes you've removed.

css removed the majority of the payment css file has now gone

frontend/assets/javascripts/src/utils/form/Password.js
feedmypixel: @mattandrews I have changed this to the bonzo suggestion which makes sense from the conversation. Lets review what the JS plan at some point before we look into refactoring it or write new JS

PR #59, opened 2015-01-12
Stop unwanted requests to /user/me

davidrapson:
Spotted an issue this morning where the recent change to how we create the `GU_MEM` cookie (moved from client-side to server-side), combined with Plays handling of cookies (extra quoting) meant that we were spamming the `/user/me` endpoint.

Switching to `base64` encoded cookies solves the problem, It's also inline with what Identity does. Tests updated to work with base64

// Paired with @wpf500


frontend/test/js/spec/user.spec.js
```diff
@@ -68,9 +68,9 @@ define([
                 spyOn(ajax, 'reqwest');
 
                 // PhantomJS double encodes cookies so this enables the getCookie inside of user.js to return the decoded value
-                spyOn(cookie, 'getCookie').and.returnValue(friendUserDetails);
+                spyOn(cookie, 'getDecodedCookie').and.returnValue(friendUserDetails);
 
-                cookie.setCookie(MEM_USER_COOKIE_KEY, friendUserDetails, 1, true);
+                cookie.setCookie(MEM_USER_COOKIE_KEY, window.btoa(JSON.stringify(friendUserDetails)), 1, true);
```
davidrapson: This is a bit of a code smell as we no longer set the cookies on the client-side. Having to base64 encode ourselves is not great but is needed for now.  There's probably work to be done to rework these tests to not rely on spys anyway.

wpf500:
Its worth pointing out that we are using https://github.com/marklister/base64 for our base 64 encoding.

frontend/assets/javascripts/src/utils/cookie.js
```diff
@@ -29,35 +29,34 @@ define(function () {
...
+        var cookieData = getCookie(name);
+        return decodeCookie(cookieData);
+    }
+
     function removeCookie(name) {
         setCookie(name, '', -1);
     }
 
+    function decodeCookie(cookieData) {
+        return (cookieData) ? JSON.parse(decodeBase64(cookieData.split('.')[0])) : null;
```
feedmypixel: no need for brackets here right?

frontend/assets/javascripts/src/utils/cookie.js
```diff
@@ -29,35 +29,34 @@ define(function () {
...
-            }
+            var c = ca[i];
+            while (c.charAt(0) === ' ') { c = c.substring(1, c.length); }
+            if (c.indexOf(nameEQ) === 0) { return c.substring(nameEQ.length, c.length); }
         }
         return null;
     }
 
+    function getDecodedCookie(name) {
+        var cookieData = getCookie(name);
```
feedmypixel: could be on one line?

frontend/assets/javascripts/src/utils/user.js
```diff
@@ -12,39 +12,18 @@ define([
...
-            }
-            return null;
-        }
-
-        function decodeBase64(str){
-            /*global escape: true */
-            /* See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/escape */
-            return decodeURIComponent(escape(new AtoB()(str.replace(/-/g, '+').replace(/_/g, '/').replace(/,/g, '='))));
+        var userFromCookieCache;
+        var cookieData = cookie.getCookie('GU_U'),
```
feedmypixel: probably best to use either comma separated vars or vars on one line.

frontend/assets/javascripts/src/utils/user.js
```diff
@@ -12,39 +12,18 @@ define([
...
+            userData = cookie.decodeCookie(cookieData);
+        if (userData) {
+            userFromCookieCache = {
+                id: userData[0],
+                displayname: userData[2],
+                accountCreatedDate: userData[6],
+                emailVerified: userData[7],
```
feedmypixel: are we always guaranteed to have these in userData? It may be a little overkill but maybe 
``userData[0] && userData[0]`` could be used as assignments to the relevant objects properties?

davidrapson: Not sure tbh, I've not really touched this code, just cleaned up the decoding part. Can we pick it up in a separate PR?

feedmypixel: sure

frontend/assets/javascripts/src/utils/user.js
davidrapson: Good call

feedmypixel:
Few comments - :+1: for JS work.

davidrapson:
Added some better error catching to handle the case where the user has an old non-base64 encoded cookie. We now check that the value is base64 encoded otherwise re-request from `/user/me` which drops a new cookie.

davidrapson:
@jamesoram Mind giving this a once over?

frontend/app/utils/Base64.scala
```diff
@@ -0,0 +1,69 @@
...
+ *         This applies worldwide. In some countries this may not be legally possible;
+ *         if so: I grant anyone the right to use this work for any purpose, without any
+ *         conditions, unless such conditions are required by law.
+ *
+ *         The repo for this Base64 encoder lives at  https://github.com/marklister/base64
+ *         Please send your issues, suggestions and pull requests there.
+ *
+ */
+
+object Base64 {
```
mattandrews: I'm no Scala guru but there might be a native package in Java/Scala to do this? @jennysivapalan?

jamesoram: the sun.misc package has a Base64Encoder

jennysivapalan: Yep we can do this with a Java library (which I'd prefer).

davidrapson: Cool, if you can let me know which package to use then I can replace it with that. Unless @wpf500 has a strong preference either way as this was his suggestion.

wpf500: I don't really see the point of adding another dependency just for base 64 encoding. This seems like a nice compact bit of code that doesn't add any extra stuff we won't use

jamesoram:
:+1:

davidrapson:
Switched to `org.apache.commons.codec.binary.Base64` which is already in Play as pointed out by @rtyley

mattandrews:
:+1:

jennysivapalan:
:+1: :secret: :lock:

PR #60, opened 2015-01-12
Remove copy to clipboard

davidrapson:
The copy to clipboard icon on test-users / zeroclipboard was the onle place left where we are doing a work around for asset hashing folders; It's also a lot of code and config for such a small feature.

I'm so boring, but I think it should go. 

Sorry @rtyley :cry:


feedmypixel:
:+1:

PR #61, opened 2015-01-13
Update homepage to show different messaging for logged in users

davidrapson:
screen shot 2015-01-13 at 10 54 26

There's a slight jump of content when switching to the logged in state but I think it's small enough that I'm happy with it. There's a question about if we could move a minimal version of the logged in check into the head and add the classes as soon as possible, but I think that can wait for a separate PR.


frontend/assets/stylesheets/components/_slideshow.scss
```diff
@@ -2,22 +2,37 @@
...
+$slideshow-height-tablet: 440px;
+$slideshow-height-desktop: 460px;
+
 .slideshow {
     position: relative;
     overflow: hidden;
+    background-color: $c-neutral4;
+
+    @include mq(tablet) {
+        min-height: rem(440px);
```
mattandrews: Is this meant to be using `$slideshow-height-tablet` above?

mattandrews:
LGTM other than variable comment. :+1:

frontend/assets/stylesheets/components/_slideshow.scss
davidrapson: Good catch

jamesoram:
looks good, not sure if there should be a comma after the name though :+1:

PR #62, opened 2015-01-13
Use flexbox for sticky footer

davidrapson:
The sticky footer currently requires quite a few magic number / fixed heights. It's a good candidate for flexbox which lets us do this without all the positioning.

http://philipwalton.github.io/solved-by-flexbox/demos/sticky-footer/

Only downside is that if you are on IE 9 and you are on a short page (probably only signup) and you have a deep enough viewport you'll see a gap underneath the footer.

Given IE9 is ~2% of traffic and the number of conditions in the previous sentence mean it will only affect a fraction of that % means that this feels OK to me…

### DIFF
--- a/frontend/assets/stylesheets/_vars.scss
+++ b/frontend/assets/stylesheets/_vars.scss
@@ -21,7 +21,6 @@ $gutter-width-fluid: 2%;
 $gutter: 16px; // TODO: Remove this, only used in _joiner.scss
 $trailblock-img-width: 140px;
 $mobilePopupOffset: 43px;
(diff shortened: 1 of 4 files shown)


mattandrews:
Always a :+1: from me for flexbox – agree with the IE9 rationale. Ship it!

wpf500:
I can't publicly endorse flexbox for reputational reasons but :+1:

mattandrews:
You've been slagged off by Nigel Farage today; your reputation is currently unimpeachable.

PR #66, opened 2015-01-13
Clean up RichEvent type matching in templates

wpf500:
This pull request removes all the horrible `event match { case _: GuLiveEvent => ... case _: MasterclassEvent => ... }` code from the template files. A `RichEvent` now specifies a `Metadata` class which contains all the stuff templates need to display the event correctly (in a similar vein to `Benefits` for tiers).

Additionally, removed `allowDiscountCodes` and `maxDiscounts` from `RichEvent` so it is now contains only presentational information.


mattandrews:
:-1:

mattandrews:
Just kidding.

mattandrews:
Nice work. Looks like the key to your best productivity is making you work with annoying bits of the codebase. :+1:

davidrapson:
:100: This is so much better. This felt a little bit weird when I was working with it, glad to know there is a better way!

jennysivapalan:
I'm okay with the code but this is going to cause a lot of merge issues with https://github.com/guardian/membership-frontend/pull/46
I'm hoping to get my PR in tomorrow - can we merge this after?

PR #67, opened 2015-01-13
Too many blues

davidrapson:
Spotted chatting to Ben W that our membership brand colours were a bit all over the place, with some blues being defined but never used and others being defined incorrectly. This PR:

- Rationalises the different blues we have, now clearly defined
- Remove a load of dead variable colours
- Use pasteup-palette where possible for any core brand colours
- Fixes a bug where the Partner level blue was incorrect
- Adds dedicated tone variables to make these a little clearer

There's some future tweaks to be consistent with `camelCase` vs `hyphenated-names` but I'm going to look to see if it makes sense for use to use Sass maps for colours in a later PR anyway.

### DIFF
--- a/frontend/assets/stylesheets/_vars.scss
+++ b/frontend/assets/stylesheets/_vars.scss
@@ -18,7 +18,6 @@ $browser-supports-flexbox: true !default;
 // =============================================================================
 
 $gutter-width-fluid: 2%;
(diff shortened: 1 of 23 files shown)


feedmypixel:
:+1: good stuff - nice to start to get this file under control

PR #69, opened 2015-01-14
Fix build failure. Correct old var from merge

davidrapson:
Getting a build failure on `master`. Caused by merging two branches in which caused an old variable to be introduced.


davidrapson:
Verbal +1 form @rtyley

PR #71, opened 2015-01-14
reduce white space below photograhers credit

feedmypixel:
## Before
screen shot 2015-01-14 at 15 47 47

## After
screen shot 2015-01-14 at 15 47 29


davidrapson:
:+1:

PR #75, opened 2015-01-14
Don't add padding when base64 encoding

davidrapson:
After base64 encoding the `GU_MEM` cookie we were still seeing issues with the cookie value getting quoted. This was due to padding on the `base64` value. Comparing with Identity we should be using `encodeBase64URLSafeString` which does not add padding.

**See:** http://commons.apache.org/proper/commons-codec/apidocs/org/apache/commons/codec/binary/Base64.html#encodeBase64URLSafeString(byte[])

With assistance from @rtyley extracted out the encoding method and added a test.

### DIFF
--- a/frontend/app/controllers/User.scala
+++ b/frontend/app/controllers/User.scala
@@ -1,17 +1,14 @@
 package controllers
 
-import org.apache.commons.codec.binary.Base64
-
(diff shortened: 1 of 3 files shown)


wpf500:
:+1:

PR #78, opened 2015-01-15
Don't throw an exception on encountering an ex-member

rtyley:
This is just publishing the updated membership-common lib from https://github.com/guardian/membership-common/pull/47 - the fix stops these:

https://app.getsentry.com/the-guardian/membership/group/50506635/

...which started https://github.com/guardian/membership-common/pull/45 went out - we were throwing an exception when encountering `null` for `Membership_Tier__c` - but as ex-members can have this state, it's not actually an error.


rtyley:
I'm going to treat https://github.com/guardian/membership-common/pull/47#issuecomment-70066172 as the :+1: :-)

rtyley:
Early days yet, but we shouldn't see either of these after build 4229:

https://app.getsentry.com/the-guardian/membership/group/50648334/
https://app.getsentry.com/the-guardian/membership/group/50506635/

PR #82, opened 2015-01-16
Use SVG sprite for icons

davidrapson:
This is a first step to using SVG sprites for icons. The main benefit is that it allows us to style and manipulate the icons in CSS, removing the need to have one icon per colour.

- New directory `inline-icons` which includes a set of initial test icons
- Sprite is inlined using the same technique that frontend use to load their CSS https://github.com/guardian/frontend/blob/master/common/app/views/fragments/loadCss.scala.html
- Adds `@fragments.inlineIcon` to help with referencing the icons.

From speaking to Dominic, this is very similar to the approach taken on Soulmates. Nextgen have also done something similar this week https://github.com/guardian/frontend/pull/7785

Current test icons are the G icon for tiers and the home icon.
The tier icons are a good test case as they are they are colour-coded so allow us to test styling icons directly.

**Chrome**
screen shot 2015-01-16 at 11 20 05

**IE9**
screen shot 2015-01-16 at 11 21 46

**iOS**
img_1067


frontend/app/views/fragments/inlineIcon.scala.html
```diff
@@ -0,0 +1,5 @@
+@(name: String, classes: Seq[String] = List())
+
+<svg class="icon-inline @classes.map("" + _).mkString(" ")">
```
wpf500: What does the `map` do here?

davidrapson: TBH I used the same pattern from here https://github.com/guardian/frontend/blob/aa183a4395b1f89404585c85500471c083e2c036/common/app/views/fragments/inlineSvg.scala.html

I'm not appending a prefix here, so it feels a little awkward. I'm sure there's a better way of doing this. I just want to loop over an optional list of extra classes.

wpf500:
:+1:. @jamesoram, not sure if you want to take a look at this? Only icon that has currently changed is the G with different coloured backgrounds.

jamesoram:
:+1:

PR #83, opened 2015-01-16
Rework of EventbriteService/RichEvent

wpf500:
Some general clean up of `EventbriteService` and work around how a service is linked to a `RichEvent`.

- Use `ScheduledTask` for Eventbrite's tasks (yes I'm finally forcing this one in there!)
- Renamed `getAllEvents` to `getEvent`, which is what it actually does

I'm playing around with the idea of an implicit which looks up the service that an event belongs to. I'd like to eventually get to a point where we can (definitely) remove `GuLiveEvent` and `MasterclassEvent` and (possibly) remove `RichEvent`.

This approach removes almost all `GuLiveEvent`/`GuMasterclassEvent` specific code except for access codes/discount codes, which I've ignored because @feedmypixel  is removing this anyway.

Thoughts/opinions? Especially @rtyley @jennysivapalan


frontend/app/services/EventbriteService.scala
```diff
@@ -18,33 +16,43 @@ import model.Eventbrite._
...
+      events <- getPaginated[EBEvent]("users/me/owned_events?status=live")
+      richEvents <- Future.sequence(events.map(mkRichEvent))
+    } yield richEvents
+  }
+
   val refreshTimeArchivedEvents = new FiniteDuration(Config.eventbriteRefreshTime, SECONDS)
-  val refreshTimePriorityEvents = new FiniteDuration(Config.eventbriteRefreshTimeForPriorityEvents, SECONDS)
+  lazy val archivedEventsTask = ScheduledTask[Seq[RichEvent]]("Eventbrite archived events", Nil, 1.second, refreshTimeArchivedEvents) {
+    for {
+      eventsArchive <- get[EBResponse[EBEvent]]("users/me/owned_events?status=ended&order_by=start_desc")
```
jennysivapalan: Might still want the comment about getting one page due to Masterclasses?

jennysivapalan:
Nice :+1: :ticket:

mattandrews:
Agree with @jennysivapalan about the comment (not just because I wrote it!)

PR #84, opened 2015-01-16
Minor design refinements

davidrapson:
Few minor design refinements from Ben W:

- Border on featured section in events list should be blue
- Decrease size of page headers on small viewports
- Cleanup the way social/share this headers are defined


mattandrews:
LGTM :+1:

PR #85, opened 2015-01-16
Update 500 page to use functional page layout

davidrapson:
Update 500 page to use functional page layout


mattandrews:
Nice. :+1: 

Hopefully nobody will ever see this, mind.

PR #86, opened 2015-01-16
Cleanup actions styles

davidrapson:
Initial refactor to cleanup actions styles. Reduces number of modifiers and removes some single-use classes.


mattandrews:
:+1:

PR #87, opened 2015-01-16
Be able to preview events

jennysivapalan:
This is to help the Events team preview draft events before launching.

As we don't want to expose this to everyone there is a new endpoint which is `/preview-event/{id}`. This endpoint:
- requires Google auth (so must have a guardian email). 
- directly calls the Eventbrite API to get the most recent details of the event
- has a short cache of 1 second (to get updates quickly).

I've only applied this to Guardian Live events and not Masterclasses. (We can add this later if they need but it requires more work for figuring which service to call.)

I've pushed some of the logic on which status text to display from the template to the Scala code. This makes it testable. The downside is there is a horrible if condition. Sorry @wpf500 

@jamesoram let me know when you're free and we can go through the testing of this. 

Did I mention I'm going on holiday? :ski:


frontend/app/model/Eventbrite.scala
```diff
@@ -102,7 +102,13 @@ object Eventbrite {
...
     val isNoTicketEvent = description.exists(_.html.contains("<!-- noTicketEvent -->"))
     val isBookable = status == "live" && !isSoldOut
-    val isPastEvent = status != "live"
+    val isPastEvent = status != "live" && status != "draft"
+
+    val statusText =
+      if(isPastEvent) "Past event"
+      else if(isSoldOut) "Sold out"
+      else if(status == "draft") "Preview of Draft Event"
+      else ""
```
wpf500: Not a massive fan but seen as you're going on holiday! :ski:

wpf500:
:+1:

frontend/app/model/Eventbrite.scala
jennysivapalan: No, it's rubbish. but not sure how to improve this :put_litter_in_its_place:

jamesoram:
:+1:

rtyley:
Looks great!

https://membership.theguardian.com/preview-event/15351670259
https://membership.theguardian.com/preview-event/15355009246

:+1: :+1: :+1: :+1: :+1:

PR #89, opened 2015-01-19
Use working mixin

mattandrews:
We had a rendering issue on FF19 which was admittedly not affecting many people, but stems from what I suspect is [an issue](https://github.com/guardian/guss-css3/issues/11) with guss-css3. 

#### Before:
screen shot 2015-01-19 at 11 38 52

#### After:
screen shot 2015-01-19 at 11 39 41


davidrapson:
:+1: 

Nice. Looks like Firefox wen't unprefixed in version 21, so thankfully not affecting too many people.  Good catch though. As an aside, I'd be interested in looking into autoprefixer at some point as an alternative for handling vendor prefixes.

mattandrews:
Okay with you @jamesoram?

jamesoram:
:+1:

PR #90, opened 2015-01-19
Add has member ticket property to EbEvent case class

feedmypixel:
- added a hasMemberTicket helper property to EBEvent case class
- added tests

PR #91, opened 2015-01-19
Show discounts for guardian membership events only

feedmypixel:
- If an event has a "Guardian Member" ticket on an event we allow this event to discounted for all users as it is currently.
- If there is no "Guardian Member" ticket on an event then the event will not be discounted for any user.

This pull request is connected with [this one](https://github.com/guardian/membership-frontend/pull/90)

# List page (no discount event on the right)
screen shot 2015-01-19 at 12 10 16

# Event page (discounted event)
screen shot 2015-01-19 at 12 12 53

# Event page (no discount event)
screen shot 2015-01-19 at 12 13 07

### DIFF
--- a/frontend/app/views/event/page.scala.html
+++ b/frontend/app/views/event/page.scala.html
@@ -172,19 +172,35 @@
                                         </div>
                                     } else {
                                         @for(pricing <- ticket.cost) {
-                                            <div class="js-event-price event-content__item">
-                                                <div class="event-content__price event-content--line-divider js-event-price-value" data-discount-text="@pricing.discountPrice" itemprop="highPrice">
-                                                    @pricing.formattedPrice
+
+                                            @if(event.hasMemberTicket) {
+                                                <div class="js-event-price event-content__item">
+                                                    <div class="event-content__price event-content--line-divider js-event-price-value"
+                                                         data-discount-text="@pricing.discountPrice"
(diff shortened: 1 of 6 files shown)


frontend/app/views/event/page.scala.html
```diff
@@ -172,19 +172,35 @@ <h1 class="event-masthead__name">@event.name.text</h1>
                                         </div>
                                     } else {
                                         @for(pricing <- ticket.cost) {
-                                            <div class="js-event-price event-content__item">
-                                                <div class="event-content__price event-content--line-divider js-event-price-value" data-discount-text="@pricing.discountPrice" itemprop="highPrice">
-                                                    @pricing.formattedPrice
+
+                                            @if(event.hasMemberTicket) {
+                                                <div class="js-event-price event-content__item">
+                                                    <div class="event-content__price event-content--line-divider js-event-price-value"
```
wpf500: Can't the if statement just wrap the discounted price/saving price code?

feedmypixel: Sure I started with this approach. I also want to take the .js-event-price class out of action so the check in js is just if this element exists. I also don't want a data-discount-text attribute sat in the markup for non discounted events. It was this or wrap if statements around those classes and attributes. This felt cleaner.

wpf500:
:+1:

PR #93, opened 2015-01-19
Show better event URLs

wpf500:
Add a nice slug to the event URLs so they look nicer and have better SEO.

e.g.
`/event/15327407689` becomes `/event/guardian-live-guantanamo-diary-15327407689`

URLs without slugs or with invalid slugs (e.g. after an event name change) redirect to the correct URL.

### DIFF
--- a/frontend/app/controllers/Event.scala
+++ b/frontend/app/controllers/Event.scala
@@ -1,5 +1,7 @@
 package controllers
 
+import model.Eventbrite.EBEvent
+
 import scala.concurrent.Future
 
 import play.api.mvc._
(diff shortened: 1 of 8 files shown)


frontend/app/controllers/Event.scala
```diff
@@ -38,10 +40,19 @@ trait Event extends Controller {
   private def BuyAction(id: String) = NoCacheAction andThen recordBuyIntention(id) andThen
     authenticated(onUnauthenticated = notYetAMemberOn(_)) andThen memberRefiner()
 
-  def details(id: String) = CachedAction { implicit request =>
-    EventbriteService.getEvent(id).map { event =>
-      eventDetail(event, request.path)
-    }.getOrElse(Redirect(Config.guardianMembershipUrl))
+  def details(slug: String) = NoCacheAction { implicit request =>
+    val eventOpt = for {
+      id <- EBEvent.slugToId(slug)
```
mattandrews: This is more of a learning question than a review, but does redefining `id` here only apply to the scope of the `for`?

mattandrews: ... I can't read.
:eyeglasses:

mattandrews:
:+1: i love urls

jamesoram:
:+1:

PR #95, opened 2015-01-19
Event overview staff page

feedmypixel:
This is a page that gives easy to check information on all live events on membership and can be found ``/staff/event-overview``

- behind staff oauth,
- shows live events and displays if they are 
  - free, 
  - discounted, 
  - have the &lt;!-- noTicketEvent --&gt; comment, 
  - are sold out, 
  - contain a "Guardian Member" ticket

It also shows the:
  - current main image,
  - title of the event and links to,
  - ophan,
  - event preview

A number of things worth mentioning:
- I have created a mainBasic.scala.html file as we had a lot of markup in the main.scala.html that is not needed here. 
- This mainBasic.scala.html brings in our CSS and JS. Not a lot of this is being used, currently it is just the Imager work and some styles so potentially a utils stylesheet and js file could be made or not as this is for staff only.

# Desktop
event overview the guardian membership

# Mobile
event overview the guardian membership 1

### DIFF
--- a/frontend/app/controllers/Staff.scala
+++ b/frontend/app/controllers/Staff.scala
@@ -1,0 +1,16 @@
+package controllers
+
+import play.api.mvc.Controller
+import services.GuardianLiveEventService
+
+import scala.concurrent.Future
+
+trait Staff extends Controller {
+  val guLiveEvents = GuardianLiveEventService
+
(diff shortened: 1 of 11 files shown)


tackley:
Can I suggest that you change the wording of the Ophan link to "View in Ophan", which is what we consistently use elsewhere.

PR #98, opened 2015-01-20
wrapped the price in the header with the hasMemberTicket conditional

feedmypixel:
The price in the header also needs to not be discounted when an event does not have a "Guardian Member" ticket. Connected with [this](https://github.com/guardian/membership-frontend/pull/91) pull request

## No Discount
screen shot 2015-01-20 at 16 29 10

## With Discount
screen shot 2015-01-20 at 16 29 17


wpf500:
:+1:

PR #99, opened 2015-01-21
Fix page header line-height

davidrapson:
Pretty radical change here:

**Before**
screen shot 2015-01-21 at 11 36 30

**After**
screen shot 2015-01-21 at 11 36 37


feedmypixel:
:+1:

PR #101, opened 2015-01-21
Move some more icons over to inline SVGs

davidrapson:
Now that we have the ability to use inline SVGs for icons (hooray!), I'm starting to move over a few more icons. This PR moves the caption icon, the help / FAQ and the social icons.

Social icons are the biggest change as it allows us to test some hover effects out:

social-hover mov

Look at that sweet hover.

It also extracts out the circle colour from the icon itself so these icons can be used standalone.

As a reference, there is a section in the pattern library which auto-documents the icons in the SVG sprite so we can see how they work together as a set (and make sure there are no duplicates)

screen shot 2015-01-21 at 17 32 21


frontend/assets/stylesheets/components/_icons.scss
```diff
@@ -82,3 +82,26 @@
...
+    vertical-align: middle;
+    padding: 0; margin: 0;
+    width: 26px; height: 26px;
+    background-color: transparent;
+    border-radius: 1000px;
+    border: 1px solid $c-border-brand;
+    color: $mem-brandBlue;
+
+    .icon-inline {
+        @include transform(translateY(-50%));
```
mattandrews: Could you add a comment just explaining what this is for?

frontend/assets/stylesheets/components/_social.scss
```diff
@@ -1,20 +1,58 @@
...
 
-.social__item {
-    float: left;
-    padding-left: rem(10px);
-    width: rem(36px + 10px);
+    &:hover {
+        background-color: darken($c-neutral2, 10%);
+    }
+
+    .icon-inline {
```
mattandrews: This is basically the same as the other definition above, minus the dimensions – could it be a modifier instead?

mattandrews:
LGTM besides comments

frontend/assets/stylesheets/components/_social.scss
davidrapson: I went down this route originally but it felt a little bit more explicit to have them separately as they server different functions. Going to move the vertical centring styles to a mixin I think.

davidrapson:
@jamesoram Mind giving this a once over?

jamesoram:
Some SVGs are not appearing in IE9

davidrapson:
@jamesoram This is ready for you to test again

davidrapson:
Put a separate PR up for the IE9 issue https://github.com/guardian/membership-frontend/pull/112

We'll get that live first (once confirmed) and then we can be more confident to roll out further icons.

jamesoram:
:+1:

PR #100, opened 2015-01-21
eEvent overview staff page

feedmypixel:
This is a page that gives easy to check information on all Live events and Draft events associated with Membership. It can be found on ``/staff/event-overview``

- behind staff oauth,
- shows Live and Draft events and displays if they are a:
  - Free Event, 
  - No Discount Event, 
  - Not Sold through Eventbrite, 
  - Sold out

It also shows the:
  - Current main image,
  - Title of the event,
  - link to view in Ophan,
  - link to preview event,
  - link to edit page in eventbrite 

## Notes
- I have created a tools.scala.html file as we had a lot of markup in the main.scala.html that is not needed here. 
- This tools.scala.html brings in a separate stylesheet. There a few dependencies to be fixed in another pull request meaning I have had to bring in more than I need.

# Example - Events with Tags
screen shot 2015-01-21 at 11 57 50

# Desktop
event overview the guardian membership

# Mobile
event overview the guardian membership 1

@tackley for consistency the Ophan link now reads - "View in Ophan"

PR #102, opened 2015-01-21
use secure image for social url

feedmypixel:
I noticed that the social image url is not https as these are used directly on the [event overview page](https://github.com/guardian/membership-frontend/pull/100). 
- Use secure image
- Update to tests


mattandrews:
:+1:

PR #103, opened 2015-01-22
discard GU_MEM cookie on upgrade

feedmypixel:
- remove the GU_MEM cookie when a user upgrades 
@wpf500 can you have a look over do we need this anywhere else?


wpf500:
Looks OK to me :+1:

PR #104, opened 2015-01-22
Fix layout issues in Safari 6

davidrapson:
Safari before 6.1 needs [webkit prefix for calc ](http://caniuse.com/#feat=calc)

**Before**
safari-before

**After**
safari-after

(Note web-fonts are not present in the after screenshot because this was pointing to my dev version on the internal network which doesn't have permission)

If we want to support Safari 5 as well we should look into not relying on `calc` for critical layout styles (mostly just using px for the `html` reset and refactoring the header layout a bit).

### DIFF
--- a/frontend/assets/stylesheets/base/_base.scss
+++ b/frontend/assets/stylesheets/base/_base.scss
@@ -33,6 +33,7 @@ html {
     // This is used as a baseline for rem (root ems) values
     // Calc needed for IE11 to do the math properly
     // See http://bit.ly/1g4X0bX — thanks @7studio and @dawitti
+    font-size: -webkit-calc(1em * .625);
     font-size: calc(1em * .625);
 }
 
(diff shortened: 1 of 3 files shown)


mattandrews:
Ha, as simple as that? I think your suggestion of autoprefixer is a good shout for future issues.

davidrapson:
Thankfully yes! I think autoprefixer would be worth it, manually managing vendor prefixes is a fools errand.

mattandrews:
Also, Safari 5 is 3% of our Safari traffic, which in turn is 25% of total, so I don't think we should worry about it.

davidrapson:
@jamesoram Are you able to test this one? Needs a version of Safari pre-6.1 (`6.0.5` is our reference version).

wpf500:
Personally I thought it looked better before the fix

jamesoram:
:+1:

PR #110, opened 2015-01-22
Update to use the latest version of the membership-common library

rtyley:
...which doesn't include the whole AWS sdk anymore, so we need to explicitly include awsSimpleEmail (the overall deployable artifact size should reduce).


wpf500:
:+1:

PR #112, opened 2015-01-22
Fix issue with SVG icons occasionally not showing in IE 9

davidrapson:
As part of this PR https://github.com/guardian/membership-frontend/pull/101 , @jamesoram spotted an issue where inline SVG icons would occasionally not show up. We managed to narrow it down to an issue with `DOMContentLoaded` not always firing (and checking `readyState` first didn't reliably resolve the problem).

Thankfully Paul Irish documented a more reliable approach (used in Modernizr) which in testing with James seems to be working for us and avoids the need to wait for a ready event: http://www.paulirish.com/2011/surefire-dom-element-insertion/ 

screen shot 2015-01-22 at 17 31 23


jamesoram:
:+1:

PR #113, opened 2015-01-22
update locators

jamesoram:


functional-tests/src/main/scala/com/gu/membership/pages/ChangeTierPage.scala
```diff
@@ -9,9 +9,9 @@ class ChangeTierPage(driver: WebDriver) extends BaseMembershipPage(driver) {
 
   private def becomeAFriendButton = driver.findElement(By.cssSelector(".package .action.action-cta--confirm"))
 
-  private def becomeAPartnerButton = driver.findElement(By.cssSelector(".package .action.action-cta--confirm"))
+  private def becomeAPartnerButton = driver.findElement(By.xpath(".//*[@id='container']/div/main/div[2]/div[1]/section/div[2]/div[2]/div[2]/a"))
 
-  private def becomeAPatronButton = driver.findElement(By.cssSelector(".package .action.action-cta--confirm"))
+  private def becomeAPatronButton = driver.findElement(By.xpath(".//*[@id='container']/div/main/div[2]/div[1]/section/div[3]/div[2]/div[2]/a"))
```
davidrapson: Can we provide you with better hook here? This looks pretty brittle.

jamesoram: Agreed, it would be nice! IDs are preferred... or a class that is specific to each button.

davidrapson: Will get some IDs added.

PR #114, opened 2015-01-22
Remove dead join/choose-tier IE9 styles

davidrapson:
Found some dead styles in `_ie9.scss` which caused this interesting bug:

**Before**
screen shot 2015-01-22 at 17 38 52

(It keeps going…)

**After**
screen shot 2015-01-22 at 17 39 31


davidrapson:
@feedmypixel Mind giving this a quick review?

feedmypixel:
:+1:

PR #116, opened 2015-01-23
IE9 cleanup and patrons pricing update

davidrapson:
Did some cleanup of the IE9 specific styles which means we can now remove the IE9 override stylesheet!

- **Updated patrons page to use new style package promos (Agreed with Ben W). Removes need for remaining IE9 overrides**
- Cleanup of footer to use floats throughout (no more CSS columns)
- Bugfix to the main navigation to remove overflow IE9 fix

screen shot 2015-01-23 at 13 52 00

screen shot 2015-01-23 at 12 46 08

### DIFF
--- a/frontend/app/views/info/about.scala.html
+++ b/frontend/app/views/info/about.scala.html
@@ -111,10 +111,10 @@
         </div>
     </div>
 
-    <div class="page-slice">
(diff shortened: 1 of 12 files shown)


wpf500:
:+1:

PR #119, opened 2015-01-23
Fill hero events with normal events if necessary

wpf500:
Before:
image

After:
image


frontend/app/services/EventbriteService.scala
```diff
@@ -161,13 +160,18 @@ object MasterclassEventService extends EventbriteService {
...
 }
 
-object MasterclassEventServiceHelpers {
+object EventbriteServiceHelpers {
   def availableEvents(events: Seq[RichEvent]): Seq[RichEvent] =
     events.filter(_.memberTickets.exists { t => t.quantity_sold < t.quantity_total } )
+
+  def getFeaturedEvents(orderedIds: Seq[String], events: Seq[RichEvent]): Seq[RichEvent] = {
+    val (orderedEvents, normalEvents) = events.partition { event => orderedIds.contains(event.id) }
+    orderedEvents.sortBy { event => orderedIds.indexOf(event.id) } ++ normalEvents.filter(!_.isSoldOut).take(4 - orderedEvents.length)
```
feedmypixel: will we have a case where orderedEvents.length is less than 4?

wpf500: Yeah, when they don't select 4 events to feature or when one of the featured events is non-live (i.e. probably already happened)

feedmypixel:
LGTM, Nice work :+1:

PR #122, opened 2015-01-23
Enable multiple ticket classes for events

wpf500:
Eventbrite have fixed their API, so we can now associate multiple ticket classes with an access code!

image

Note that the access code still only has a maximum usage of 2 so a user can't buy 2 of each member ticket class


feedmypixel:
:+1:

PR #123, opened 2015-01-26
Add hooks for functional tests

davidrapson:
Paired with @jamesoram to add some `qa-` hooks to elements that had the most brittle selectors in the functional tests.


jamesoram:
:+1:

PR #124, opened 2015-01-26
Use MemberId over String

wpf500:
Use the type system to enforce that these methods need a member's Salesforce contact/account ID, rather than just letting them take arbitrary `String`s


mattandrews:
Types FTW.

jennysivapalan:
This change now makes sense now after reading this https://github.com/guardian/membership-common/pull/51
:+1:

PR #125, opened 2015-01-26
Remove need for duplicate navigation markup

davidrapson:
Noticed this whilst working through refactoring functional tests with @jamesoram . Currently the global navigation works by including two versions of the nav on the page and toggling between them.

Whilst the nav is based on the guardian.com responsive nav, we don't have the same behaviour when toggling so we can happily handle the nav toggle with CSS, rather than duplicate markup.

- Removes need for duplicate navigation markup
- Removes need for magic numbers in layout / side-margin styles / Centralises layout for header, footer etc.
- Extracts navigation toggle JS out from `Header.js` and into a standalone component

Also fixes a firefox layout bug for the signed in members link:

screen shot 2015-01-26 at 12 59 30

screen shot 2015-01-26 at 12 59 41

### DIFF
--- a/frontend/app/views/fragments/header.scala.html
+++ b/frontend/app/views/fragments/header.scala.html
@@ -75,28 +75,19 @@
                         </ul>
                     </nav>
                 </div>
-                <div class="mobile-menu mobile-only">
-                    <a href="#" role="menuitem" class="mobile-menu__button js-menu-icon" title="Menu">
-                        <span class="control__icon">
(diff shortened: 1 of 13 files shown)


jamesoram:
:+1: nice

mattandrews:
This always bothered me. Tried to look at it early on in the project but we were too tightly-coupled to NGW to modify it. Great work :+1:

PR #127, opened 2015-01-26
Remove unused code

wpf500:


wpf500:
This is live but prout is being silly

PR #128, opened 2015-01-27
Use new Tier case objects

wpf500:
Upgrade `membership-common` to 0.50 to make use of the new `Tier` case objects.

Most of this change involves changing `Tier.Value` or `Tier.Tier` to just `Tier`.

Tiers now contain some properties, one of which is `slug`, which can be used in place of `Tier.toString.toLowerCase`, it is also what we use for routing.


wpf500:
cc @rtyley @jennysivapalan especially for code review

wpf500:
@jamesoram this contains quite a few changes, some of which affect the join flows. Do you want to do anything to check you are happy with the changes?

jennysivapalan:
Nice :+1: :runner:

jamesoram:
:+1:

PR #131, opened 2015-01-27
Upgrade Content API client

jennysivapalan:
In preparation for making another call to to the Content API I have updated the version of the Content API client and moved the queries to a separate trait.


wpf500:
:+1:

PR #129, opened 2015-01-27
Update event thank you page to use functional layout

davidrapson:
In line with the recent cleanup of the registration thank you page, this PR updates the event thank you page 

- Refactor thankyou page
- Share item shapshot and event stats accross event types
- Update event icons to use inline SVGs / correct grey colour (as per Ben W designs)
- Clean up dead styles from old thank you page
- Move eventbrite logo out of icons and load it as an image

**New**
screen shot 2015-01-27 at 15 54 47

**Old**
screen shot 2015-01-27 at 14 47 03

PR #132, opened 2015-01-28
Cleanup unused data-uri icons

davidrapson:
Now that a large number of icons have moved over to inline SVG we can clean out quite a few of the data-uri'd icons.

- Removes benefits icons as these have all moved over to inline SVGs
- Removed some unused colour variants: arrows / g logo
- Updates the search filter icons and back to top icon to use inline SVGs
- Directly inline blockquote image rather than extending

Quite a few more can go once https://github.com/guardian/membership-frontend/pull/129 has been merged in too.

As a rough metric:

Stylesheet size pre cleanup: **213 KB / 48.6 KB (gzip)**
Stylesheet size post cleanup: **188 KB / 43.5kb (gzip)**

Gzip chews through a fair bit of the saving but it's still a decent decrease. (As an aside I'm going to be looking at getting some asset metrics data into the dashboard in the near future)

screen shot 2015-01-27 at 17 52 45

screen shot 2015-01-28 at 10 50 24


feedmypixel:
:+1:  nice cleanup

PR #133, opened 2015-01-28
User me endpoint cookie rework

feedmypixel:
### This has a breaking change for frontend/Identity - DO NOT merge until [this](https://github.com/guardian/frontend/pull/8042) pull request merged on frontend

I noticed the code for the /user/me endpoint wasn't quite doing what it should be doing and the creation/removal of the GU_MEM cookie wasn't quite right.
This work is involved with the discarding of the GU_MEM cookie and the creation of a minimal GU_MEM cookie for identity signed in users. As well as a good rework of the client side code used to call the /user/me endpoint.

When a user hits the /user/me endpoint:
 - Identity not signed in = 403
 - Identity signed in, not a member = minimal GU_MEM cookie 
 - Identity signed in, member = same response as present 

- Move the creation of the minimal GU_MEM cookie to the backend, this is created when a user has signed in with Identity but does not have a membership tier
- Discard the GU_MEM cookie when a thank you page is shown, covering joining and upgrading,
- Tidy up the client side work involved with creating the minimal GU_MEM cookie
- Tidy up the client side work involved with calling the /user/me endpoint
- Removed client side tests
- Cookie refresh work is no longer needed
- Thnak you page client side work is now being done on the backend


feedmypixel:
@jamesoram please have a look over this and corresponding front end pull request
@rtyley or @wpf500 can you sanity check the scala work in this pull request

frontend/assets/javascripts/src/main.js
```diff
@@ -84,7 +77,6 @@ require([
     (new JoinFree()).init();
     (new PaidForm()).init();
     (new StaffForm()).init();
-    (new Thankyou()).init(header);
```
davidrapson: Nice! Now that `Thankyou` is gone we can also update the Header initialisation to be `(new Header()).init();` as we no longer need to pass the header instance around (which is good)

davidrapson:
Minor comment but really nice cleanup work on the JS :+1:

jamesoram:
:+1:

frontend/app/controllers/User.scala
```diff
@@ -19,8 +19,6 @@ trait User extends Controller {
     Ok(json).withCookies(Cookie("GU_MEM", GuMemCookie.encodeUserJson(json), secure = true, httpOnly = false))
   }
 
-
-
```
rtyley: Always worth trying to reduce the size of your diff, so consider leaving out formatting changes if they're not essential (can always tidy up in a separate PR)

frontend/app/services/AuthenticationService.scala
```diff
@@ -20,6 +20,11 @@ trait AuthenticationService {
     user = guUCookieData.user if user.id == minimalSecureUser.id
   } yield IdMinimalUser(user.id, user.publicFields.displayName)
 
+  def identityAuthenticatedUser[A](request: RequestHeader): Option[IdMinimalUser] = for {
```
rtyley: Very curious about this. Why has this been added, when `authenticatedUserFor()` exists above, and does very similar stuff?

feedmypixel: Good spot - I can use the authenticatedUserFor method I will swap this out.

frontend/assets/javascripts/src/main.js
feedmypixel: thats been added in

feedmypixel:
> Could we maybe add another simpler test for just the decodeBase64 method too? i.e., generic base64 string and test that that is decoded correctly?

tests beefed up

feedmypixel:
[#8042](https://github.com/guardian/frontend/pull/8042) is live on front-end pushing this to membership now

feedmypixel:
on Prod and works with [#8042](https://github.com/guardian/frontend/pull/8042) membership tap on Identity displayed as intended

PR #136, opened 2015-01-28
Add js-opt-become-member class to package actions

davidrapson:
Add js-opt-become-member class to package actions. Already included on all other actions aside from the ones on Join / choose-tier. Gives us a hook to track the relevant action.


feedmypixel:
:+1:

PR #138, opened 2015-01-30
Fix bug with active state on links + clean up disabled styles

davidrapson:
Fix up an annoying visual bug with the active state for buttons, and simplify disabled states for buttons.

**Before** (no text-colour when active)
screen shot 2015-01-30 at 11 01 38

**After**
screen shot 2015-01-30 at 11 02 52


mattandrews:
:+1:

PR #141, opened 2015-02-02
Schema / rich data updates

davidrapson:
Couple of updates to fix Schema / structured data errors reported by https://developers.google.com/webmasters/structured-data/testing-tool/

- AggregateOffer requires a url
- Organization requires a logo

As part of "Organization requires a logo" I've updated the header and footer logos to be inline images, rather than icon sprites. As this feels like a more appropriate way of representing them (keeping icons solely to UI elements).

Still have a double logo in the header but this can probably go when we come out of beta.

screen shot 2015-02-02 at 12 10 09
screen shot 2015-02-02 at 12 10 13
screen shot 2015-02-02 at 12 10 24


mattandrews:
Nice. All about that sweet metadata. :+1:

PR #144, opened 2015-02-02
Disable analytics in DEV mode

davidrapson:
Just thought I'd check if there there a reason we have analytics enabled for all stages?

### DIFF
--- a/frontend/app/views/fragments/javaScriptFirstSteps.scala.html
+++ b/frontend/app/views/fragments/javaScriptFirstSteps.scala.html
@@ -39,7 +39,7 @@
             trackingId: '@Config.googleAnalyticsTrackingId',
             cookieDomain: @if(Config.stage == "PROD") { 'auto' } else { 'none' }
         },
-        analyticsEnabled: true,
+        analyticsEnabled: @if(Config.stage == "PROD") { true } else { false },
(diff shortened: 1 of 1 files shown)


mattandrews:
We load a dev account for Google Analytics (useful for testing), although Omniture/Ophan are still loaded for DEV (and presumably disregard data that doesn't match the domain whitelist). @rtyley?

tackley:
Yup - ophan will definitely exclude by default all traffic inside a guardian office. Personally I'd recommend you keep it enabled in DEV stages as (I'm biased) analytics is probably the most important function not to break. Therefore having it working (/breaking) in DEV and pre-production I'd view as very important.

In the ophan dashboard, you can view inside-the-guardian traffic by adding country=GNM to any url.

mattandrews:
Ha, the independent republic of GNM...

Also, I swear @tackley has a batsignal-like alert every time a comment mentions Ophan :grinning:

davidrapson:
@tackley Thanks Graham! Totally agree, I was sure there would be a good reason.

tackley:
@mattandrews actually yes - all my github notifications skip my inbox **unless** the mention ophan :)  That then turns into a call to strip down to my pants and put my cape on.

PR #145, opened 2015-02-03
Stop Twitter cards derailing users into the mobile apps

rtyley:
Unfortunately the mobile apps don't understand Membership urls, and so when a user clicks 'Open in the Guardian app', they just get taken to the 'homepage' of the app - not to Membership.

While it's nice to promote the apps, until they open the links in a way that's helpful to the user, we should remove the 'Open in the Guardian app' functionality from our Twitter cards.

https://dev.twitter.com/cards/mobile

cc @mattandrews

### DIFF
--- a/frontend/app/configuration/Config.scala
+++ b/frontend/app/configuration/Config.scala
@@ -143,10 +143,6 @@ object Config {
   }
 
   val twitterUsername = config.getString("twitter.username")
-  val twitterIphoneAppName = config.getString("twitter.app.iphone.name")
-  val twitterIphoneAppId = config.getString("twitter.app.iphone.id")
-  val twitterGoogleplayAppName = config.getString("twitter.app.googleplay.name")
-  val twitterGoogleplayAppId = config.getString("twitter.app.googleplay.id")
 
   val googleAnalyticsTrackingId = config.getString("google.analytics.tracking.id")
 
--- a/frontend/app/views/main.scala.html
+++ b/frontend/app/views/main.scala.html
@@ -28,10 +28,6 @@
         <meta property="og:type" content="website"/>
         <meta property="fb:app_id" content="@Config.facebookAppId"/>
         <meta name="twitter:site" content="@@@Config.twitterUsername"/>
-        <meta name="twitter:app:name:iphone" content="@Config.twitterIphoneAppName"/>
-        <meta name="twitter:app:id:iphone" content="@Config.twitterIphoneAppId"/>
-        <meta name="twitter:app:name:googleplay" content="@Config.twitterGoogleplayAppName"/>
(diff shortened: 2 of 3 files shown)


mattandrews:
Ah good point – in all honesty I didn't know what those tags did when I added them, assumed it'd just upsell the app somewhere (or associate the links with a promo for our app, not try to open them in it). I wonder if @rupertbates has any thoughts on this – should the app(s) ignore `membership.theguardian.com/*` links?

rtyley:
> should the app(s) ignore membership.theguardian.com/* links

As it happens, for the Android app at least, they don't specifically capture membership.theguardian.com in their intent-filters:

https://github.com/guardian/android-news-app/blob/658a2c37/android-news-app/AndroidManifest.xml#L141-L191

I think the twitter tags we include on the page basically seal the deal of going off to the mobile app (which knows nothing about membership, consequently revert to the app homepage). By adding the tags, we tell the twitter app that we're confident the referenced app can handle the url - tho' unfortunately it can't :-]

mattandrews:
Ah, fair enough. Let's kill them, then! :hocho: 

:+1:

rupertbates:
Is there anything useful the apps could do with these urls (other than just open it in a webview)?

rtyley:
> Is there anything useful the apps could do with these urls (other than just open it in a webview)?

If effort was no object, then perhaps the apps could use the users authenticated details to ensure that the user saw pricing appropriate to them, or make event data available offline. If membership becomes truly huge that might be worth the effort, in the mean time, I think the effort is better spent elsewhere!

For instance, bundling membership and premium app access is probably a better use of our time :-)

rupertbates:
Makes sense :+1:

rtyley:
I've checked out the feminism event on the android Twitter app, unfortunately, even though the page source has the 'twitter:app' meta tags removed, the Twitter App is still showing the 'Open in App' link:

screenshot_2015-02-03-11-39-23

This happens even if I tweak the url slightly to https://membership.theguardian.com/event/feminism-14942366019 - I think Twitter must fully resolve the url, and then cache the metadata for a long time (permanently?). So probably have to wait until we publish a new event before we can see this change (unless there's an existing event that hasn't been tweeted about yet?!)

mattandrews:
There's probably a way to force Twitter to clear its cache for a given page's metadata; Facebook let you do it. Not sure how we'd log in to do it though, might need to be in control of the Guardian twitter account...?!

rtyley:
Thanks to the events team, got a brand new event to test against - [Freeing the Arctic Thirty](https://membership.theguardian.com/event/guardian-live-freeing-the-arctic-thirty-15598348079) - and it looks good:

gif_20150203_153522

PR #146, opened 2015-02-03
added prevent default to toggle

feedmypixel:
added a prevent default to toggle to stop screen moving when hash is in the toggle url when using an anchor for a toggle elem


davidrapson:
:+1:

wpf500:
:+1:

PR #134, opened 2015-01-28
Use a Membership-specific Ophan script that identifies us as our own platform

rtyley:
See https://github.com/guardian/ophan/pull/821 for the generation of the Membership-specifc Ophan script.

The 'ophan/ng' to just 'ophan' change is not strictly necessary, I just made it because 'ng' is what the Ophan project calls the www.theguardian.com (formerly NGW) platform.

cc @davidrapson

PR #153, opened 2015-02-04
change ophan module name to stop define() curl.js error

feedmypixel:
<h3>before</h3>
We have been seeing the following error when curl loads the Membership Ophan script. 
Note the Membership Ophan script loads.

screen shot 2015-02-04 at 12 05 12

<h3>after</h3>
Changing the module name to 'ophan/membership' 

<i>curl.js expects a named module to use its true name. There are no single-word module names in a package. All module names in packages take the form "[package-name]/[module-path]".<i> [ref](https://github.com/cujojs/curl/issues/256#issuecomment-35163242)

The error goes away and the Membership Ophan script loads. 

screen shot 2015-02-04 at 12 17 36

### DIFF
--- a/frontend/app/views/fragments/javaScriptLaterSteps.scala.html
+++ b/frontend/app/views/fragments/javaScriptLaterSteps.scala.html
@@ -8,7 +8,7 @@
         apiName: 'require',
         paths: {
             zxcvbn: '@Asset.at("javascripts/lib/zxcvbn/zxcvbn.js")',
-            ophan: '@Config.ophanJsUrl',
+            'ophan/membership': '@Config.ophanJsUrl',
             omniture: '@Asset.at("javascripts/lib/omniture/omniture.js")'
         }
     };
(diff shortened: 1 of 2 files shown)


rtyley:
:+1: just talked through this with @feedmypixel . To resummarise:

* we've always had _some_ `define()` errors in sentry against Ophan, since before I introduced PR #134
* @feedmypixel thinks those errors were probably device dependant
* With PR #134 we started getting those errors for EVERYBODY.
* the errors don't actually seem to stop us getting Ophan stats
* once this PR is in, we should have less noise and be able to diagnose device-specific problems

philwills:
AMD named module loading is such a needless minefield. :crying_cat_face:

rtyley:
This went out with TeamCity build 4479... let's see if the errors die down...

rtyley:
So far, looking better...

https://app.getsentry.com/the-guardian/membership/?query=duplicated

davidrapson:
@feedmypixel @rtyley Great work, glad to get rid of this noise.

@feedmypixel Is the named module issue that caused this just curl idiosyncrasy? Would be interesting to know if this would happen with the default requirejs module loader / almond.

PR #155, opened 2015-02-04
Updating to latest version of common, so that when I change common I know what broke it

rtyley:
Note this is the only change going 0.51 to 0.52: https://github.com/guardian/membership-common/commit/fc9f87d376118953c93a755e1ef5fa835a961c4a

cc @wpf500


wpf500:
I don't think you can do this, the `Tier.all` is used by the change tier page so it would try to show the supporter tier. Probably worth changing that first

rtyley:
Gonna recklessly fold this into https://github.com/guardian/membership-frontend/pull/158

PR #156, opened 2015-02-04
Rationalise action styles

davidrapson:
This was a bit of work to update the style of the secondary actions that turned into a more general cleanup.
Somehow we've ended up with a lot of duplicate styles for actions (weird merge maybe?) so this rationalises the types of action modifiers we have down to just what's required.

**Updated patterns showing all available actions**

screen shot 2015-02-04 at 14 50 28


frontend/app/views/fragments/tier/benefits.scala.html
```diff
@@ -55,9 +55,15 @@ <h2 class="tier-header__title">
                     <a class="action js-opt-become-member" href="@routes.Joiner.enterDetails(tier)" id="qa-@tier.slug" data-link-name="Become a @tier">Become a @tier</a>
                     @if(showLearnMore) {
                         @if(tier == Tier.Patron) {
```
feedmypixel: the only difference here is the href. Could this be just be surrounded by the if statement? (makes scales hand movement)

davidrapson: Definitely agree, but probably out of scope of the PR and this is part of the old benefits fragment which has many more issues and is about to go anyway.

frontend/assets/stylesheets/components/_action.scss
```diff
@@ -21,7 +21,7 @@
     cursor: pointer;
     border: 0;
     text-align: left;
-    min-width: rem(gs-span(2));
+    // min-width: rem(gs-span(2));
```
feedmypixel: comment needs removing?

feedmypixel:
:+1: LGTM

PR #160, opened 2015-02-05
Improvements to feedback form validation

feedmypixel:
- added (optional) marker to the optional "Which page on the web site are you referring to?" input
- added styles for textarea to add error border-styles to highlight when it has an error
- added email regex from [w3c](http://www.w3.org/TR/html-markup/input.email.html#input.email.attrs.value.single)

<h4>Before</h4>
 old

<h4>After</h4>
new

### DIFF
--- a/frontend/app/views/fragments/form/feedback.scala.html
+++ b/frontend/app/views/fragments/form/feedback.scala.html
@@ -64,7 +64,8 @@
             @fragments.form.errorMessage("Please enter your name")
         </div>
 
-        @* email regex taken from http://www.w3.org/TR/html-markup/input.email.html#input.email.attrs.value.single *@
+        @* //TODO this email check needs to be done on the server
+        email regex taken from http://www.w3.org/TR/html-markup/input.email.html#input.email.attrs.value.single *@
         <div class="form-field">
             <label class="label" for="email">Email address</label>
             <input name="email"
--- a/frontend/app/views/fragments/form/feedback.scala.html
+++ b/frontend/app/views/fragments/form/feedback.scala.html
@@ -64,12 +64,14 @@
             @fragments.form.errorMessage("Please enter your name")
         </div>
 
+        @* email regex taken from http://www.w3.org/TR/html-markup/input.email.html#input.email.attrs.value.single *@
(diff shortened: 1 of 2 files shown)


davidrapson:
:+1: LGTM

frontend/app/views/fragments/form/feedback.scala.html
```diff
@@ -64,12 +64,14 @@ <h2 class="fieldset__headline">About you</h2>
...
         </div>
 
+        @* email regex taken from http://www.w3.org/TR/html-markup/input.email.html#input.email.attrs.value.single *@
         <div class="form-field">
             <label class="label" for="email">Email address</label>
             <input name="email"
                    id="email"
                    class="input-text"
                    type="email"
+                   pattern="^[a-zA-Z0-9.!#$%&’*+/=?^_`{|}~-]+@@[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)*$"
```
wpf500: Not sure you need this pattern, most browsers validate emails anyway.

davidrapson: That's not quite true: `type=email` will give users an email specific keyboard but the actual validation part has not-quite-good-enough browser support http://caniuse.com/#search=form%20validation

wpf500: I don't think its worth having some obscure regex just to cover browser badness, also I would have thought if they don't support type validation then they won't support pattern anyway

davidrapson: The pattern is used in the JS, not as part of the native browser validation. So I think @feedmypixel  is using the pattern as a regex in the JS. Arguably this should be defined in the JS in one location though, rather than per-element.

feedmypixel: The pattern/regex work is to stop the blank white page when a non email string is added in the email input. 
After a little natter I'll add a //TODO and move this work server-side when I have some downtime. @wpf500 pointed out there is email check stuff we can use in scala hich feels a better place for it.

PR #162, opened 2015-02-05
Add lodash-amd

davidrapson:
Wanted to use some more functional helpers in another piece of work around improving client-side metrics. Noticed we had our own implementation of `extend` as well so have pulled in `lodash-amd` to avoid us re-implementing this kind of thing.

Main benefit of `lodash-amd` is that we can require just a single function so only the functions we actually use get included in the build.

Can also potentially replace `toArray` in helpers with the lodash version but going to do that as a separate PR.

// @feedmypixel


feedmypixel:
:+1: LGTM

PR #164, opened 2015-02-05
Improve client-side metric tracking

davidrapson:
- Adds a centralised component for tracking client-side metrics, decouples the usage from google analytics (although we still use this for the actual tracking).
- Introduces naming conventions for groups of events to improve reporting in Analytics
- Removes some noise by only tracking at key actions.
- Adds a couple of fragments to event pages to make sure there is only one location for primary actions (notably booking buttons)

There is a separate PR to come to abstracts this further and adds form tracking (errors, abandonment etc.)


davidrapson:
Mind giving this a once over @mattandrews

frontend/assets/javascripts/src/modules/metrics.js
```diff
@@ -0,0 +1,87 @@
...
+            var metricEvent = TRACKING_EVENTS[key];
+            if (metricEvent) {
+                metricEvent.call(this, elem);
+            }
+        });
+    }
+
+    function extractMetricsFromListener(elem, label, extras) {
+        if (elem && label) {
+            logMetric(extend({
```
mattandrews: In here, would it be worth checking for the existence of all the required attributes before logging the event? Just in case someone forgets/misspells one and we lose data?

mattandrews:
LGTM :+1: – great work as ever, looking forward to seeing the data from this!

frontend/assets/javascripts/src/modules/metrics.js
davidrapson: Good call, will add some better checking in here.

PR #165, opened 2015-02-05
fix for edge case month/year input bug

feedmypixel:
<h4>bug</h4>
There is a bug with the validation for month/year expiry in the new form work. If you have a valid year but not a valid month then there is an error recorded in the form but this error is not communicated in red to the user so you cant submit the form. The only way to get around this is to blur the month and year elements to flush the errors out.

<h4>fix</h4>
This pull request fixes this by treating the validation profile of the month/year inputs as a pair.
The change I have made means that if the month is reporting it is valid this then means that year is also valid so flush both month and year errors. Visa versa if the year is not valid then the month is also not valid so keep the errors.

### DIFF
--- a/frontend/assets/javascripts/src/modules/form/payment/validationProfiles.js
+++ b/frontend/assets/javascripts/src/modules/form/payment/validationProfiles.js
@@ -1,4 +1,7 @@
-define(['stripe'], function (stripe) {
+define([
+    'stripe',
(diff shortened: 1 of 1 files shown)


davidrapson:
:+1: Makes sense.

jamesoram:
:+1:

PR #166, opened 2015-02-06
Minor layout fix for patron page

davidrapson:
Spotted a minor layout issue with the Patron page. Joining panel was going full width, should be constrained

screen shot 2015-02-06 at 10 35 25

screen shot 2015-02-06 at 10 35 50


mattandrews:
:+1:

PR #168, opened 2015-02-06
Update cancellation summary to use functional page layout

davidrapson:
Update cancellation summary to use functional page layout

**New**
screen shot 2015-02-06 at 12 50 13

**Old**
screen shot 2015-02-06 at 12 50 54


mattandrews:
:+1:

PR #171, opened 2015-02-09
Cleanup event templates to use a single pricing fragment

davidrapson:
This PR cleans up events templates so there is once place where event pricing is defined, instead of three. This also helps @rtyley with some related refactoring work.

*Slightly* alters the format of pricing on list pages, but better to be consistent:

**New:**
screen shot 2015-02-09 at 15 04 13

**Old:**
screen shot 2015-02-09 at 15 04 00

// @rtyley


rtyley:
:+1: from the land of dealing with these tiny people:
cutty sark and cable cars - 15

davidrapson:
Adorable.

@jamesoram Mind giving this a once over. This changes updates how the pricing is included on the list pages and event detail pages.

jamesoram:
:+1:

PR #172, opened 2015-02-09
Use Salesforce to prefill address fields

wpf500:
Prefill addresses on upgrade using Salesforce and don't attempt to prefill join data (it almost certainly won't exist anyway)

This also means removing address line two as Salesforce doesn't have it.


jennysivapalan:
Do we have an approximate numbers on how many identity users have included their address? I understand pre-fill forms are a nice to have but it feels like we're taking this out and therefore could cause more drop off points for sign up.

wpf500:
I'm not sure, I think the assumption is that most of our Identity users have never given us an address anyway. Do we know how many people already have Identity accounts when they become a member?

jennysivapalan:
We don't separate the analytics tracking between sign in and registration to Identity. We should do this, I don't think it would be hard.

rtyley:
Closing, as @guardian/identity team have decided to keep addresses in Identity after all, rather than mastering in Salesforce.


> Do we have an approximate numbers on how many identity users have included their address? I understand pre-fill forms are a nice to have but it feels like we're taking this out and therefore could cause more drop off points for sign up.

It would have been a temporary degradation, according to [the plan](https://docs.google.com/a/guardian.co.uk/document/d/1_j55VtTlaj666YyIgWTWbw2gSuLXvnWqKlenixjKLrw/edit?usp=sharing) - but it's academic now.

PR #173, opened 2015-02-09
Tier fragments cleanup

davidrapson:
As the way we display benefits has been revised a few times we're left with a few bits of dead code. This PR does the following:

- Remove some unused benefits related fragments
- Remove some related dead styles
- Refactors the staff page slightly to avoid needing a one-time-use fragment for that page only.


mattandrews:
:+1:

PR #175, opened 2015-02-09
Replace masterclasses image on about page

davidrapson:
We've had a request to replace the image used for Masterclasses on the about page. This PR updates the page to use an alternative image.

screen shot 2015-02-09 at 16 55 35


feedmypixel:
:+1:

PR #176, opened 2015-02-10
Remove isBrochure flag

davidrapson:
This PR removes the need for the isBrochure flag and moves all layout constraints into individual templates (no global constraints on layout).

This frees us up to have much more flexible layouts, treating pages as vertical slices of content rather than a big rectangle to fill in.

Feels good to clean this out. This is a bit of a big PR, but I'm not sure there's a better way to do it other than dozens of identical PRs.

It touches nearly every template, but I've gone through each template to test that nothing has changed visually.

How do people feel about this? Better to rip the plaster off?

### DIFF
--- a/frontend/app/views/event/masterclass.scala.html
+++ b/frontend/app/views/event/masterclass.scala.html
@@ -7,7 +7,9 @@
 
 @import model.RichEvent.MasterclassEvent
 
-@main("Masterclasses", htmlClass="events-bg theme--masterclasses", pageInfo=pageInfo) {
+@main("Masterclasses", pageInfo=pageInfo, isBrochure=true) {
+
+    <main role="main" class="l-constrained">
 
     @fragments.event.headerBar(
         "Courses and workshops taught by award-winning professionals, with discounts for Partners and Patrons",
@@ -71,5 +73,7 @@
     </div>
 
     @views.html.event.list(eventPortfolio, "Sorry, no matching Masterclasses were found.", isFilterable=true)
+
+    </main>
 
 }
--- a/frontend/app/views/event/masterclass.scala.html
+++ b/frontend/app/views/event/masterclass.scala.html
@@ -7,7 +7,7 @@
 
 @import model.RichEvent.MasterclassEvent
 
-@main("Masterclasses", pageInfo=pageInfo, isBrochure=true) {
+@main("Masterclasses", pageInfo=pageInfo) {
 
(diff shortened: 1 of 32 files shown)


davidrapson:
Actually looking at change count / files changed it's not as bad as I thought it would be.

frontend/app/views/event/masterclass.scala.html
```diff
@@ -7,7 +7,9 @@
 
 @import model.RichEvent.MasterclassEvent
 
-@main("Masterclasses", htmlClass="events-bg theme--masterclasses", pageInfo=pageInfo) {
+@main("Masterclasses", pageInfo=pageInfo) {
+
+    <main role="main" class="l-constrained">
```
mattandrews: Being fussy here, but everything inside this `<main>` element needs to be indented one level.

mattandrews:
Gonna take it on faith that templates aren't broken – @jamesoram, you happy?

Also, worth adding anything on this stuff into the pattern library? Or is it there and I missed it...

mattandrews:
:+1: to the code changes, mind.

davidrapson:
@mattandrews Fairly confident this doesn't introduce any broken layouts, but definitely needs a good review from @jamesoram

jennysivapalan:
What did the isBrochure flag do?!

davidrapson:
This `isBrochure` flag previously removed a global wrapper which added a fixed width to the entire page.

Most templates now use an approach similar to guardian.com where templates are built up of individual layout slices, rather than one big wrapper so the flag is more hinderance than help.

davidrapson:
There is an alternative approach if people are not comfortable with removing it in one go, which is to slowly update all pages to use the `isBrochure` flag and then remove it completely once we're comfortable no pages actually need it.

I started the PR off this way but it felt that it was achievable to just make the change in one go.

mattandrews:
...I know that was a typo (or hope it was) but I like the anarchy of "slowly updat[ing] all pages to use the `isBrochure` flag" and then abruptly removing it. Let's do that instead. :black_joker:

davidrapson:
@mattandrews Not a typo, I just probably explained myself poorly… isBrochure is not needed as long as the templates have appropriate `l-constrained` classes in them.

jamesoram:
:+1: LGTM

davidrapson:
Going in…

PR #179, opened 2015-02-11
Use real events data for testing where possible

rtyley:
Some of the events used in the original `events.json` test data (which did come from eventbrite, just from our development account) didn't reflect real events that we work with - some didn't have any ticket classes at all
specified, for instance.

For some tests I've mapped events to their ids as I found it gave much less noisy test match-failure descriptions.


davidrapson:
Hold on, are you telling me I've missed Chris' big time jamboree?

rtyley:
> are you telling me I've missed Chris' big time jamboree?

You betcha.

PR #180, opened 2015-02-11
Remove a small bit of repetition on displaying pretty dates

rtyley:
Good to remove scala code from the templates where possible!


davidrapson:
:+1:

PR #183, opened 2015-02-11
Mark which tier the downgrade is from 

jennysivapalan:
For metrics to AWS we are marking all downgrades under Friend (which is the only possibility for downgrades). We could instead mark which tier the downgrade is coming from. @mattandrews this will change the radiator stats so let me know your thoughts on in if it is worth changing this. We could have two metrics (the one downgraded from and one downgraded to) but I'm hoping the Events tracking tool will cover this.


mattandrews:
I don't think the radiator currently shows downgrades anyway, just upgrades and cancellations, so this should be fine from that perspective...?

PR #184, opened 2015-02-11
Refactor ticketing model for InternalTicketing & DiscountBenefitTicketing

rtyley:
Some questions only make sense for certain events - eg "What's the member discount for this ticket?". Events can be broken down like this:

* **Internal** vs **External** ticketing - Russell Brannd cinema tickets were externally ticketed, by the cinemas involved - all other events, sold thru our Eventbrite, were _Internally_ ticketed.
* **Discounted Events** with a discount for certain tiers (Partner, Patron) - which covers most events - and events with _no_ discount, like the Charlie Hebdo charity event. We can _only_ display information around discounts for Internally ticketed events

There's a lot of conditional logic around how to display this ticket-pricing information, but rather than relying on devs surrounding the relevant code with 'if' statements, it's better to use optional fields and surround with for-comprehensions instead - this way, displaying `salesDates` for an externally-ticketed event can be enforced as a compile error - rather than relying on devs to remember to use the right `if` statement.

So now there's a nesting of optional fields, that goes like this:

```
event --> internalTicketing --> memberDiscountOpt
```

As an additional benefit of this work, we actually correctly display the discount offered based on consideration of both the general-release ticket and the members-benefit ticket - previously, **we were just assuming that the discount was indeed 20% all the time** - I'd rather the discount displayed correlates with what the user actually sees when they go to Eventbrite:

image

cc @jamesoram @jennysivapalan @davidrapson @feedmypixel

### DIFF
--- a/frontend/app/controllers/Event.scala
+++ b/frontend/app/controllers/Event.scala
@@ -1,25 +1,21 @@
 package controllers
 
+import actions.AnyMemberTierRequest
+import actions.Fallbacks._
+import actions.Functions._
+import com.gu.membership.salesforce.Tier
+import com.gu.membership.util.Timing
+import com.netaporter.uri.dsl._
+import configuration.{Config, CopyConfig}
 import model.Eventbrite.EBEvent
+import model.RichEvent._
+import model.{EventPortfolio, Eventbrite, PageInfo}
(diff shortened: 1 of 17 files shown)


jamesoram:
:+1:

feedmypixel:
:+1: great stuff, feels much cleaner

PR #187, opened 2015-02-11
Use inline icons for toggle buttons

davidrapson:
This PR updates toggle actions to use inline SVGs instead of needing to use `@extends`.

- Updates toggle actions to use inline SVGs
- Removes the need to have a dedicated `action--toggle` modifier
- New `data-toggle-icon` allows toggling of any icon, icon can be positioned anywhere and is not coupled to a specific button style.
- Moves `toggle.js` out of the `events/` sub-directory because it's not really related to events.

The approach is based on the fact that you can arbitrarily change the `xlink:href` attribute of any one of our inline icons to change it to a different icon in the set. This lets us do some fun stuff like this:

toggle-fun mov

Somebody please come up with a reason to use that second version.


davidrapson:
Don't worry everyone I found a much better use for the `xlink:href` trick:

color-wheel mov

frontend/assets/javascripts/src/modules/toggle.js
```diff
@@ -10,12 +10,14 @@
 define(['$', 'bean'], function ($, bean) {
 
     var TOGGLE_BTN_SELECTOR = '.js-toggle',
-        TOGGLE_DATA_ELM     = 'toggle',
-        TOGGLE_DATA_LABEL   = 'toggle-label',
-        TOGGLE_CLASS        = 'is-toggled',
-        ELEMENTS_TO_TOGGLE  = '[data-toggle-hidden]';
+        TOGGLE_LABEL_SELECTOR = '.js-toggle-label',
```
mattandrews: Booo, I love lining things up for no reason at all.

(this is fine, ignore me)

mattandrews:
This is clever. Nice work :+1:

PR #181, opened 2015-02-11
Paid to paid upgrades

davidrapson:
This PR is a combination of work from @jennysivapalan and @wpf500 enabling paid-to-paid upgrades and updates to the front-end to make this work. Front-end updates based on wires from @davidmcdowell 

**New paid-to-paid upgrade screen**

upgrade to patron

**Upgrade thank you**

upgrade to patron - thank you

PR #189, opened 2015-02-12
[DO NOT MERGE] Remove code to stop partners from upgrading from events

davidrapson:
**Dependant on PR** https://github.com/guardian/membership-frontend/pull/181

- Removes code to stop partners from upgrading from events
- Fixes a bug where the become a member / upgrade button didn't show in the header (button was only available in the sidebar)

Screenshots are from a test event which is only available to Patrons

**Signed out**

signed-out

**Signed in as partner**
partner

**Signed in as patron**
patron

Looking through this code, I think the approach in the JS could be simplified to only manipulate one button (rather than the two needed now) and to remove some of the coupling to specific tiers, possibly remove `component` too. It feels like there's a simpler way to reason about this.

This PR, however, makes the minimum changes to remove the upgrade restriction.

### DIFF
--- a/frontend/app/model/Eventbrite.scala
+++ b/frontend/app/model/Eventbrite.scala
@@ -135,6 +135,7 @@ object Eventbrite {
 
     val salesDates = TicketSaleDates.datesFor(eventTimes, primaryTicket)
 
+    val salesStart = primaryTicket.sales_start
     val salesEnd = primaryTicket.sales_end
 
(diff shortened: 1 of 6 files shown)


feedmypixel:
:+1: good tidy up

davidrapson:
This PR is superseded by https://github.com/guardian/membership-frontend/pull/247

PR #191, opened 2015-02-12
Events tracking - a starting point

jennysivapalan:
This PR sets up event tracking for Member events such as registration, upgrade, downgrade and cancel. There is a lot more things we want to track e.g Guardian events but we want to get this out to see if it will work in Prod. The data will push to https://github.com/guardian/snowplow which @dominickendrick has set up.

- https://github.com/t3hnar/scala-bcrypt is required to hash the ids 
- the postcode to the first part (eg N1, SW12)
- we'd like to push gender and age but this requires more integration with Identity so I have left this out for now.

Before merging we need to ensure the properties files have the relevant config.

@dominickendrick could you have a look through given @rtyley is away till next Friday? :pill:


frontend/app/services/TouchpointBackend.scala
```diff
@@ -70,6 +72,15 @@ case class TouchpointBackend(
       _ <- subscriptionService.downgradeSubscription(member, FriendTierPlan)
     } yield {
       memberRepository.metrics.putDowngrade(member.tier)
+      track(
+        MemberActivity(
+          "downgradeMembership",
+          MemberData(
+          member.salesforceContactId,
```
dominickendrick: Maybe we can just pass `member` in here, and add another `apply` method on `MemberData` to extract the correct data ?

frontend/app/tracking/ActivityTracking.scala
```diff
@@ -0,0 +1,118 @@
...
+      val dataMap = data.toMap
+      tracker.trackUnstructuredEvent(dataMap)
+    } catch {
+      case error: Throwable =>
+      Logger.error(s"Activity tracking error: ${error.getMessage}")
+    }
+  }
+
+  private def getTracker: Tracker = {
+    val emitter = new Emitter(ActivityTracking.url, HttpMethod.GET)
```
dominickendrick: We can enable the async client with `emitter.setRequestMethod(RequestMethod.Asynchronous)`

frontend/app/services/TouchpointBackend.scala
jennysivapalan: I did start with this but in some cases member just references the common trait MemberId which has just salesforce details. So other information has to come from Identity or the form.

frontend/app/tracking/ActivityTracking.scala
jennysivapalan: Oh yes sorry I forgot to try this out - will give it a go now.

frontend/app/services/TouchpointBackend.scala
dominickendrick: ah ok : )

project/Membership.scala
```diff
@@ -32,6 +32,7 @@ trait Membership {
     scalaVersion := "2.11.5",
     resolvers ++= Seq(
       "Guardian Github Releases" at "https://guardian.github.io/maven/repo-releases",
+      "Guardian Github Snapshots" at "http://guardian.github.com/maven/repo-snapshots",
```
rtyley: This extra resolver can probably be removed now in https://github.com/guardian/membership-frontend/pull/1757/files

PR #193, opened 2015-02-13
Remove unneeded action styles

davidrapson:
This PR removes some unneeded action styles. These classes don't add anything beyond the core set of actions.


feedmypixel:
:+1:

PR #194, opened 2015-02-13
Variable cleanup, move all tones over to tone map

davidrapson:
Variable cleanup, move all tones over to tone map. We had a slight split with some tone colours being traditional variables and the majority using a sass-map. This PR moves the remaining tone colours into the map.

This feels like a more readable way of describing our colour palette so I'll probably move the main colours into a map soon.


feedmypixel:
:+1:

PR #195, opened 2015-02-13
Remove some more dead svg icons

davidrapson:
This PR removes some more dead SVG icons. As an update we only have the following icons that are not using inline SVG:

- Brand icons on event list pages (will likely be moved to content `` tags)
- `g-white` and `arrow-white-small` are both used with the old benefits styles which is nearly gone…
- Remaining button arrow icons - These are only used in the primary actions. Secondary actions are already using inline icons, the primary actions need an update to get them to the same place.

Nearly there. Really pleased with the flexibility inline SVGs have given us so it's good to nearly be using them for 100% of icons.


feedmypixel:
:+1:

PR #196, opened 2015-02-13
Subs landing page

feedmypixel:
Work on the subscriptions offer landing page
- controller and routes created for ```/offer/subscriber```
- call out styles
- added offer images
- modifier added for slim version of functional layout
- fragments for offer Image and offer callout

## Desktop
subscribers the guardian membership

## Mobile
subscribers the guardian membership 1


frontend/app/controllers/Offer.scala
```diff
@@ -0,0 +1,11 @@
+package controllers
+
+import play.api.mvc.Controller
+
+trait Offer extends Controller {
+  def subscriber = NoCacheAction { implicit request =>
```
davidrapson: Does this need to be a `NoCacheAction`?

frontend/app/views/fragments/offer/callout.scala.html
```diff
@@ -0,0 +1,12 @@
+@(imgSrc: String, text: String = "")
+
+@import views.support.Asset
+
+<div class="offer-letterbox">
```
davidrapson: Is letterbox the ful-width images? I don't think this should be coupled to an offer, a generic letterbox component would be preferable.

frontend/app/views/fragments/offer/image.scala.html
```diff
@@ -0,0 +1,9 @@
+@(imgSrc: String)
+
+@import views.support.Asset
+
+<div class="promo-letterbox">
```
davidrapson: I think this should be combined with `offer/callout` into a generic letterbox component which takes an Option of a title and an option of a tone class

frontend/app/views/fragments/offer/option.scala.html
```diff
@@ -0,0 +1,10 @@
+@(title: String, offer: String, fullPrice: String, rightGutter: Boolean = false, note: String = "")
```
davidrapson: `note` should be an Option of a string, not an empty string. To keep @wpf500 happy

frontend/app/views/offer/subscriber.scala.html
```diff
@@ -0,0 +1,93 @@
...
+
+@main("Subscribers") {
+    @defining(Benefits.details(Tier.Partner)) { benefits =>
+        <main role="main">
+
+            <div class="page-slice page-slice--slim l-constrained l-side-margins">
+            @fragments.offer.callout("images/offer/subscriptions/audience-1140.jpg", "Pay nothing for 6 months")
+            </div>
+
+            <section class="tone-white l-gutter l-constrained">
```
davidrapson: Not 100% about the tone-white class it's coupled a little but to the colour. Could this not also be a `.page-slice`?

frontend/app/views/offer/subscriber.scala.html
```diff
@@ -0,0 +1,93 @@
...
+            </section>
+
+            <section class="tone-white tone-featured-tint l-gutter l-constrained">
+                <div class="page-section page-section--slim">
+                    <div class="page-section__content">
+                        <h2 class="h-feature">What's included</h2>
+                        <ul class="package-benefits grid grid--3up">
+                        @for(benefit <- benefits.list){
+                            <li class="grid__item">
+                                <div class="u-divider-dotted">@fragments.benefits.item(benefit)</div>
```
davidrapson: Nitpick but can `u-divider-dotted go on the parent `li` or does that break the layout?

frontend/app/views/offer/subscriber.scala.html
```diff
@@ -0,0 +1,93 @@
...
+                            </li>
+                        }
+                        </ul>
+                    </div>
+                    <div class="page-section__supplementary">
+                        <aside class="aside-feature" role="complementary">
+                            <p class="text-feature">
+                                Get closer to the stories that matter and experience the Guardian brought to life, with priority booking and discounted tickets</p>
+                            <div>
```
davidrapson: Any need for the naked `div` here?

frontend/app/views/offer/subscriber.scala.html
```diff
@@ -0,0 +1,93 @@
...
+                                <a class="action" href="@routes.Joiner.tierList">Join now</a>
+                            </div>
+                            <div>
```
davidrapson: Any need for the naked div here?

frontend/app/views/offer/subscriber.scala.html
```diff
@@ -0,0 +1,93 @@
...
+                                <a class="action action--secondary u-no-margin" href="@routes.Info.about">
+                                    More about Membership</a>
+                            </div>
+                        </aside>
+                    </div>
+                </div>
+            </section>
+
+            <section class="tone-white l-gutter l-constrained">
```
davidrapson: Again should probably still be a `page-slice` rather than `tone-white`

frontend/app/views/offer/subscriber.scala.html
```diff
@@ -0,0 +1,93 @@
...
+                <div class="page-section page-section--slim">
+                    <div class="page-section__content">
+                        <h2 class="h-feature">Subscriber offer</h2>
+                        @for(pricing <- benefits.pricing) {
+                            @fragments.offer.option(
+                                title="Pay annually",
```
davidrapson: Rather than having the `rightGutter` option, is there an oportunity to to use the grid component here?

frontend/app/views/offer/subscriber.scala.html
```diff
@@ -0,0 +1,93 @@
...
+                                note=s"equivalent to ${pricing.yearlyMonthlyCost.pretty}/year"
+                            )
+                        }
+                        <a class="action u-no-bottom-margin" href="@routes.Joiner.tierList">Join now</a>
+                    </div>
+                </div>
+            </section>
+
+            <div class="page-slice page-slice--slim l-constrained l-side-margins">
+            @fragments.offer.image("images/offer/subscriptions/guardian-window-1140.jpg")
```
davidrapson: Mentioned above, but it would be cool to make this fragment more generic.

frontend/app/views/offer/subscriber.scala.html
```diff
@@ -0,0 +1,93 @@
...
+            </div>
+
+            <section class="tone-white l-gutter l-constrained">
```
davidrapson: `.page-slice`?

frontend/assets/stylesheets/components/_asides.scss
```diff
@@ -44,3 +44,9 @@
     padding: rem($gs-baseline * 2);
     margin-bottom: rem($gs-baseline * 2);
 }
+
+.aside-feature {
+    @include mq(tablet) {
+        margin-top: rem(62px);
```
davidrapson: :trollface:

frontend/assets/stylesheets/components/_callout.scss
```diff
@@ -0,0 +1,39 @@
+.call-out-container {
```
davidrapson: This should be the root component if `.call-out` is dependant on that container (`.call-out` becomes `.call-out__content`)

frontend/assets/stylesheets/layout/_layout.scss
```diff
@@ -56,6 +56,14 @@ body {
     }
 }
 
+.l-gutter {
+    padding: rem($gs-baseline) rem(20px) rem($gs-baseline*3);
```
davidrapson: 20px?

frontend/assets/stylesheets/components/_callout.scss
davidrapson: Picky but could we add a block comment at the top of this file describing what it does, like other components?

frontend/assets/stylesheets/theme/_tones.scss
```diff
@@ -178,6 +178,12 @@
...
    ========================================================================== */
 
+.tone-white {
+    background-color: $white;
+}
 .tone-tint {
     background-color: $c-neutral7;
 }
+.tone-featured-tint {
+    background-color: $c-neutral6;
```
davidrapson: Should this not be the same grey as the tone above? We should change `.tone-tint` to be one or the other across the board.

frontend/conf/routes
```diff
@@ -86,6 +86,9 @@ POST           /feedback                          controllers.Info.submitFeedbac
 GET            /about                             controllers.Info.about
 GET            /patrons                           controllers.Info.patron
 
+# Offer
+GET            /offer/subscriber                  controllers.Offer.subscriber
```
davidrapson: As discussed, I'm not a huge fan of this kind of URL structure but it's consistent with the rest of the app so ¯\(°_o)/¯

davidrapson:
:+1: Lots of comments, sorry! This is really good, but probably worth chatting through a few bits of the CSS as theres a couple of things I think could be more abstract.

frontend/app/views/fragments/offer/option.scala.html
davidrapson: @wpf500 Gone but not forgotten

frontend/app/controllers/Offer.scala
jennysivapalan: I think it should be a `CachedAction`, there isn't anything on this page that requires user input.

frontend/app/views/fragments/offer/option.scala.html
jennysivapalan: :+1: Option is better here - we can then use a `map` rather an `if` condition below.

frontend/app/views/offer/subscriber.scala.html
feedmypixel: As discussed there are override issues with the grid component adding it on this element stops having to override this

frontend/app/views/offer/subscriber.scala.html
feedmypixel: to make the anchor behave like a block element

frontend/app/views/offer/subscriber.scala.html
feedmypixel: to make the anchor behave like a block element

frontend/conf/routes
feedmypixel: understood - this url needs a once over from @davidmcdowell

feedmypixel:
> Lots of comments, sorry! This is really good, but probably worth chatting through a few bits of the CSS as theres a couple of things I think could be more abstract.

no - probs thanks for input. Please have a check over with changes in now.

frontend/assets/stylesheets/components/_asides.scss
davidrapson: As discussed I think we can get rid of this by introducing a `page-section__header`

frontend/assets/stylesheets/layout/_layout.scss
```diff
@@ -56,6 +56,14 @@ body {
     }
 }
 
+.l-gutter {
+    padding: rem($gs-baseline) rem($gs-gutter) rem($gs-baseline*2);
```
davidrapson: Still not quite sure about this, given the name it feels like this should only be adding gutters (e.g., left/right padding). Could this be renamed perhaps?

davidrapson:
Aside from the one last comment about `aside-feature` this looks really good.

feedmypixel:
AS a side note this work has been placed up on Staff Oauth to show internally - Once everything is finished the staff Oauth can be removed

PR #201, opened 2015-02-16
Add PostCSS/Autoprefixer

davidrapson:
Following on from conversations in PR https://github.com/guardian/membership-frontend/pull/104 and https://github.com/guardian/membership-frontend/pull/89 where we've been caught out by vendor prefixes, I've added [autoprefixer](https://github.com/postcss/autoprefixer) support to the Gruntfile.

Main things autoprefixer does for out stylesheet:

- Unnecessary prefixes from box-sizing, box-shadow and border-radius (can all be un-prefixed now)
- Automatically outputs needed flex-box prefixes based on our target browsers.

Flex-box is the one that needs the most consideration as it has the largest no. of prefixes.

We can configure the list of supported browsers but I've gone with: Browsers which have > 5% global usage, last 2 versions of all browsers and then explicitly included IE 9 and Safari 6.

Diff showing pre vs. post autoprefixer on the main stylesheet here: https://gist.github.com/davidrapson/fde36036e383de093fb3


feedmypixel:
:+1:

davidrapson:
@jamesoram Mind giving this a once over? Particularly interested in the change tier page to make sure the flexbox layout doesn't break with this change.

mattandrews:
I'd like to make sure this doesn't break Safari 6 or Firefox 19 as these have been problematic for us in the past, albeit with minor audiences.

Also, if we're adding this, should there be a corresponding CSS change to remove the existing prefixes so there's only one source of truth?

davidrapson:
"I'd like to make sure this doesn't break Safari 6 or Firefox 19". Agreed, partly why I included Safari 6 explicitly.

There probably should be an existing CSS change to remove unneeded prefixes, but that's part of what autoprefixer handles for us anyway so I can be a later PR.

feedmypixel:
> Also, if we're adding this, should there be a corresponding CSS change to remove the existing prefixes so there's only one source of truth?

Some of these are found in guss-css3 so maybe worth having a chat with the .com team to see if support is wanted to be dropped for these prefixes or let autoprefixer handle it.

davidrapson:
@feedmypixel I think that would be a little odd, all `guss-css` does is *add* prefixes, so autoprefixer just means we don't need to use guss-csss for Membership.

davidrapson:
@feedmypixel I think that would be a little odd, all `guss-css` does is *add* prefixes, so autoprefixer just means we don't need to use guss-csss for Membership.

feedmypixel:
Good point

davidrapson:
@jamesoram Mind giving this a once over. In particular do you have an older version of Firefox you can test on?.

davidrapson:
For reference:

**Safari 6 w/Autoprefixer**
davidspc

**Safari 6 live**
membershiponsafari

Default fonts is a limitation of the local URL, it can't load the fonts. The nav bug in the screenshot is present on live too so doesn't seem to be related to autoprefixer. So Safari 6 is definitely good.

jamesoram:
Do we mind having the cancel button at the top of the cancel page in IE9?

screen shot 2015-02-20 at 14 44 33 2

davidrapson:
Hmm, I think so for now. I'm literally updating these pages as we speak anyway.

davidrapson:
@jamesoram Could you check if this page looks like this on live too?

jamesoram:
it's the same on prod

davidrapson:
@jamesoram In that case I think we are OK to leave it as is for now. As long as it's not different to PROD.

jamesoram:
:+1:

PR #202, opened 2015-02-16
tweak to month expiry validation

feedmypixel:
### A tweak on the errors shown for the expiry date month validationProfile validation on our forms.

Currently the expiry date is validated when the ```blur/change``` listener is fired on the month select. This work fixes that so the ```blur/change``` listener on the month select will only attempt to validate if the year select has a number as its value. Other than this the validation works as previously.

#### Before
before-error

#### After
after-error

### DIFF
--- a/frontend/assets/javascripts/src/modules/form/payment/validationProfiles.js
+++ b/frontend/assets/javascripts/src/modules/form/payment/validationProfiles.js
@@ -31,11 +31,16 @@ define([
      */
     var validCreditCardMonth = function (monthElem) {
         var yearElem = document.querySelector('.js-credit-card-exp-year');
-        var isValid = stripe.card.validateExpiry(monthElem.value, yearElem.value);
+        var isValid = true;
 
-        if (isValid) {
-            // treat month/year inputs as a pair if month validates both month and year are valid so flush errors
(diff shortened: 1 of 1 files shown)


feedmypixel:
@jamesoram can you have a look over thanks

frontend/assets/javascripts/src/modules/form/payment/validationProfiles.js
```diff
@@ -31,11 +31,16 @@ define([
...
     var validCreditCardMonth = function (monthElem) {
         var yearElem = document.querySelector('.js-credit-card-exp-year');
-        var isValid = stripe.card.validateExpiry(monthElem.value, yearElem.value);
+        var isValid = true;
 
-        if (isValid) {
-            // treat month/year inputs as a pair if month validates both month and year are valid so flush errors
-            display.flushErrIds([monthElem.id, yearElem.id]);
+        // we only want to validate expiry if the year select has a number value i.e not the default value
+        if (!isNaN(parseInt(yearElem.value, 10))) {
```
davidrapson: Oh JavaScript (this is fine, just JS is funny sometimes)

davidrapson:
:+1:

PR #206, opened 2015-02-17
Use dummy config for salt and pepper

jennysivapalan:
We want to check the data we sent to activity tracking is in the correct format. Part of this uses bcrypt so I've created dummy config for dev environment. This should hopefully make Team city happy.


mattandrews:
salt-n-pepa-510f00199631f

Push it! :+1:

PR #207, opened 2015-02-17
Add masterclass overview page

mattandrews:
Might as well add this too.

### DIFF
--- a/frontend/app/views/staff/eventOverview.scala.html
+++ b/frontend/app/views/staff/eventOverview.scala.html
@@ -1,4 +1,8 @@
-@(liveEvents: Seq[model.RichEvent.RichEvent], draftEvents: Seq[model.RichEvent.RichEvent])
+@(
+    liveEvents: Seq[model.RichEvent.RichEvent],
+    draftEvents: Seq[model.RichEvent.RichEvent],
+    previewType: String = "event"
+)
 
 @import model.Eventbrite.InternalTicketing
 @import model.Eventbrite.ExternalTicketing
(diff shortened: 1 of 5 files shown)


mattandrews:
screen shot 2015-02-17 at 12 29 47

jennysivapalan:
:+1: :rocket:

mattandrews:
Okay with you @feedmypixel?

feedmypixel:
Awesome - I'll let you mail back Jessica with this -  "you asked and we built it" :) 
- Id add a conditional to not show the preview and eventbrite links as the wont work for Masterclasses right? 
- Not all the status flags will show but that will be silent so don't worry about that. 
- The draft functionality - is that just empty? Maybe also don't include this section for Masterclasses?

:+1: great stuff

feedmypixel:
Also https://github.com/guardian/membership-frontend/pull/210 has been added to make the OAuth staff not just perm staff so you will need to update that here also

mattandrews:
Good point @feedmypixel – I've added preview support for Masterclasses with help from @jennysivapalan.

The EventBrite links work fine though, you just need to have access to that account.

mattandrews:
Oh, and it shows drafts too!

feedmypixel:
:8ball:

frontend/app/views/staff/eventOverview.scala.html
```diff
@@ -1,4 +1,8 @@
-@(liveEvents: Seq[model.RichEvent.RichEvent], draftEvents: Seq[model.RichEvent.RichEvent])
+@(
+    liveEvents: Seq[model.RichEvent.RichEvent],
+    draftEvents: Seq[model.RichEvent.RichEvent],
+    previewType: String = "event"
```
feedmypixel: can we work this previewType out from doing a match on the event type?

PR #209, opened 2015-02-17
Make feedback thank you inline

feedmypixel:
- made the feedback message inline using the flash message work rather than redirecting to another page
cvcvc


davidrapson:
:+1:

PR #210, opened 2015-02-17
Change OAuth on event-overview page

feedmypixel:
Change the event overview page just to google oauth for GU staff no need for it to be permenant staff members only


jennysivapalan:
:+1: :calendar:

feedmypixel:
:metal: :octopus:

PR #215, opened 2015-02-18
Update to package.json to fix build issues in Grunt

feedmypixel:
- Update all package.json dependencies that can be
- Fix a JsHint error that showed us having 2 metric imports in main.js (I suspect this is not to do with the build failure just a side effect of updating JSHint)


davidrapson:
:+1:

PR #217, opened 2015-02-18
Link to Polly Toynbee article from the homepage

davidrapson:
Spoke to Julia and agreed that we should be at least linking to the Polly Toynbee article from the homepage. We're also going to review the copy on the about page to see if we can replace something with a quote from that article, most likely the "Founding members: join us at the start" section.

screen shot 2015-02-18 at 14 10 21


jennysivapalan:
:+1: :rainbow:

PR #219, opened 2015-02-18
Event overview tweaks

feedmypixel:
Bit of a tweak on the event overview work.
- separate Guardian Live and Masterclass work into separate fragments
- Create different text for masterclass and gulive event
- split out overviewSection into its own fragment
- define the preview Url Sub Category for differing preview urls on masterclass and event overview pages


feedmypixel:
@mattandrews thoughts?

mattandrews:
Thanks for this!

Personally not convinced on having a file each for `eventOverview/masterclass.scala.html` and `eventOverview/guLive.scala.html` as they're essentially the same besides substituting "Guardian Live" for "Masterclass". 

But if this removes the fairly hacky logic then go for it :)

feedmypixel:
Yeah I think the removal of logic sounds good. 
It feels right although you could do an overView fragment which you just sent in a title and description - Although this does feel like a <strike>bridge</strike> fragment too far!

davidrapson:
:shipit: I think this looks OK, we can probably consider this in more detail when if we decide to expand these tool pages.

mattandrews:
Yeah, let's ship it! :+1:

feedmypixel:
@mattandrews  - want to reply to Jessica and let her know this is available in ~10 mins

PR #225, opened 2015-02-19
Simplify outdated browser message

davidrapson:
This PR proposes simplifying the outdated browser message. My thinking is:

- IE8 and below already only show a HTML only page due to our HTTPS/SSL requirements
- The unsupported message is really just about JavaScript support and not display, so they may still be able to view the website content. Hiding the entire website seems quite aggressive in this case.
- As < IE8 is excluded the people who actually see this are mobile users, where there could easily be some combination of support where they don't cut the mustard for JS but still have good CSS support. Again hiding the whole site seems a bit much.

This update replaces the large message with a simple banner like the no-JavaScript message. Wording is taken from HTML5 Boilerplate.

screen shot 2015-02-19 at 14 20 42


davidrapson:
Will happily expand this message if it's felt we want to keep the contact details in this message.

mattandrews:
I can't remember the initial argument for making the site unusable for non-cutting-the-mustard browsers. Might've been for layout reasons but dunno.

feedmypixel:
> I can't remember the initial argument for making the site unusable for non-cutting-the-mustard browsers. Might've been for layout reasons but dunno.

I think the reason for this was before launch and the amount of resources and time we had

davidrapson:
@feedmypixel Yep, spoke to Matt at the end of last week and I think we we're both happy that it was OK to tone down the unsupported browser stuff, so AFAIK this is good to go.

feedmypixel:
:+1:

feedmypixel:
@wpf500 would be proud
screen shot 2015-02-24 at 10 44 44

wpf500:
:+1:

PR #226, opened 2015-02-19
Only show series details for Guardian-only events

davidrapson:
Spotted an issue with partner events where we show the partner logo next to a description for guardian live.

screen shot 2015-02-19 at 16 28 06

Long term we should come up with a way of showing a custom description for partners, but in the short term this PR removes the detail panel for partner events.

screen shot 2015-02-19 at 17 20 17

// @mattandrews


mattandrews:
:+1:

PR #227, opened 2015-02-19
Fix issue with being able to click full link to become a partner

davidrapson:
Spotted an issue where the button was correctly disabled for partner upgrades but you could still click the tier name (used to increase hit area on mobile) to become a partner. This PR disables that action.


feedmypixel:
:+1:

PR #228, opened 2015-02-20
Warn on bad discounts on the event overview

rtyley:
This adds warnings for these mis-configurations:

* Discount for Guardian Member tickets not being 20%(eg '**36% Discount**')
* Fewer Guardian Member tickets than General Release tickets (not an error on Masterclasses, but wrong for Guardian Live events) - '**Fewer discount tickets than general**'

image

https://membership.theguardian.com/staff/event-overview
https://membership.theguardian.com/staff/masterclass-overview

cc @feedmypixel

### DIFF
--- a/frontend/app/configuration/Config.scala
+++ b/frontend/app/configuration/Config.scala
@@ -165,6 +165,8 @@ object Config {
 
   val discountMultiplier = config.getDouble("event.discountMultiplier")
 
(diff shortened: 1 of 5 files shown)


feedmypixel:
:+1: good thing to see 
- it would be good to inform downstairs this flag has been added search "Event Overview tool" in your inbox and reply to that thread.

PR #231, opened 2015-02-20
Event list powered by two accounts

feedmypixel:
- DiscoverEventService added
- DiscoverEvent case class in RichEvent added

cc @jennysivapalan @rtyley 

## Event list with displaying events from 2 accounts
events the guardian membership

## Discover Event
event detail member pre-sale event the guardian membership


frontend/app/controllers/Event.scala
```diff
@@ -85,13 +87,31 @@ trait Event extends Controller {
...
       Some(CopyConfig.copyDescriptionEvents)
     )
-    Ok(views.html.event.guardianLive(guLiveEvents.getEventPortfolio, pageInfo))
+    val pastEvents = discoverEvents.getEventPortfolio.pastEvents.fold {
+      guLiveEvents.getEventPortfolio.pastEvents
+    } { discoverPastEvents =>
+      guLiveEvents.getEventPortfolio.pastEvents.fold(Some(discoverPastEvents)) { guPastEvents =>
+        Some(chronologicalSort(guPastEvents ++ discoverPastEvents))
+      }
+    }
```
feedmypixel: This combination of options feels a little awkward

feedmypixel:
After discussions with @jennysivapalan on friday there has been some tidy up. The getEventPortfolio method is no longer needed and the work around combining two Option[Seq[RichEvent]] has been simplified.

feedmypixel:
Something I have noticed is the archived events do not show due to https://github.com/guardian/membership-frontend/blob/master/frontend/app/views/event/list.scala.html#L90 do we need to have a look at increasing this and then a look at pagination as Eventbrite returns 50 events per page? all thoughts for future PR

frontend/app/controllers/Event.scala
```diff
@@ -85,13 +87,24 @@ trait Event extends Controller {
     ))
   }
 
+  private def chronologicalSort(events: Seq[model.RichEvent.RichEvent]) = {
```
jennysivapalan: You could move this as an inner private method in `def list`

frontend/app/controllers/Event.scala
```diff
@@ -85,13 +87,24 @@ trait Event extends Controller {
...
       CopyConfig.copyTitleEvents,
       request.path,
       Some(CopyConfig.copyDescriptionEvents)
     )
-    Ok(views.html.event.guardianLive(guLiveEvents.getEventPortfolio, pageInfo))
+    Ok(views.html.event.guardianLive(
+      EventPortfolio(
+        guLiveEvents.getFeaturedEvents,
+        chronologicalSort(guLiveEvents.getEvents ++ discoverEvents.getEvents),
+        (guLiveEvents.getEventsArchive ++ discoverEvents.getEventsArchive).headOption.map(chronologicalSort(_).reverse),
```
jennysivapalan: Might be worth pulling this out as a `val` as it's doing quite a lot and the `reverse` it hidden from this view.

frontend/app/model/RichEvent.scala
```diff
@@ -133,6 +133,43 @@ object RichEvent {
...
+    } yield assetOpt
+
+    val availableWidths = widths.mkString(",")
+
+    val imageMetadata = image.map(_.metadata)
+
+    val socialImgUrl = image.flatMap(_.assets.find(_.dimensions.width == widths.max)).fold(fallbackImage){ asset =>
+      asset.secureUrl.getOrElse(asset.file)
+    }
+
```
jennysivapalan: Alot of these methods are duplicates of GuLiveEventsService - I wonder if it's worth extending that and overriding the specific methods?

frontend/app/services/EventbriteService.scala
```diff
@@ -153,6 +144,33 @@ object GuardianLiveEventService extends EventbriteService {
...
+
+  val gridService = GridService(Config.gridConfig.url)
+  val contentApiService = GuardianContentService
+
+  def mkRichEvent(event: EBEvent): Future[RichEvent] = {
+
+    val eventbriteContent = contentApiService.content(event.id)
+
+    event.mainImageUrl.fold(Future.successful(DiscoverEvent(event, None, eventbriteContent))) { url =>
+      gridService.getRequestedCrop(url).map(DiscoverEvent(event, _, eventbriteContent))
```
jennysivapalan: Are Discover events going to be using the Grid? Fine if it is but if not we could remove the getRequestCrop method.

frontend/app/services/EventbriteService.scala
```diff
@@ -207,6 +227,10 @@ object EventbriteService {
     GuardianLiveEventService.getPreviewEvent(id)
   }
 
+  def getPreviewDiscoverEvent(id: String): Future[RichEvent] = Cache.getOrElse[Future[RichEvent]](s"preview-event-$id", 2) {
+    DiscoverEventService.getPreviewEvent(id)
+  }
+
   def getPreviewMasterclass(id: String): Future[RichEvent] = Cache.getOrElse[Future[RichEvent]](s"preview-event-$id", 2) {
```
jennysivapalan: If we have a preview method you'll need to add it to the routes file and Event controller.

frontend/app/services/EventbriteService.scala
feedmypixel: to confirm they will be also using the grid

frontend/app/services/EventbriteService.scala
feedmypixel: I will take this out for the moment and do this as a separate pull request when adding in discover events to the event-overview tool

frontend/app/controllers/Event.scala
feedmypixel: this is also needed below in the method listFilteredBy

feedmypixel:
kudos to @jennysivapalan for pairing on this PR

frontend/app/controllers/Event.scala
```diff
@@ -70,7 +72,7 @@ trait Event extends Controller with ActivityTracking {
       request.path,
       Some(CopyConfig.copyDescriptionMasterclasses)
     )
-    Ok(views.html.event.masterclass(masterclassEvents.getEventPortfolio, pageInfo))
+    Ok(views.html.event.masterclass(EventPortfolio(Nil, masterclassEvents.events, None, None), pageInfo))
```
davidrapson: Personal preference, but I quite like named parameters for this kind of thing.

frontend/app/tracking/ActivityTracking.scala
```diff
@@ -101,6 +101,7 @@ case class EventData(event: RichEvent) {
 
   val group = event match {
```
rtyley: This (already existing) `match` statement is a little sad, you would have thought these strings could live within the Event class, or better, into the refactored EventProvider class that I keep talking about but never writing. Can let it go for the moment.

rtyley:
:+1:

feedmypixel:
@jamesoram please have a good look over

jamesoram:
I'm getting a 500 on /event/<id>/buy

screen shot 2015-02-26 at 10 54 15

jamesoram:
:+1: good job

PR #232, opened 2015-02-20
Update designs for downgrade and cancellation page

davidrapson:
Paired with @davidmcdowell to clean up the downgrade and cancellation pages based on our existing design patterns as they were both looking a bit worse for wear.

Now that these have been updated we can also remove the old mammoth benefits fragment!

**Cancel - Paid**
cancel - paid

**Cancel - Friend**
cancel - friend

**Downgrade**
downgrade

// @jamesoram Let me know if you need any new hooks. I've tried to keep all old ones but worth checking as your tests might need updating.


davidrapson:
The before version just for comparison…

before

jamesoram:
:+1:

mattandrews:
:+1:

PR #237, opened 2015-02-23
Revert "Disable Eventbrite healthcheck, in order to bring rest of site u...

rtyley:
Reverts guardian/membership-frontend#235 - otherwise deploys will lead to the site showing no events while the servers warm up. The healthcheck itself should pass now that we've cleared the bad data. 

Cc @jennysivapalan @JuliaBellis @davidrapson

### DIFF
--- a/frontend/app/controllers/Healthcheck.scala
+++ b/frontend/app/controllers/Healthcheck.scala
@@ -11,7 +11,7 @@ case class Test(name: String, result: () => Boolean)
 object Healthcheck extends Controller {
 
   val tests = Seq(
-    // Test("Events", GuardianLiveEventService.events.nonEmpty _),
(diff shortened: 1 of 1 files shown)


jennysivapalan:
Do we want to leave this healthcheck disabled until we fix the issue with bad data (in case it happens again today)? Alternatively we can remind the use of the preview link (I think the Events team are using it as I've seen requests for it)? I'm in training all day so can't do much today to help on Membership.

rtyley:
> Do we want to leave this healthcheck disabled until we fix the issue with bad data (in case it happens again today)?

Assuming the team are going to be deploying at all, every deploy is going to make the Guardian Live events screw up for a couple of minutes at a time, unless we merge this in.

PR #238, opened 2015-02-23
Code to fix the main-image parsing problem

rtyley:
Unfortunately, this PR omits the most important part of the pull request... the failing test, I just want to save @mattandrews some work.

### DIFF
--- a/frontend/app/model/Eventbrite.scala
+++ b/frontend/app/model/Eventbrite.scala
@@ -7,10 +7,13 @@ import com.netaporter.uri.dsl._
 import configuration.Config
 import org.joda.time.Instant
 import org.joda.time.format.ISODateTimeFormat
+import play.api.Logger
 import play.api.data.validation.ValidationError
 import play.api.libs.functional.syntax._
 import play.api.libs.json._
 import utils.StringUtils._
+
+import scala.util.{Failure, Success, Try}
 
 object Eventbrite {
 
@@ -201,7 +204,18 @@ object Eventbrite {
       if EBEvent.providerWhitelist.contains(provider)
     } yield provider
 
-    val mainImageUrl: Option[String] = description.flatMap(desc => "<!--\\s*main-image: (.*?)\\s*-->".r.findFirstMatchIn(desc.html).map(_.group(1)) )
+    val mainImageUrl: Option[Uri] = for {
+      desc <- description
+      m <- """\smain-image:\s*(.*?)\s""".r.findFirstMatchIn(desc.html)
+      uri <- Try(Uri.parse(m.group(1))) match {
+        case Success(uri) =>
+          Logger.debug(s"Event $id main-image source url: $uri")
+          Some(uri)
+        case Failure(e) =>
+          Logger.error(s"Event $id - can't parse main-image url from text '${m.matched}'", e)
+          None
+      }
+    } yield uri
 
     val slug = slugify(name.text) + "-" + id
 
--- a/frontend/app/model/Eventbrite.scala
+++ b/frontend/app/model/Eventbrite.scala
@@ -208,9 +208,7 @@ object Eventbrite {
       desc <- description
       m <- """\smain-image:\s*(.*?)\s""".r.findFirstMatchIn(desc.html)
       uri <- Try(Uri.parse(m.group(1))) match {
-        case Success(uri) =>
-          Logger.debug(s"Event $id main-image source url: $uri")
-          Some(uri)
+        case Success(uri) => Some(uri)
         case Failure(e) =>
           Logger.error(s"Event $id - can't parse main-image url from text '${m.matched}'", e)
           None
--- a/frontend/test/model/EBEventTest.scala
+++ b/frontend/test/model/EBEventTest.scala
@@ -1,5 +1,6 @@
 package model
 
+import com.netaporter.uri.Uri
 import model.Eventbrite._
 import play.api.test.PlaySpecification
 import org.joda.time.{ Instant, DateTime }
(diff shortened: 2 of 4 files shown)


mattandrews:
Cool, thanks!

mattandrews:
This doesn't pass tests for me locally, complains about `EBDes` not being defined.

mattandrews:
(I added the missing test, though!)

rtyley:
> This doesn't pass tests for me locally, complains about EBDes not being defined.

Garg, that was the test which I half-wrote before signing off - sorry:

https://github.com/guardian/membership-frontend/blob/9fbf02/frontend/test/model/EBEventTest.scala#L61-L63

mattandrews:
Yep, I could see where the error came from, but no idea what you were trying to do – copy a different event?

frontend/app/model/Eventbrite.scala
```diff
@@ -201,7 +203,11 @@ object Eventbrite {
       if EBEvent.providerWhitelist.contains(provider)
     } yield provider
 
-    val mainImageUrl: Option[String] = description.flatMap(desc => "<!--\\s*main-image: (.*?)\\s*-->".r.findFirstMatchIn(desc.html).map(_.group(1)) )
+    val mainImageUrl: Option[Uri] = for {
+      desc <- description
+      m <- """\smain-image:\s*(.*?)\s""".r.findFirstMatchIn(desc.html)
+      uri <- Try(Uri.parse(m.group(1))).toOption
```
jennysivapalan: Can we log the failure so we can spot this in the production.

frontend/test/model/EBEventTest.scala
```diff
@@ -55,8 +57,13 @@ class EBEventTest extends PlaySpecification {
       ebDraftEvent.isBookable mustEqual(false)
     }
     "should return media service url if present" in {
-      ebLiveEvent.mainImageUrl mustEqual Some("https://some-media-tool.co.uk/images/sdf8u8sdf898hnsdcvs89dc?crop=0_3_480_288")
-
+      ebLiveEvent.mainImageUrl mustEqual Some(Uri.parse("https://some-media-tool.co.uk/images/sdf8u8sdf898hnsdcvs89dc?crop=0_3_480_288"))
+    }
+    "should tolerate additional characters after the media service url" in {
+      ebLiveEvent.copy(description = Some(EBDes)).mainImageUrl mustEqual Some(Uri.parse("https://some-media-tool.co.uk/images/sdf8u8sdf898hnsdcvs89dc?crop=0_3_480_288"))
```
jennysivapalan: Where is EBDes defined?

frontend/test/model/EBEventTest.scala
```diff
@@ -55,8 +57,13 @@ class EBEventTest extends PlaySpecification {
...
+    }
+    "should not fail parsing if the HTML comments are malformed" in {
+      malformedHtmlEvent.mainImageUrl mustEqual Some(Uri.parse("https://some-media-tool.co.uk/images/sdf8u8sdf898hnsdcvs89dc?crop=0_3_480_288"))
```
jennysivapalan: I think rather than parsing the malformed comment we should fail on the single event and log it?

frontend/test/model/EBEventTest.scala
rtyley: Removed these errant lines with 58241a8 :blush:

rtyley:
I've squashed the commits for this into a single commit, and tweaked the tests so that it's a bit clearer to see what we do for a given url:

https://github.com/guardian/membership-frontend/blob/df37629/frontend/test/model/EBEventTest.scala#L63-L70

I think we should never kill an event, just because it's image is missing or unparsable- and we can be tolerant when we encounter something like:

```
<!-- main-image: https://media.gutools.co.uk/images/e5b35b1f20588172b24960314cbf8e9c8482d3bf?crop=27_0_1800_1080 ! -->
```

frontend/app/model/Eventbrite.scala
```diff
@@ -201,7 +204,18 @@ object Eventbrite {
       if EBEvent.providerWhitelist.contains(provider)
     } yield provider
 
-    val mainImageUrl: Option[String] = description.flatMap(desc => "<!--\\s*main-image: (.*?)\\s*-->".r.findFirstMatchIn(desc.html).map(_.group(1)) )
+    val mainImageUrl: Option[Uri] = for {
+      desc <- description
+      m <- """\smain-image:\s*(.*?)\s""".r.findFirstMatchIn(desc.html)
+      uri <- Try(Uri.parse(m.group(1))) match {
+        case Success(uri) =>
+          Logger.debug(s"Event $id main-image source url: $uri")
```
jennysivapalan: I don' think we need to log the success(or we need to change the log level in Production!).

frontend/test/services/GridServiceTest.scala
```diff
@@ -4,6 +4,7 @@ import model.Grid.GridResult
 import org.specs2.mutable.Specification
 import utils.Resource
 import model.GridDeserializer._
+import com.netaporter.uri.dsl._
```
jennysivapalan: Probably can be removed?

jennysivapalan:
:+1: :calendar: :man: :microphone:

frontend/test/services/GridServiceTest.scala
rtyley: Yeah, surprisingly, it _is_ required, the type of `mainImageUrl` has changed to `Option[Uri]` from `Option[String]`, and the implicit conversion is required for things to compile

frontend/app/model/Eventbrite.scala
rtyley: Ok, done

PR #240, opened 2015-02-24
Move comment builder to separate page

feedmypixel:
- removed the comment builder from the event overview pages
- placed on /staff/event-overview behind staff OAuth
- rejig of form filed to put less used fields at the bottom
- addition of description text

event overview the guardian membership 5


davidrapson:
:+1:

PR #241, opened 2015-02-24
Use for comprehension for event description

davidrapson:
Whilst looking into embedding videos in event descriptions I noticed that we fallback to a `foobarhtml` string if there is no event description. Let me know if I'm misreading this or if it's needed for something but falling back to nothing feels better than accidentally showing a dummy string.


feedmypixel:
:+1:

mattandrews:
:+1:

PR #248, opened 2015-02-24
open link in new window

feedmypixel:
open link to the grid in a new window on the /staff/event-details page


davidrapson:
:shipit:

PR #249, opened 2015-02-24
Empty listing message for Partner events

feedmypixel:
- added empty listing message for partner events
- made internal template for empty listing message

screen shot 2015-02-24 at 17 35 35

### DIFF
--- a/frontend/app/views/event/list.scala.html
+++ b/frontend/app/views/event/list.scala.html
@@ -8,10 +8,6 @@
 @import com.gu.membership.salesforce.Tier
 @import model.Benefits
 
-@emptyListing = {
-    <div class="listing__empty">@noResultsMessage</div>
-}
-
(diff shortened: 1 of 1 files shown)


davidrapson:
:+1: :tada:

frontend/app/views/event/list.scala.html
```diff
@@ -8,6 +8,10 @@
 @import com.gu.membership.salesforce.Tier
 @import model.Benefits
 
+@emptyListing = {
+    <div class="listing__empty">@noResultsMessage</div>
```
davidrapson: Although this is repeated, I actually prefer the repetition; my personal preference is to avoid anything that could accidentally cause a BEM fragment (child used without the parent). What do you think?

davidrapson:
:+1: LGTM

PR #250, opened 2015-02-24
View tidy up

feedmypixel:
- remove import thats not needed, 
- add ul opening tag

### DIFF
--- a/frontend/app/views/event/list.scala.html
+++ b/frontend/app/views/event/list.scala.html
@@ -15,11 +15,9 @@
                 <h3 class="listing__title h-intro">Featured</h3>
             </div>
             <div class="listing__content">
-                <ul>
                 @for(heroEvent <- eventPortfolio.heroOpt) {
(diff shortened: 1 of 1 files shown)


frontend/app/views/event/list.scala.html
```diff
@@ -16,6 +15,7 @@
                 <h3 class="listing__title h-intro">Featured</h3>
             </div>
             <div class="listing__content">
+                <ul>
```
davidrapson: You can remove both `ul` items, this doesn't need a list around it for the hero item (as it's just one full width block)

frontend/app/views/event/list.scala.html
```diff
@@ -6,7 +6,6 @@
 )(implicit token: play.filters.csrf.CSRF.Token)
 
 @import com.gu.membership.salesforce.Tier
-@import model.Benefits
```
davidrapson: :dancers:

davidrapson:
:+1:

PR #251, opened 2015-02-24
add npm clean cache to grunt-tc

feedmypixel:
We are seeing a lot of grunt issues on build in Team city - this appears to fix them. Thoughts on putting this in?
Build wise it doesn't appear to add to much time on.


davidrapson:
:+1: Given we essentially do a cache clean anyway at the moment by having to run the build again with `clear all files` checked this feels OK. Lets get this in but keep an eye on it.

PR #252, opened 2015-02-25
Clear out old benefits styles

davidrapson:
Do what the TODO says and clear out the old benefits styles


feedmypixel:
//DONE :+1:

PR #254, opened 2015-02-25
Simplify loader down to one-element

davidrapson:
This PR simplifies how the loader is constructed.

- Cleans up CSS conventions (use `is-` class rather than styling a `js-` class)
- Removes need for inline image / nested markup.
- Updates `processSubmit` to not be in utils (anything that's initialised in `main.js` should probably be in `modules`) and use loader helper to reduce code duplication

**Form loader**
loader mov

**Processing**
processing mov


feedmypixel:
for the cancel membership the processing on the left feels a little far away - more a question for BenW bar that good tidy up = :+1:

davidrapson:
@feedmypixel Yeah it does a little, might ask him and update in a separate PR if needed.

davidrapson:
@jamesoram Mind giving this a once over, needs a check of forms to make sure loading spinners still display as expected.

jamesoram:
:+1:

PR #256, opened 2015-03-02
JSON not being defined error being thrown by jshint - 

feedmypixel:
this stops the error being thrown on grunt build.


davidrapson:
:+1:

feedmypixel:
Both myself and @davidrapson feel there is something weird going on here. This has suddenly stopped building but we haven't made any changes around this area. We will keep an eye on it

PR #257, opened 2015-03-02
Add primary promo images to The Grid

davidrapson:
This PR adds promo primary images to The Grid. Also need to add secondary promo images but waiting on larger source images from Ben so this will happen in a separate PR.

Gets content images out of the codebase and allows gives us performance benefits of using responsive images here.

screen shot 2015-03-02 at 11 26 17

### DIFF
--- a/frontend/app/views/fragments/promos/promoPrimary.scala.html
+++ b/frontend/app/views/fragments/promos/promoPrimary.scala.html
@@ -1,6 +1,6 @@
 @(
     title: Html,
-    imageUrl: String,
+    imageBasePath: String,
     stampImageOpt: Option[String] = None,
     isLead: Boolean = false,
     toneClass: Option[String] = None
@@ -14,7 +14,7 @@
     </div>
     <div class="promo-primary__media@if(stampImageOpt){ stamped-image}">
         <div class="promo-primary__image">
-            <img src="@Asset.at(imageUrl)" alt="@title" class="responsive-img">
+            <div class="js-imager-loaded-image" data-src="@imageBasePath/{width}.jpg" data-available-widths="1000,500,140" data-alt="@title" data-class="responsive-img"></div>
         </div>
         @for(url <- stampImageOpt) {
             <div class="stamped-image__container">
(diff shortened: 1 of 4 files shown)


frontend/app/views/fragments/promos/promoPrimary.scala.html
```diff
@@ -14,7 +14,7 @@ <h2 class="promo-primary__title h-promo@if(isLead){ h-promo--primary}">@title</h
     </div>
     <div class="promo-primary__media@if(stampImageOpt){ stamped-image}">
         <div class="promo-primary__image">
-            <img src="@Asset.at(imageUrl)" alt="@title" class="responsive-img">
+            <div class="js-imager-loaded-image" data-src="@imageBasePath/{width}.jpg" data-available-widths="1000,500,140" data-alt="@title" data-class="responsive-img"></div>
         </div>
```
feedmypixel: potential to either 
- Change the current fragments.event.image() fragment to accept three arguments? 
- Or if more similar fragments are required when more promo images are added to the Grid then abstract this into another fragment?

frontend/app/views/info/about.scala.html
```diff
@@ -8,7 +8,7 @@
 
     @* ===== About Membership ===== *@
     <div class="page-slice l-constrained l-side-margins">
-        @fragments.promos.promoPrimary(title=Html("About<br>Membership"), imageUrl="images/about/about-intro-1140.jpg", isLead=true) {
+        @fragments.promos.promoPrimary(title=Html("About<br>Membership"), imageBasePath="https://media.guim.co.uk/0f30a12f5d00d2b0e445e81c171cd0168d3cf5a7/0_0_1140_684", isLead=true) {
```
feedmypixel: feels like media.guim.co.uk belongs in config seeing as its repeated

feedmypixel:
couple of comments - :+1:

frontend/app/views/fragments/promos/promoPrimary.scala.html
davidrapson: @feedmypixel I think I'm happy to leave this as it at the moment. My plan is to try a picture/srcset polyfill in the near future so will likely revisit that then.

frontend/app/views/info/about.scala.html
davidrapson: @feedmypixel As discussed, going to leave this as is.

PR #259, opened 2015-03-02
fix for the partner programming event showing up in whats on

feedmypixel:


feedmypixel:
:+1: from @jennysivapalan

PR #260, opened 2015-03-03
Native responsive images for event images

davidrapson:
Following a conversation in the last CS meeting with @sndrs and @phamann regarding responsive images this PR adds `srcset` / `sizes` support for event images

For anyone who want to familiarise themselves with the syntax for `srcset`, the following article does a good job of explaining it [ericportis.com/posts/2014/srcset-sizes/]()

- Update event images to support `srcset`/`sizes`
- Remove the regex / string interpolation from `RichEvent`
- I am using [https://github.com/aFarkas/respimage]() for the pollyfil based on recommendations from @phamann  
- Maintains lazyloading using [https://github.com/aFarkas/lazysizes]()

CC// @jennysivapalan I managed to remove the regex for event images which is the thing that really needed tests around it, let me know if you think there is anything here that needs tests revising around it.

One additional change I want to make is to abstract out the media queries / breakpoints that are passed to sizes but keen to get some eyes on this before I go too far with it.


frontend/app/model/RichEvent.scala
```diff
@@ -97,14 +96,21 @@ object RichEvent {
...
 
+    val imageMetadata = image.map(_.metadata)
+
+    val imgUrl = image.flatMap(_.assets.headOption).fold(fallbackImage){ asset =>
+      asset.secureUrl.getOrElse(asset.file)
+    }
+
+    val availableWidths = widths.mkString(",")
+
+    val srcsetOpt = image.map(_.assets.map { asset =>
```
jennysivapalan: You could add a test for this (to ensure it doesn't get changed later without reason).

davidrapson: @jennysivapalan Good call, added a test.

jennysivapalan:
:+1: :flower_playing_cards: 
Not sure how much time you have but could we try making a case class / trait that is specifically for image methods? Might be a nice tidy up.

davidrapson:
@jamesoram Mind giving this a once over. What you are looking to test is that event images (listings and detail pages) display OK.

davidrapson:
Note: Discussed with @jennysivapalan and going pair on extracting out some of the common image logic into a case class / trait as a separate PR

jamesoram:
:+1:

jamesoram:
:+1:

PR #261, opened 2015-03-03
For dev-purposes, tolerate having no network connection at all

rtyley:
Practically, this means lazy loading the GoogleGroupChecker so that the entire server doesn't die on class iniitalisation - this helps a little in on-the-train dev, obviously most of the site will /not/ work without a network connection!


davidrapson:
:+1: Offline first…

PR #264, opened 2015-03-03
Remove unused CSRF token which is definitely not wanted on cached pages

rtyley:
CSRF tokens wouldn't be much use at all if we shared them to lots of different users over the CDN!

CSRF tokens are needed on authenticated pages (which we don't cache), to ensure that form submission passes the CSRF token-check - a check designed to stop people being hacked by carefully-crafted links constructed by attackers - if the attackers don't know the token, they can't send the user directly to the POST action - not without the user seeing the form first.

I wonder if some of these errors in Sentry come from us using CSRF where we shouldn't:

https://app.getsentry.com/the-guardian/membership/group/41899380/

@mattandrews @davidrapson

### DIFF
--- a/frontend/app/views/event/guardianLive.scala.html
+++ b/frontend/app/views/event/guardianLive.scala.html
@@ -1,4 +1,4 @@
-@(eventPortfolio: model.EventPortfolio, pageInfo: model.PageInfo)(implicit token: play.filters.csrf.CSRF.Token)
+@(eventPortfolio: model.EventPortfolio, pageInfo: model.PageInfo)
 
 @main("Events", pageInfo=pageInfo) {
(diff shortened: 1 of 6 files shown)


davidrapson:
:+1: Good call.

mattandrews:
Agree with the PR but that Sentry error is about OAuth stuff (so staff sign-in?)...?!
:+1:

jennysivapalan:
Nice spot. :+1: :tanabata_tree:

rtyley:
> Sentry error is about OAuth stuff (so staff sign-in?)...?!

Sorry for the type - damn, have lost the right sentry link, and can't find it now.

PR #269, opened 2015-03-04
Separate out tools styles from mem app

feedmypixel:
Remove <strike>as many</strike> all dependencies from membership-frontend css for the tools css work. 
This is just good housekeeping and in preparation for any split of the /staff/event-overview work.

<strike>The two dependencies left are </strike>
<strike>- [normalize](https://github.com/guardian/membership-frontend/compare/separate-out-tools-styles-from-mem-app?expand=1#diff-1c5842acf841b30c4baf4f3880fd57eeR12) </strike>
<strike>- [webfonts](https://github.com/guardian/membership-frontend/compare/separate-out-tools-styles-from-mem-app?expand=1#diff-1c5842acf841b30c4baf4f3880fd57eeR36). </strike>

<strike>I have left webfonts in as the page looks nice with them.</strike>
I also took this opportunity to split the tools work out into separate files and also create its own vars file.

cc @davidrapson


davidrapson:
I'd actually say to just duplicate normalize in your new base directory too. It's a third-party dependancy so  there's not a big maintenance issue with duplicating the file.

davidrapson:
Might as well go all the way and dupe normalize and webfonts into tools too. Then at least from a CSS point of view we are decoupled from the public styles.

feedmypixel:
> Might as well go all the way and dupe normalize and webfonts into tools too. Then at least from a CSS point of view we are decoupled from the public styles.

done

davidrapson:
:+1: Much better

feedmypixel:
Total separation achieved :metal: (of the styles)

PR #270, opened 2015-03-05
Form validity tests

feedmypixel:
- form validity tests (more to come)
- improved code by return a boolean from profileValidation method


davidrapson:
:+1:

PR #271, opened 2015-03-05
Add method of modelling responsive images in scala

davidrapson:
This PR adds a couple of new case classes to help model images. The main goal is to use these to reduce the duplication of image logic in `RichEvent` but as an initial test I have updated the homepage slideshow to use the new case class.

- Add `ResponsiveImageGroup` case class
- Add tests for `srcset` string generation
- Update homepage slideshow to use new `ResponsiveImageGroup` as a test case
- Bonus (basic) lazy-loading of images in the slideshow

Thanks to @jennysivapalan for the assistance on this one.

### DIFF
--- a/frontend/app/controllers/FrontPage.scala
+++ b/frontend/app/controllers/FrontPage.scala
@@ -1,11 +1,112 @@
 package controllers
 
+import model.{ResponsiveImage, ResponsiveImageGroup}
 import play.api.mvc.Controller
 
 trait FrontPage extends Controller {
 
   def index = CachedAction { implicit request =>
-    Ok(views.html.index())
+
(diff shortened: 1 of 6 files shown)


jennysivapalan:
Looks good - I think the case classes will be more useful when we start pulling images for events / masterclasses etc.

davidrapson:
@jennysivapalan Yeah definitely, probably going to have a look at that tomorrow as it looked fairly straightforward.

davidrapson:
@jamesoram Mind giving this a once over? To confirm: Make sure that the carousel images on the homepage load OK.

jamesoram:
:+1:

PR #276, opened 2015-03-05
Fix alignment flash message

feedmypixel:
@marchibbins spotted the alignment of the feedback flash message was out of alignment when you were between the tablet and desktop breakpoints.
The flash message is displayed on pages with different layouts, functional and form layout pages.
- moved the layout out of the flash message fragment
- placed layout around the flash message fragment

## old (miss aligned)
old

## aligned correctly 
fixed

### DIFF
--- a/frontend/app/views/fragments/notifications/flashMessage.scala.html
+++ b/frontend/app/views/fragments/notifications/flashMessage.scala.html
@@ -1,10 +1,5 @@
 @(flashMessage: model.FlashMessage)
 
-<section class="page-section page-section--no-padding">
-    <div class="page-section__content">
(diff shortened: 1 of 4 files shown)


jennysivapalan:
I think there might be a few more cases where the flash message is used - I added one for paid to paid upgrade.

davidrapson:
:confused: Feels a little awkward to wrap these in a whole page section just for the flash message (not saying there should be any layout styles on the message component). Can these message live inside another container?

feedmypixel:
> I think there might be a few more cases where the flash message is used - I added one for paid to paid upgrade.

merged and added

davidrapson:
:+1:

PR #279, opened 2015-03-05
Video what is membership

feedmypixel:
Added video to about page 

about the guardian membership


davidrapson:
:+1: LGTM for now. I think we should take another look at this afterwards to see if there's a clearer way of expressing what this component is. I think it's sensible to piggyback on the secondary promo now. Lets get this in.

PR #285, opened 2015-03-06
add promo video to patterns

feedmypixel:
### An example of promo video fragment on the patterns page
***
screen shot 2015-03-06 at 11 22 14


davidrapson:
:+1: Nice

PR #198, opened 2015-02-16
Add config for new tier & restrict access for now

davidrapson:
This is work from @wpf500 to enable a new tier. Just opening it up as a PR so we have some more visibility on it.

PR #287, opened 2015-03-06
Don't crash if the social-conversion tracker ids aren't available for Tier

rtyley:
Got a crash in PROD (and only in PROD!) because we don't have these codes for Supporter tier yet...

https://github.com/guardian/membership-frontend/pull/198#issuecomment-77542832
https://app.getsentry.com/the-guardian/membership/group/56964217/

cc @davidrapson


feedmypixel:
LGTM :+1:

PR #288, opened 2015-03-06
Add order data to tracking tool

jennysivapalan:
To improve how we track event orders we need to push the order data. Note we hash the order id to maintain privacy stuff.


mattandrews:
LGTM :+1:

PR #289, opened 2015-03-06
Merge Joiner controller with Joining

davidrapson:
Whilst looking at the member's only content user journey for the choose tier page I noticed the `Joining` controller should probably be merged with `Joiner` as they cover largely the same concept.

@rtyley I'd appreciate your opinion on this one.


rtyley:
:+1:

PR #291, opened 2015-03-06
Make Sale end date show *time* if it's happening today-ish

rtyley:
Actually a 24 hour window either side at the moment

Madeleine Williams requested:

- Would it be possible to state on the final day of ticket sales the end time for sales? The end time is the start time of the event as people can buy online right up until then.

cc @davidrapson

### DIFF
--- a/frontend/app/views/fragments/event/ticketSales.scala.html
+++ b/frontend/app/views/fragments/event/ticketSales.scala.html
@@ -2,6 +2,7 @@
 
 @import views.support.Dates._
 @import org.joda.time.Instant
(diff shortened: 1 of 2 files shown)


davidrapson:
:+1:

rtyley:
Confirmed, looks good on Production:

screen shot 2015-03-07 at 21 51 42

https://membership.theguardian.com/event/guardian-live-kazuo-ishiguro-15220129818

Note that the sale end is 2pm, not the show start time of 3pm - which is correct, that matches what's set in the EventBrite console for that event:

screen shot 2015-03-07 at 22 02 54

PR #294, opened 2015-03-06
Clean up how event images are modeled in RichEvent

davidrapson:
Update RichEvent to model images as ResponsiveImageGroups

- Switch event images to be an `Option` so we can handle the case where there is no image
- Remove need for `fallbackImage`
- Remove need for `availableWidths` and `srcset` to be calculated in `RichEvent`

The generation of the `imgOpt` in each case is the only slightly complex bit of Scala so I'd appreciate feedback on that in case there's a simpler way of building these up (maybe move the generation to `Asset` or `Grid` at least for live events)

// @jennysivapalan

### DIFF
--- a/frontend/app/model/RichEvent.scala
+++ b/frontend/app/model/RichEvent.scala
@@ -81,46 +81,34 @@ object RichEvent {
 
   trait RichEvent {
     val event: EBEvent
-    val imgUrl: String
+    val imgOpt: Option[model.ResponsiveImageGroup]
     val socialImgUrl: String
     val tags: Seq[String]
     val metadata: Metadata
     val imageMetadata: Option[Grid.Metadata]
     val contentOpt: Option[Content]
-    val srcsetOpt: Option[Seq[String]]
-    val availableWidths: String
-    val fallbackImage = views.support.Asset.at("images/event-placeholder.gif")
     val pastImageOpt: Option[Asset]
-
     def deficientGuardianMembersTickets: Boolean
   }
 
   abstract class LiveEvent(image: Option[EventImage], contentOpt: Option[Content]) extends RichEvent {
 
-    private val widths = image.fold(List.empty[Int])(_.assets.map(_.dimensions.width))
-
     val imageMetadata = image.map(_.metadata)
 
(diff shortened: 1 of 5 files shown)


frontend/app/model/RichEvent.scala
```diff
@@ -148,17 +143,23 @@ object RichEvent {
...
+        )
+      }
+    }
 
-    val srcsetOpt = None
+    val imageMetadata = None
 
-    val socialImgUrl = imgUrl
+    //TODO: Should be Option[String]
+    val socialImgUrl = imgOpt.map(_.defaultImage).getOrElse("")
```
davidrapson: This is a bit naff, returning an empty string. But switching this to an `Option` is more refactoring than feels appropriate in this PR and returning an empty string for the `socialImageUrl` is OK in the short term.

davidrapson:
@jennysivapalan Mind taking a look at this one?

frontend/app/model/RichEvent.scala
```diff
@@ -81,35 +81,42 @@ object RichEvent {
...
+      Some(ResponsiveImageGroup(
+        altText=event.name.toString(),
+        availableImages=imgData.assets.map { asset =>
+          ResponsiveImage(
+            path=asset.secureUrl.getOrElse(asset.file),
+            width=asset.dimensions.width
+          )
+        }
+      ))
+    }.getOrElse(None)
```
jennysivapalan: You don't need the `getOrElse(None)` as the `map` will do this if `image` is `None`.

frontend/app/model/RichEvent.scala
jennysivapalan: Okay as long as we can change this fairly soon after the PR goes out (or we'll forget).

jennysivapalan:
Looks good other than comment on the `None` :+1: :o:

frontend/app/model/RichEvent.scala
davidrapson: Definitely

frontend/app/model/RichEvent.scala
davidrapson: Had to use a `flatMap` in order to get rid of the `getOrElse(None)`  but that seems to have done the trick.

PR #295, opened 2015-03-09
Disable source maps in production

davidrapson:
**TL;DR Production CSS is now 3x smaller**

We have a couple of different grunt tasks which generate source maps, so there's a few different places so we end up with this situation:

screen shot 2015-03-09 at 10 05 42

This PR ensures that source maps are fully disabled in production.

**Before**
css-size-before

**After**
css-size-after

I did start some work to monitor asset sizes in Cloud Watch but never actually finished it, this might prompt me to pick that back up again.

### DIFF
--- a/frontend/Gruntfile.js
+++ b/frontend/Gruntfile.js
@@ -55,7 +55,7 @@ module.exports = function (grunt) {
                 }],
                 options: {
                     style: 'compressed',
-                    sourcemap: isDev ? 'auto' : 'none',
+                    sourcemap: isDev ? true : false,
                     noCache: true,
                     quiet: isDev ? false : true,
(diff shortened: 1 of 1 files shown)


feedmypixel:
:+1:

mattandrews:
We probably still want them for JS (if we don't already) though? Depends on how much it pushes the filesize I guess.

davidrapson:
@mattandrews Yeah it's OK for JS at the moment, no need to disable them there.

mattandrews:
In that case :+1:

davidrapson:
Ah that's better…

screen shot 2015-03-10 at 10 40 18

PR #299, opened 2015-03-09
Tagging for video rmd media

feedmypixel:
Tagging for PHD Media on
- /join
- /about
- /join/*/thankyou

I have adapted their scripts to use images rather than iframes.

For the thank you page we are sending
- totalCost

- free

#### Join
join

#### About
about

#### Thank you - Paid
thankyou-paid

#### Thank you - Free
thankyou - free


davidrapson:
:+1: Let's get this out, but worth checking a couple of things in the morning:

1. Are we OK using an image rather than the iframe in the provided code
2. Do we need to have this so configurable for code that we're just going to remove again?

PR #302, opened 2015-03-10
Switch to use iframe for video tracking

feedmypixel:
- switch to use Iframe
- remove details from config so they are all in one place

This is code straight from phd Medias Floodlight tags. The code is not great but it is now in one place.
The only thing changed is the removal of the protocal.

cc @davidrapson


davidrapson:
:+1: I think it makes sense to use the tracking code as provided. LGTM as long as you're happy that the document.write calls are safe here.

PR #303, opened 2015-03-10
Event page class-name and specifictiy fixes

davidrapson:
This PR fixes a rogue `u-cf` that was being appended to the `event-details` container. And cleans up a few layout and specificity details that fixing this classname highlighted.


mattandrews:
Good catch! :+1:

PR #305, opened 2015-03-10
Re-add Patrons to nav

davidrapson:
Looks like we lost Patrons from the nav recently. This PR adds it back.

screen shot 2015-03-10 at 15 15 32

screen shot 2015-03-10 at 15 15 38


feedmypixel:
:+1:

PR #308, opened 2015-03-11
Check if user has a valid subscription

jennysivapalan:
Hooks into the CAS service to check if a subscriber is valid. Note this isn't used on the frontend, but we require a Salesforce update before we continue on this part of the work. However this seemed like a small chunk that could go in anyway.

I've created a new endpoint for this under `/user/subscriber/details` which requires authentication via Identity.


rtyley:
:+1:

PR #309, opened 2015-03-11
Refactor social images

davidrapson:
- Update socialImgUrl to be an `Option`, now that event images are considered optional
- Remove need to use `socialImgUrl` on event items (not required for Schema markup)
- Tweak event overview page to not use social image URL

// @jennysivapalan


mattandrews:
:+1:

PR #310, opened 2015-03-11
click on element 10

jamesoram:


davidrapson:
@jamesoram Out of interest, why do you keep needing to change this number? Is this something we can help with?

jamesoram:
this is due to events being sold out

davidrapson:
@jamesoram I think we can make this less brittle by giving you a new `qa-` selector for `eventLinkList` which only gets applied for available events. We can always give you another selector for sold-out events if you need them too.

PR #312, opened 2015-03-12
Improve ResponsiveImage default sizes

davidrapson:
This PR makes a couple of adjustments to `ResponsiveImage`

- Tries to use a medium sized image for `defaultImage` otherwise uses the smallest (rather than defaulting to largest which can be ~2000px)
- Removes `src` for lazy-loaded event images

// @feedmypixel


feedmypixel:
nice addition for the image choice and fallback :+1:

PR #315, opened 2015-03-13
Remove l-side-margins classes

davidrapson:
We actually don't need the `l-side-margin` classes on individual components yet as we have this class globally on the wrapper container. For now we can remove them.

The double side margins sometimes cause the following issue so this resolves a layout issue too:

screen shot 2015-03-13 at 11 33 45


feedmypixel:
:+1:

PR #316, opened 2015-03-13
Remove default margin from actions

davidrapson:
Nearly all of our `action`s have a no margin utility class on them. Given we more often remove the margin than not we should just remove the default margin. As an example (included in this PR) every action on the About page has a no margin class on it.

To safe my self the trouble, I'm just going to quote Harry Roberts (again…):
> Peg things onto simpler things that came before it, do not start too complex and risk having to undo your work later on; you’ll end up **writing more CSS to achieve less styling**.
[http://csswizardry.com/2012/11/code-smells-in-css/]()

Thankfully it required relatively few changes to fix this:

- Remove default margin from actions
- Updates styles to address the few cases where the margin was assumed
- Removes all margin utility classes from actions on about and patron page (as an example of the improvement)


frontend/app/views/info/about.scala.html
```diff
@@ -15,8 +15,8 @@
                 <p>Join before March 31st to become a Founding Member and the price you pay will not rise.</p>
             </div>
             <div class="action-group">
-                <a class="action u-no-top-margin" href="@routes.Joiner.tierList">Join today</a>
-                <a class="action action--secondary u-no-top-margin js-scroll-to" href="#whats-included">
+                <a class="action" href="@routes.Joiner.tierList">Join today</a>
+                <a class="action action--secondary js-scroll-to" href="#whats-included">
```
feedmypixel: for another PR ```.js-scroll-to``` is not being used in the js - did this slip in when the scroll to functionality was canned?

davidrapson: @feedmypixel Possibly, I did spot it and nearly removed it, but didn't want to muddy the PR

feedmypixel:
:+1:

PR #317, opened 2015-03-13
Add new fields for Zuora (ready to make a setting update)

jennysivapalan:
In order for subscribers to get a specific rate plan we need to enable a setting change in the Zuora console. This setting change will requires us to pass new fields for all subscriptions. Existing subscriptions will be unaffected but new subscriptions and amendments to all subscriptions will be affected by this setting change. We can start passing the required fields before we make the setting change. They will be ignored until the setting change comes into affect. 

For Subscribe calls to Zuora the new field is called `ContractAcceptanceDate`. For Amend (upgrades, downgrades and cancellations) the new field is called `CustomerAcceptanceDate`. Sadly this has taken me the best part of a week to figure out, I need :cake:.

@jamesoram I've tested this with dev environment with both the setting change off and on and things seem to work fine. I've spoken to Jason Burr about the compabitility testing for this and we're meeting on Monday. In the meantime if you are able to test this, we just need to check all flows work as expected with the setting change off.

@rtyley does the above explanation make sense in terms of how we want to manage the change?

### DIFF
--- a/frontend/app/services/zuora/ZuoraActions.scala
+++ b/frontend/app/services/zuora/ZuoraActions.scala
@@ -139,6 +139,7 @@ case class Subscribe(memberId: MemberId, customerOpt: Option[Stripe.Customer], r
           <ns1:Subscription xsi:type="ns2:Subscription">
             <ns2:AutoRenew>true</ns2:AutoRenew>
             <ns2:ContractEffectiveDate>{now}</ns2:ContractEffectiveDate>
+            <ns2:ContractAcceptanceDate>{now}</ns2:ContractAcceptanceDate>
             <ns2:InitialTerm>12</ns2:InitialTerm>
             <ns2:RenewalTerm>12</ns2:RenewalTerm>
             <ns2:TermStartDate>{now}</ns2:TermStartDate>
@@ -166,6 +167,7 @@ case class CancelPlan(subscriptionId: String, subscriptionRatePlanId: String, da
         <ns1:Amendments>
           <ns2:EffectiveDate>{dateStr}</ns2:EffectiveDate>
           <ns2:ContractEffectiveDate>{dateStr}</ns2:ContractEffectiveDate>
+          <ns2:ContractAcceptanceDate>{dateStr}</ns2:ContractAcceptanceDate>
           <ns2:Name>Cancellation</ns2:Name>
           <ns2:RatePlanData>
             <ns1:RatePlan>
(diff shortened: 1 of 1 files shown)


frontend/app/services/zuora/ZuoraActions.scala
```diff
@@ -139,6 +139,7 @@ case class Subscribe(memberId: MemberId, customerOpt: Option[Stripe.Customer], r
           <ns1:Subscription xsi:type="ns2:Subscription">
             <ns2:AutoRenew>true</ns2:AutoRenew>
             <ns2:ContractEffectiveDate>{now}</ns2:ContractEffectiveDate>
+            <ns2:ContractAcceptanceDate>{now}</ns2:ContractAcceptanceDate>
```
rtyley: For (my own, mostly) reference, I'm just going to annotate these lines with the Zuora documentation:

https://knowledgecenter.zuora.com/BC_Developers/SOAP_API/E1_SOAP_API_Object_Reference/Subscription

image

frontend/app/services/zuora/ZuoraActions.scala
```diff
@@ -166,6 +167,7 @@ case class CancelPlan(subscriptionId: String, subscriptionRatePlanId: String, da
         <ns1:Amendments>
           <ns2:EffectiveDate>{dateStr}</ns2:EffectiveDate>
           <ns2:ContractEffectiveDate>{dateStr}</ns2:ContractEffectiveDate>
+          <ns2:CustomerAcceptanceDate>{dateStr}</ns2:CustomerAcceptanceDate>
```
rtyley: https://knowledgecenter.zuora.com/BC_Developers/SOAP_API/E1_SOAP_API_Object_Reference/Amendment#CustomerAcceptanceDate

image

rtyley:
:+1: 

Just for the reference of other devs, this seems to be the pertinent bit of the Zuora doc:

https://knowledgecenter.zuora.com/CB_Billing/G_Subscriptions/B_Creating_Subscriptions/E_Activating_a_Subscription#Dates_Required_to_Activate_a_Subscription

```
Dates Required to Activate a Subscription
Your business may not require the use of all three dates, but instead
 require the use of two out of three dates, or just one of the dates.
 If only one date is used, that date must be the contract effective date.

The required trigger dates can vary by product and can be configured
 using the Default Subscription Settings. If your company's default
 subscription settings have Require Customer Acceptance of Orders?
 or Require Service Activation? set to Yes, you must enter those dates
 to fully activate the subscription. If either of these options are marked
 No, the applicable date (customer acceptance or service activation)
will be made to match the contract effective date.

Often the contract effective date (or the date that triggers the subscription
 charges, such as the service activation date or customer acceptance
 date) lines up with the bill cycle day specified in the customer account.
 When the two dates align, your customer will be billed on the anniversary
 date of subscription activation. This is known as "anniversary billing."
```

jamesoram:
:+1:

jennysivapalan:
Just waiting on Jason Burr to look over this in Zuora.

PR #318, opened 2015-03-13
Respect Guardian Style for displaying times

rtyley:
http://www.theguardian.com/guardian-observer-style-guide-t

    times: 1am, 6.30pm, etc

So we display times like this:

image

...and this...

image

Note the `.`, not `:` between the hour and the minute - that's _Guardian_ style :-)


davidrapson:
:+1:

rtyley:
Verified in Production on these urls:

https://membership.theguardian.com/event/dior-and-i-preview-screening-and-director-qa-15946684964
https://membership.theguardian.com/event/guardian-live-john-gray-and-will-self-is-free-will-an-illusion-15006844877

PR #320, opened 2015-03-13
change url, change controller name, tidy up staff OAuth

feedmypixel:
- a url change from ```offer/subscriber``` to ```offers/subscriber```
- relevant changes to the controller name
- tidy up of OAuth so it is staff not just perm staff
- added //TODO


rtyley:
:+1:

PR #325, opened 2015-03-16
Keep track of item when signing up for members area content

davidrapson:
- Store the referrer in a cookie if we are coming from a piece of content with `membershipAccess`
- Show the content on the thank you page to allow the user to return to the content they were trying to see.

~~Need to confirm with @davidmcdowell exactly what needs to be shown on the thank you page.~~ Removed the changes to the thank you page content, this will come in a separate PR.

### DIFF
--- a/frontend/app/controllers/Joiner.scala
+++ b/frontend/app/controllers/Joiner.scala
@@ -1,3 +1,5 @@
+
+
 package controllers
 
 import actions.Functions._
@@ -5,6 +7,7 @@ import actions._
 import com.gu.membership.salesforce.{PaidMember, ScalaforceError, Tier}
 import com.gu.membership.stripe.Stripe
 import com.gu.membership.stripe.Stripe.Serializer._
+import com.netaporter.uri.Uri
 import com.netaporter.uri.dsl._
 import configuration.{Config, CopyConfig}
 import controllers.Testing.AuthorisedTester
@@ -22,6 +25,7 @@ import scala.concurrent.Future
 import tracking.{EventData, EventActivity, MemberData, ActivityTracking}
 
 trait Joiner extends Controller with ActivityTracking {
+  val JoinReferrer = "join-referrer"
 
   val contentApiService = GuardianContentService
 
@@ -48,15 +52,10 @@ trait Joiner extends Controller with ActivityTracking {
       Some(CopyConfig.copyDescriptionChooseTier)
     )
 
-    val contentReferer = request.headers.get("referer");
+    val contentReferer = request.headers.get("referer")
     val contentAccess = request.getQueryString("membershipAccess")
 
-    if(contentReferer.nonEmpty && contentAccess.nonEmpty) {
-      Ok(views.html.joiner.tierChooser(eventOpt, pageInfo)).withCookies(Cookie("GU_MEM_REFERER", contentReferer.get, secure = true, httpOnly = false))
-    } else {
-      Ok(views.html.joiner.tierChooser(eventOpt, pageInfo))
-    }
-
+    Ok(views.html.joiner.tierChooser(eventOpt, pageInfo)).withSession(request.session.copy(data = request.session.data ++ contentReferer.map(JoinReferrer -> _)))
   }
 
   def staff = PermanentStaffNonMemberAction.async { implicit request =>
@@ -183,10 +182,8 @@ trait Joiner extends Controller with ActivityTracking {
       Future.sequence(optFuture.toSeq).map(_.headOption)
     }
 
-    val refererCookie = request.cookies.get("GU_MEM_REFERER").map(_.value.replace(Config.guardianUrl, ""))
-
-    val futureContentOpt = refererCookie.map { referer =>
-      contentApiService.contentItemQuery(referer).map(_.content.map(MembersOnlyContent))
+    val futureContentOpt = request.session.get(JoinReferrer).map { referer =>
+      contentApiService.contentItemQuery(referer.path).map(_.content.map(MembersOnlyContent))
     }.getOrElse(Future.successful(None))
 
     for {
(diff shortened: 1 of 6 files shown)


frontend/app/controllers/Joiner.scala
```diff
@@ -170,11 +183,24 @@ trait Joiner extends Controller with ActivityTracking {
       Future.sequence(optFuture.toSeq).map(_.headOption)
     }
 
+    val futureContentOpt = request.cookies.get("GU_MEM_REFERER").map(_.value).map { referer =>
+      val refererPath = referer.replace("http://www.theguardian.com", "")
```
jennysivapalan: Use the `Config` class to define http://www.theguardian.com. That way we can configure it for dev environments etc.

frontend/app/controllers/Joiner.scala
```diff
@@ -170,11 +183,24 @@ trait Joiner extends Controller with ActivityTracking {
       Future.sequence(optFuture.toSeq).map(_.headOption)
     }
 
+    val futureContentOpt = request.cookies.get("GU_MEM_REFERER").map(_.value).map { referer =>
```
davidrapson: Not sure if I should rename this…

Refe**rr**er is the correct spelling of the word but it's actually misspelled in the spec, so the HTTP header name is Refe**r**er. Guess it makes sense to keep it consistent with the HTTP header we are checking.

frontend/app/controllers/Joiner.scala
```diff
@@ -170,11 +183,24 @@ trait Joiner extends Controller with ActivityTracking {
...
+      val refererPath = referer.replace("http://www.theguardian.com", "")
+      contentApiService.contentItemQuery(refererPath).map(_.content.map(MembersOnlyContent))
+    }.getOrElse(Future.successful(None))
+
```
jennysivapalan: I had a look at this again after destroying your attempt this morning..sorry about that.

What you have written works but I'd split it up a bit. Currently the code is mixing the working out what the item path is and calling the Content API with the path.

So I would do something like:

```
val refererCookie = request.cookies.get("GU_MEM_REFERER").map(_.value.replace("http://www.theguardian.com", ""))
    
val futureContentOpt = refererCookie.map { referer =>
  contentApiService.contentItemQuery(referer).map(_.content.map(MembersOnlyContent))
}.getOrElse(Future.successful(None))
```

davidrapson: Hah, I started out with something similar to this, until I trusted IntelliJ's `Refactor > Inline` option.

frontend/app/controllers/Joiner.scala
```diff
@@ -170,11 +183,25 @@ trait Joiner extends Controller with ActivityTracking {
       Future.sequence(optFuture.toSeq).map(_.headOption)
     }
 
+    val refererCookie = request.cookies.get("GU_MEM_REFERER").map(_.value.replace(Config.guardianUrl, ""))
```
jennysivapalan: Just realised that the `val refererCookie` could be called `path` as that's what is being formed at this point?

jennysivapalan:
:+1: :feet:

rtyley:
I'm still taking a look at this, could you hold off merging?

frontend/app/controllers/Joiner.scala
```diff
@@ -44,7 +47,16 @@ trait Joiner extends Controller with ActivityTracking {
       request.path,
       Some(CopyConfig.copyDescriptionChooseTier)
     )
-    Ok(views.html.joiner.tierChooser(eventOpt, pageInfo))
+
+    val contentReferer = request.headers.get("referer");
+    val contentAccess = request.getQueryString("membershipAccess")
+
+    if(contentReferer.nonEmpty && contentAccess.nonEmpty) {
+      Ok(views.html.joiner.tierChooser(eventOpt, pageInfo)).withCookies(Cookie("GU_MEM_REFERER", contentReferer.get, secure = true, httpOnly = false))
```
rtyley: I think this cookie _could_ be `httpOnly = true`, and that would be preferable, yes? As it happens, Play actually offers a 'Session' cookie for doing this sort of thing, which has the right defaults:

https://www.playframework.com/documentation/2.3.x/ScalaSessionFlash

I think we should probably be using that for preference.

frontend/app/controllers/Joiner.scala
rtyley: Rather than configure in the Guardian url, we can just use [Uri-parsing](https://github.com/NET-A-PORTER/scala-uri#parsing-uris) to get the 'path' part of the url:

```
request.cookies.get("GU_MEM_REFERER").map(_.value.path)
```

frontend/app/controllers/Joiner.scala
```diff
@@ -44,7 +47,16 @@ trait Joiner extends Controller with ActivityTracking {
       request.path,
       Some(CopyConfig.copyDescriptionChooseTier)
     )
-    Ok(views.html.joiner.tierChooser(eventOpt, pageInfo))
+
+    val contentReferer = request.headers.get("referer");
+    val contentAccess = request.getQueryString("membershipAccess")
```
rtyley: What happens to this `contentAccess` field? It doesn't seem to really get used?

davidrapson:
@rtyley Merged your changes and added the failure check, mind giving this another look?

frontend/app/controllers/Joiner.scala
```diff
@@ -44,7 +51,11 @@ trait Joiner extends Controller with ActivityTracking {
       request.path,
       Some(CopyConfig.copyDescriptionChooseTier)
     )
-    Ok(views.html.joiner.tierChooser(eventOpt, pageInfo))
+
+    val contentReferer = request.headers.get("referer")
```
rtyley: We probably should use the `Referer` constant here - I think it's already in scope.

rtyley:
LGTM, tho' prob do the `Referer` constant.

PR #327, opened 2015-03-16
Add new Birkbeck logo

mattandrews:
cc @jamesoram @feedmypixel 

Nothing major, just changes their logo (as requested).


mattandrews:
Tried to be clever and use `git rebase` here to make my commit history look better, but instead it resulted in two further commits. Doubtless @rtyley is gnashing his teeth and wailing in agony somewhere near Greenwich...

jamesoram:
:+1:

davidrapson:
I get the following in the diff:



Should the logo still have transparency?

mattandrews:
No, it's specifically meant to have this black background (that's their official colour). Wouldn't want the white to be transparent now either.

davidrapson:
Ah fair enough :+1:

PR #331, opened 2015-03-17
Move modal and analytics JS into modules

davidrapson:
This PR moves `modal` and `analytics` JS into modules.

The thinking is that anything that has an `#init` method and is called in `main.js` should be in modules; conversely utils should only be included in other modules.


feedmypixel:
:+1:

PR #336, opened 2015-03-17
Add asset size monitoring script

davidrapson:
This PR is an attempt to track the size of our CSS and JS so we can monitor if they get too large. The current approach was something @wpf500 help me put together several weeks ago, thought I'd actually try and finish it off.

I think the general approach was to run a shell script when new boxes start up as we only want to log asset sizes a) for PROD and b) when assets actually change. The one thing I'm not clear on is how to actually run the script as part of the deploy process.

@rtyley I'd appreciate your :eyes: on this, there's probably a number of different ways to log the asset size so if you think there is a better way I'm all :ear:s

### DIFF
--- a/frontend/asset-size.sh
+++ b/frontend/asset-size.sh
@@ -7,8 +7,8 @@ for stylesheet in ${STYLESHEETS[@]}; do
     size=`cat $stylesheet | wc -c`
     gzip_size=`gzip -c $stylesheet | wc -c`
 
-    aws cloudwatch put-metric-data --namespace TestGlob --metric-name $name --dimensions "Compression=None" --value $size
-    aws cloudwatch put-metric-data --namespace TestGlob --metric-name $name --dimensions "Compression=Gzip" --value $gzip_size
+    aws cloudwatch put-metric-data --namespace Assets --metric-name $name --dimensions "Compression=None" --value $size
+    aws cloudwatch put-metric-data --namespace Assets --metric-name $name --dimensions "Compression=Gzip" --value $gzip_size
 done
 
 JAVASCRIPTS=("public/dist/javascripts/*/*.js")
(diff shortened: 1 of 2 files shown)


davidrapson:
@rtyley as discussed removed script from artifact

mattandrews:
Script looks good, not sure myself how the deploy aspect works though.

wpf500:
You need to add it to the CloudFormation script!

mattandrews:
Who said that?

davidrapson:
Just the wind probably

davidrapson:
That @rtyley fella had a nice idea of making it part of the team city build, so still undecided what will trigger this script (hence why I removed it from the build artefact for now)

mattandrews:
I was thinking that, would we be able to restrict it to only run on builds of `master` though? Would be annoying to pollute the data with all the branches too.

rtyley:
> I was thinking that, would we be able to restrict it to only run on builds of master though?

We'd probably have to do some of the jiggery-pokery in https://confluence.jetbrains.com/display/TCD8/Working+with+Feature+Branches to get it `master`-only.

For a gold-plated solution, I'd like to see size-increase information displayed in the pull-request itself - because when bad things happen, it's nice to be able to see easily _what_ exactly caused them. This would probably require a GitHub bot so I'm basically making work for myself here :hankey:

davidrapson:
@rtyley Do you think it's worth getting this PR merged with just the script and then we can look at the best way of running it?

rtyley:
> Do you think it's worth getting this PR merged with just the script and then we can look at the best way of running it?

Yeah, go ahead and merge it in, it's a reasonable first cut.

mattandrews:
images

rtyley:
So true.

mattandrews:
spandau_ballet_true

PR #338, opened 2015-03-17
change supporter url

feedmypixel:
change request for supporter landing page url

from ```/about/supporters``` to ```/about/supporter```

- rename of controller method and view fragment

cc @davidmcdowell


mattandrews:
:+1:

PR #339, opened 2015-03-17
change to page headline for free to paid page

feedmypixel:


frontend/app/views/tier/upgrade/freeToPaid.scala.html
```diff
@@ -14,7 +14,7 @@
     <main role="main" class="page-content l-constrained">
 
         <section class="page-header">
-            <h1 class="page-headline">Payment</h1>
+            <h1 class="page-headline">Address & billing</h1>
```
rtyley: So `&amp;' is necessary?! An education for me.

rtyley:
:+1:

frontend/app/views/tier/upgrade/freeToPaid.scala.html
feedmypixel: > So &amp; is not necessary?! An education for me.

http://stackoverflow.com/questions/436615/when-should-one-use-html-entities

PR #340, opened 2015-03-17
Home Icon Bug Iphone4

feedmypixel:
A bug came in from BenW about the home icon on his Iphone 4.

### Before
screen shot 2015-03-17 at 17 37 01

### After
screen shot 2015-03-17 at 17 39 48


davidrapson:
:+1:

PR #341, opened 2015-03-17
Show different copy and CTA for supporter

feedmypixel:
In the **Get Started** section show different copy and change the cta destination if you are a Supporter.

### Supporter
screen shot 2015-03-17 at 17 34 23

### Other Tiers
screen shot 2015-03-17 at 17 42 42

cc @davidmcdowell

### DIFF
--- a/frontend/app/views/joiner/thankyou.scala.html
+++ b/frontend/app/views/joiner/thankyou.scala.html
@@ -4,6 +4,8 @@
 @import views.support.Dates._
 @import views.support.Prices._
 @import views.support.Social.joinThankyou
+@import com.gu.membership.salesforce.Tier
+@import configuration.Config
 
 @iframeHeight(event: model.RichEvent.RichEvent) = @{
     val ticketHeight = 60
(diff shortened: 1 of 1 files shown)


davidrapson:
The `case` matching in this file is starting to get pretty gnarly in this file, and will get even gnarlier soon as I will be adding a variant for members area content. :+1: for now but we'll need to come back and clean this up (move more things to fragments; split this out into multiple templates or move more config to the controller)

frontend/app/views/joiner/thankyou.scala.html
```diff
@@ -24,6 +26,17 @@
     }
 }
 
+@getStarted(content: Html) = {
```
davidrapson: Can the case matching just go in side `page-section__content`? as nothing really changes here? Personal preference but not a huge fan of the inline fragments in these cases.

davidrapson:
:+1: LGTM. A couple of comments but more about the general complexity of the file, your changes look good.

PR #344, opened 2015-03-18
change oauth list email for test-users

feedmypixel:
change email to a list that a user *is* on rather than one the *think* they are on! :)

cc @rtyley


rtyley:
Think it's best to say `crm@guardian.co.uk`, as that's the domain Google Apps still thinks we're on!

rtyley:
:+1:

PR #347, opened 2015-03-18
Youtube verification meta tag

feedmypixel:
We have had a request from the Guardian Membership YouTube team to allow us to google verify their account against our site. They are keen to show slides/extra video on their YouTube channel before live streams.

- add Guardian Membership YouTube google verification code to config
- add meta tag in main


mattandrews:
:+1:

rtyley:
@feedmypixel notes that "Removing the tag from the page can cause your site to become unverified, and you will need to go through the verification process again." - so this stuff should stay in.

https://support.google.com/webmasters/answer/35659?hl=en

PR #349, opened 2015-03-18
Do not send test user stats to tracking tool

jennysivapalan:
It's not the best implementation but this should stop test users in production being tracked in our tool. It doesn't catch two tracking activities which are view event details and buy action invoked but I don't think it matters.

@rtyley is there a better way to do this?

cc @mattandrews @dominickendrick

### DIFF
--- a/frontend/app/controllers/Joiner.scala
+++ b/frontend/app/controllers/Joiner.scala
@@ -147,7 +147,7 @@ trait Joiner extends Controller with ActivityTracking {
         } {
           event.service.wsMetrics.put(s"join-$tier-event", 1)
           val memberData = MemberData(member.salesforceContactId, request.user.id, tier.name)
-          track(EventActivity("membershipRegistrationViaEvent", Some(memberData), EventData(event)))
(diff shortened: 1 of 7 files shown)


frontend/app/controllers/Joiner.scala
```diff
@@ -147,9 +147,11 @@ trait Joiner extends Controller with ActivityTracking {
           eventId <- PreMembershipJoiningEventFromSessionExtractor.eventIdFrom(request)
           event <- EventbriteService.getBookableEvent(eventId)
         } {
-          event.service.wsMetrics.put(s"join-$tier-event", 1)
-          val memberData = MemberData(member.salesforceContactId, request.user.id, tier.name)
-          track(EventActivity("membershipRegistrationViaEvent", Some(memberData), EventData(event)))
+          if(!request.user.isTestUser) {
+            event.service.wsMetrics.put(s"join-$tier-event", 1)
```
jennysivapalan: @mattandrews this should stop being tracked in the radiator too.

jennysivapalan:
@rtyley is this any better?

rtyley:
:+1:

PR #351, opened 2015-03-19
Show members area content on joiner thank you page

davidrapson:
This PR shows an article snapshot on the joiner page if the user joined via members area content on gu.com.

screen shot 2015-03-19 at 10 59 56

It also refactors the matching on the event thank you page to take a new `Destination` option which can be one of the following values:

a. `None` => Basic joiner flow (e.g., from pricing page)
b. `EventDestination` => User joined via an event
c `ContentDestination` => User joined via members area content on gu.com

Thanks to @rtyley for the Scala assistance.


frontend/app/model/Destination.scala
```diff
@@ -0,0 +1,16 @@
+package model
+
+import com.netaporter.uri.Uri
+import model.RichEvent.RichEvent
+
+sealed trait Destination
+
+case class EventDestination(event: RichEvent, iframeUrl: Uri) extends Destination {
+  def iframeHeight = {
```
rtyley: This can be a `val` (not a `def`), because it never changes in the scope of this class!

rtyley:
:+1: apart from a [minor tweak](https://github.com/guardian/membership-frontend/pull/351#discussion_r26749229)

PR #353, opened 2015-03-19
Add social profiles

mattandrews:
This, eventually, should give us something like this: 

screen shot 2015-03-19 at 12 23 29

I don't think we'll get the description text as that's drawn from Wikipedia, but it should link to the social media profiles.

More info: https://developers.google.com/structured-data/customize/social-profiles

Test the output script tag here: https://developers.google.com/structured-data/testing-tool/ (I've confirmed it passes locally)

cc @rtyley / @feedmypixel

### DIFF
--- a/frontend/app/configuration/Config.scala
+++ b/frontend/app/configuration/Config.scala
@@ -152,6 +152,9 @@ object Config {
   val googleAnalyticsTrackingId = config.getString("google.analytics.tracking.id")
 
   val youtubeMembershipVerificationId = config.getString("youtube.membership.verification.id")
+  val youtubeUrl = config.getString("youtube.url")
+
+  val googleplusUrl = config.getString("googleplus.url")
 
   val facebookJoinerConversionTrackingId =
     Tier.allPublic.map { tier => tier -> config.getString(s"facebook.joiner.conversion.${tier.slug}") }.toMap
--- a/frontend/app/configuration/Config.scala
+++ b/frontend/app/configuration/Config.scala
@@ -85,7 +85,6 @@ object Config {
 
   val facebookAppId = config.getString("facebook.app.id")
 
-
   val touchpointDefaultBackend = touchpointBackendConfigFor("default")
   val touchpointTestBackend = touchpointBackendConfigFor("test")
 
@@ -152,9 +151,6 @@ object Config {
   val googleAnalyticsTrackingId = config.getString("google.analytics.tracking.id")
 
   val youtubeMembershipVerificationId = config.getString("youtube.membership.verification.id")
-  val youtubeUrl = config.getString("youtube.url")
-
-  val googleplusUrl = config.getString("googleplus.url")
 
   val facebookJoinerConversionTrackingId =
     Tier.allPublic.map { tier => tier -> config.getString(s"facebook.joiner.conversion.${tier.slug}") }.toMap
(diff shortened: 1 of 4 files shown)


frontend/app/configuration/Config.scala
```diff
@@ -152,6 +152,9 @@ object Config {
   val googleAnalyticsTrackingId = config.getString("google.analytics.tracking.id")
 
   val youtubeMembershipVerificationId = config.getString("youtube.membership.verification.id")
+  val youtubeUrl = config.getString("youtube.url")
+
+  val googleplusUrl = config.getString("googleplus.url")
```
rtyley: These don't need to be configuration! Maybe have a `Social`  object which just has these as constant val fields.

mattandrews: As in, just define that object directly in `Config.scala`?

rtyley: I would put it in it's own file (`Social.scala`), which could live under `frontend/app/configuration` if you like! They're not really configuration, just constants.

feedmypixel:
Can we get a WikiPedia page done for Membership?

frontend/app/views/main.scala.html
```diff
@@ -25,6 +25,21 @@
...
+            {
+                "@@context": "http://schema.org",
+                "@@type": "Organization",
+                "name": "Guardian Membership",
+                "url": "@(Config.membershipUrl)",
+                "logo": "@(Config.membershipUrl)@Asset.at("images/favicons/152x152.png")",
+                "sameAs" : [
+                    "http://www.twitter.com/@(Config.twitterUsername)",
+                    "@(Config.youtubeUrl)",
+                    "@(Config.googleplusUrl)"
```
feedmypixel: a style thing but these could be s"${Config.googleplusUrl}"

rtyley: Actually, for templates I think "@(Config.googleplusUrl)" is actually better (it's shorter, less noise) - though the `s` prefix for string interpolation is a nice feature for Scala code files.

http://docs.scala-lang.org/overviews/core/string-interpolation.html

frontend/app/configuration/Config.scala
mattandrews: Done!

rtyley:
> Can we get a Wikipedia page done for Membership?

It could fit as a 'subsite' in this article: http://en.wikipedia.org/wiki/Theguardian.com#Subsites ...though obviously membership is more than just a website, so maybe that's not the most appropriate.

frontend/app/configuration/Social.scala
```diff
@@ -0,0 +1,8 @@
+package configuration
+import configuration.Config
+
+object Social {
+  val youtube = "https://www.youtube.com/user/guardianmembership"
+  val googleplus = "https://plus.google.com/+guardianmembership/"
+  val twitter = s"http://www.twitter.com/${Config.twitterUsername}"
```
rtyley: I think we could probably remove `twitterUsername` as config too, but that can be left to another PR if you like!

rtyley:
:+1: :busts_in_silhouette: :link:

mattandrews:
Okay with you @jamesoram?

jamesoram:
:+1:

rtyley:
So far no joy on the Google search-results listing :-(

https://developers.google.com/webmasters/structured-data/testing-tool?url=https%253A%252F%252Fmembership.theguardian.com%252F

image

davidrapson:
Might be worth explicitly requesting a reindex? https://support.google.com/webmasters/answer/6065812?hl=en

PR #358, opened 2015-03-19
Remove videoWhatIsGuardianMembership from Config

davidrapson:
Doesn't feel like this needs to be config

### DIFF
--- a/frontend/app/configuration/Config.scala
+++ b/frontend/app/configuration/Config.scala
@@ -41,8 +41,6 @@ object Config {
   val membershipSupportStaffEmail = config.getString("membership.staff.email")
   val membershipDevEmail = config.getString("membership.dev.email")
 
-  val videoWhatIsGuardianMembership = config.getString("video.what.is.guardian.membership.url")
-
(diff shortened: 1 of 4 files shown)


feedmypixel:
:+1:

davidrapson:
@jamesoram Mind giving this one a quick check, just confirm that the video still displays on the about page.

jamesoram:
:+1:

jennysivapalan:
I think given it's in two places it should be defined once and used in both, the config seems fine to me?

jamesoram:
yeah I asked that but apparently the video does not have to be the same in both places

davidrapson:
I guess my thinking was it's not environment specific so doesn't hugely need to be in application.conf, and the second usage is for the pattern library an so could (and probably should) be any video.

jennysivapalan:
Okay. Anything defined in application.conf can be used for any environment but you can add specific environment overrides in files such as dev.conf or prod.conf

davidrapson:
Cheers. I'm happy to close this as it sounds like it doesn't hurt to have it in the config.

PR #359, opened 2015-03-19
import tidy up

feedmypixel:
A tidy up of the imports in our views using Intellij's *Optimize Imports tool*. I have gone through the pages affected and they look good and we are compiling so I don't see any problems here.

cc @jamesoram can you give this a thorough going over please


jamesoram:
LGTM :+1:

PR #362, opened 2015-03-19
Update promoSecondary to use ResponsiveImageGroup

davidrapson:
This PR adds a new `ResponsiveImageGenerator` object to reduce the amount of boilerplate needed to generate a sequence of `ResponsiveImage`s for a hardcoded Grid image (e.g., those use on homepage carousel, about page etc.)

It also updates the `promoSecondary` component to use this new config as an example.

The only component left using `Imager.js` is `promoPrimary` (omitted converting that here to keep the size of the PR down), once this goes in I'll raise another PR to update `promoPrimary` and remove `Imager.js` as a dependancy.

### DIFF
--- a/frontend/app/controllers/FrontPage.scala
+++ b/frontend/app/controllers/FrontPage.scala
@@ -1,6 +1,6 @@
 package controllers
 
-import model.{ResponsiveImage, ResponsiveImageGroup}
+import model.{ResponsiveImageGenerator, ResponsiveImage, ResponsiveImageGroup}
 import play.api.mvc.Controller
 
 trait FrontPage extends Controller {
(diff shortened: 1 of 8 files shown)


davidrapson:
@jennysivapalan Mind giving this on a look over as theres a fair bit of Scala in here.

frontend/app/controllers/FrontPage.scala
```diff
@@ -8,101 +8,29 @@ trait FrontPage extends Controller {
...
-            width=1000
-          ),
-          ResponsiveImage(
-            path="https://media.guim.co.uk/da867c363957d748bd04cfd9d4890033203a58c6/0_0_2279_1368/500.jpg",
-            width=500
-          )
-        )
+      ResponsiveImageGroup(
+        altText=Some("Guardian Live with Russell Brand"),
+        availableImages=ResponsiveImageGenerator("da867c363957d748bd04cfd9d4890033203a58c6/0_0_2279_1368", List(2000, 1000, 500))
```
jennysivapalan: Minor one but can the lists be defined on new line else we have to across to read them scroll.

jennysivapalan:
:+1: :key:

PR #364, opened 2015-03-20
add browse our events back onto the signed out homepage

feedmypixel:


davidrapson:
@feedmypixel Do we want this? Did some digging into analytics with David M and there has been little practical impact since removing this. Patron feels more important than events (given that the panel didn't increase traffic to /events in a meaningful way)

feedmypixel:
Lets have a natter with Jeff when he is back not sure on the reason for this but is  [#687](https://jira.gutools.co.uk/browse/MEM-687) in Jira

feedmypixel:
Confirmed with Jeff this is a request from Charles

davidrapson:
:+1:

PR #368, opened 2015-03-20
Display n tickets left

feedmypixel:
Show the amount of tickets left when there are 15 or less not sold tickets

*note the screen shot showing 77 tickets left is for illustrative purposes only*
screen shot 2015-03-20 at 15 27 57


frontend/app/model/Eventbrite.scala
```diff
@@ -187,6 +187,7 @@ object Eventbrite {
       case t: InternalTicketing => t
     }
 
+    val ticketsNotSold = capacity - ticket_classes.map(_.quantity_sold).sum
```
davidrapson: `val isLimitedAvailability = capacity - ticket_classes.map(_.quantity_sold).sum <= 15` instead maybe? Removes the logic from the template?

davidrapson: Might still need the `ticketsNotSold` if you do want to display the count in the view though.

davidrapson:
Curious to explore the thinking behind the current solution, showing it on this panel seems feels the most hidden place to show this information; especially as it's not the primary booking button.



Just as an idea, but it feels like presenting this more like we do sold out events would be nice.

screen shot 2015-03-20 at 15 47 25

I'd also want to consider changing the wording of the button to "Limited Availability" or something to that effect, but I could understand the benefits of maintaining consistent wording for the button label.

Has showing limited availability/selling quickly on the listing page been ruled out too?

screen shot 2015-03-20 at 15 48 06

frontend/app/model/Eventbrite.scala
rtyley: This variable definitely doesn't belong here, it should be on the `InternalTicketing` object - it can only be meaningful if we're using internal ticketing (ie not selling through Picturehouses or whatever).

rtyley:
Before this pull request goes out we should have a separate PR that goes out that respects the `"on_sale_status": "SOLD_OUT"` status described by @jennysivapalan below - we need to report Sold Out if Eventbrite says it's sold out, even if we _think_ there are tickets left.

```
From: Jenny Sivapalan <jennifer.sivapalan@theguardian.com>
Date: 17 March 2015 at 16:28
Subject: Re: "What's all the fuss about feminism" appears sold-out - but isn't?

We define sold out value by eventbrites value of quantity_sold > quantity_total. However looking at the API response (I've removed the key): https://www.eventbriteapi.com/v3/events/14942366019/?token=<API-KEY> it has a value on the tickets called 
"on_sale_status": "SOLD_OUT"

I'm not sure how this is set in Eventbrite but perhaps we could use that as well out quantity checks to display the sold out messaging?
```

feedmypixel:
> Is the <br> needed?

BenW requested remaining is on a separate line

davidrapson:
Closing in favour of PR https://github.com/guardian/membership-frontend/pull/379

PR #369, opened 2015-03-20
Remove hard breaks from promo blocks

davidrapson:
it seems Google is not a fan of hard breaks, particularly in the first header. This PR removes all hard `<br>`s from `promoPrimary` blocks.

screen shot 2015-03-20 at 15 58 46

It does change the wrapping of the text compared to PROD, but it's a bit naff for us to do this anyway it's better for us to just respect the natural flow of the text.

**About - New**
about - new

**About - PROD**
about - live

**Patrons - New**
patrons - new

**Patrons - PROD**
patrons - live

// @rtyley


rtyley:
:+1:

PR #370, opened 2015-03-20
Track clicks on show release dates

davidrapson:
At some point we seem to have lost click tracking on the 'Show' link for ticket release dates…or maybe we never had it.

screen shot 2015-03-20 at 17 01 52

We'll now get an event like this when someone clicks on it

screen shot 2015-03-20 at 17 01 14

// @mattandrews @davidmcdowell


feedmypixel:
:+1:

PR #371, opened 2015-03-23
Don't show mixed-content warning on past-event pages with related articles

rtyley:
"this page includes other resources which are not secure" - the side box image which was using old code to pull an image off the related article.

https://membership.theguardian.com/event/dior-and-i-preview-screening-and-director-qa-15946684964

screenshot

We've already got [fixed code](https://github.com/guardian/membership-frontend/blob/6f2e9561/frontend/app/model/ResponsiveImage.scala#L15) for this in `ResponsiveImageGroup`, which converts the image url to the secure version, so using that fixes the problem and lets us delete some code too.

cc @davidrapson

### DIFF
--- a/frontend/app/model/RichEvent.scala
+++ b/frontend/app/model/RichEvent.scala
@@ -86,7 +86,7 @@ object RichEvent {
     val tags: Seq[String]
     val metadata: Metadata
     val contentOpt: Option[Content]
-    val pastImageOpt: Option[Asset]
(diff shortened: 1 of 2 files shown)


davidrapson:
:+1: LGTM

Not sure if we should be modelling `contentOpt` and `pastImgOpt` together, as they are part of the same thing and `contentOpt` is fairly broad anyway.  Something for another PR.

rtyley:
Verified in PROD - no mixed-content warning:

image

PR #372, opened 2015-03-23
Tidy ResponsiveImageGroup code with for-comprehensions

rtyley:
cc @davidrapson


davidrapson:
So tidy. However, I extracted out the logic from RichEvent in PR https://github.com/guardian/membership-frontend/pull/362 which needs to go out 

So I might roll that in the  reapply your changes. Need to get my head around for comprehensions more.

PR #375, opened 2015-03-23
Set Cache-Control to serve *stale* cached-content on server errors

rtyley:
For our cached pages (the front of the site, the events pages) this allows Fastly to serve cached copies of the pages even when the server is down. This would have been helpful for the outage on Friday 20th February, for instance.

https://docs.fastly.com/guides/caching/does-fastly-allow-serving-old-content-while-fetching-new-content

With the static parts of the site working even with no operating servers, we've got to be aware that the uncached parts - generating an eventbrite discount for ticket purchase, or joining/upgrading as a member, would still be inoperable and require fixing ASAP!

See also: https://github.com/guardian/frontend/blob/9205d67/common/app/model/Cached.scala#L39

cc @jennysivapalan @davidrapson @mattandrews

### DIFF
--- a/frontend/app/controllers/Cached.scala
+++ b/frontend/app/controllers/Cached.scala
@@ -1,12 +1,16 @@
 package controllers
 
+import com.github.nscala_time.time.Imports._
 import org.joda.time.DateTime
 import play.api.mvc.Result
-import com.github.nscala_time.time.Imports._
+
+import scala.math.max
 
 object Cached {
 
   private val cacheableStatusCodes = Seq(200, 301, 404)
+
+  private val tenDaysInSeconds = 10.days.standardDuration.seconds
 
   def apply(result: Result): Result = apply(60)(result)
 
@@ -18,8 +22,9 @@ object Cached {
 
   private def cacheHeaders(maxAge: Int, result: Result) = {
     val now = DateTime.now
+    val staleWhileRevalidateSeconds = max(maxAge / 10, 1)
     result.withHeaders(
-      "Cache-Control" -> s"max-age=$maxAge",
+      "Cache-Control" -> s"max-age=$maxAge, stale-while-revalidate=$staleWhileRevalidateSeconds, stale-if-error=$tenDaysInSeconds",
       "Expires" -> toHttpDateTimeString(now + maxAge.seconds),
       "Date" -> toHttpDateTimeString(now)
     )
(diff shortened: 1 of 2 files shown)


davidrapson:
:+1: LGTM, although it's not 100% clear what browser support for this is like. `Stale-While-Revalidate` only seems to be in Chrome for the spec version, or if fastly handles all of that.

jennysivapalan:
:+1: :bread:

rtyley:
> LGTM, although it's not 100% clear what browser support for this is like.

My expectation would be for most (all?) browsers to ignore them, the only thing I was expecting to influence was Fastly behaviour, as per: ["Fastly can be configured to show the stale content while the new content is fetched in the background."](https://docs.fastly.com/guides/caching/does-fastly-allow-serving-old-content-while-fetching-new-content).

```
$ curl -I www.theguardian.com/uk
HTTP/1.1 200 OK
Content-Type: text/html; charset=utf-8
Expires: Tue, 24 Mar 2015 12:44:59 GMT
Server: nginx
X-Clacks-Overhead: GNU Terry Pratchett
X-Gu-Backend-App: facia
X-Gu-NG-Cached: HIT
X-Page-Speed: 1.9.32.3-4448
Via: 1.1 varnish
Cache-Control: max-age=59, stale-while-revalidate=6, stale-if-error=864000, private
Content-Length: 473945
Accept-Ranges: bytes
Date: Tue, 24 Mar 2015 12:44:24 GMT
Via: 1.1 varnish
Age: 25
Connection: keep-alive
X-Served-By: cache-lcy1120-LCY, cache-lhr6324-LHR
X-Cache: HIT, HIT
X-Cache-Hits: 1, 2
X-GU-Edition: uk
Vary: Accept-Encoding,User-Agent
X-GU-Platform: next-gen-router
```

Asked @phamann about the `private` there, he thinks it's there for a reason which makes sense for them (he couldn't remember it right now, but is going to get back tous), while it doesn't make sense for us, so we're good to go with just:

```
Cache-Control: max-age=59, stale-while-revalidate=6, stale-if-error=864000
```

rtyley:
Confirmed that the site is serving traffic fine, we'll want to test what happens when a cache page starts returning a 500 error.

rtyley:
This didn't help us during the recent Zuora outage - not even on static pages. Had a quick chat with @gklopper about this, I think it might come down to the fact we don't have a _Fastly_ healthcheck configured:

> If an origin is marked unhealthy due to health checks, Fastly will stop attempting to send requests to it. If all backends are marked unhealthy Fastly will attempt to serve stale. If no stale object is available, a 503 will be returned to the client.

https://docs.fastly.com/guides/backend-servers/health-checks-tutorial

PR #376, opened 2015-03-23
Make discount-codes clearer to users so they don't delete them

rtyley:
Madeleine on the Events team told me that users sometimes delete the promotional code when on the Eventbrite ticket page - because they think "I didn't mean to type that there!" and they assume it's something they accidentally typed. Of course, once they delete the code, their discount is gone, and they can't get their ticket.

This change means that the code looks like "MEMBER-CODE-59JUK82L" rather than just "59JUK82L", and hopefully this means users will be less likely to delete it.

*Before*
image

*After*
image

cc @davidmcdowell @davidrapson @jennysivapalan

### DIFF
--- a/frontend/app/services/DiscountCode.scala
+++ b/frontend/app/services/DiscountCode.scala
@@ -6,6 +6,7 @@ object DiscountCode {
   def generate(text: String): String = {
     val md = java.security.MessageDigest.getInstance("SHA-1")
     val digest = md.digest(text.getBytes)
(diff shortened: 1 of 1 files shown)


davidrapson:
:+1: LGTM

rtyley:
Confirmed working on PROD:

image

rtyley:
Have let Maddie know about the new codes, she's happy with them.

PR #377, opened 2015-03-23
Load pages of Eventbrite API data in parallel, not sequentially

rtyley:
The old `getPaginated` method was fun (it used the exciting `unfoldM` method! :sparkles: ) but essentially it was loading one page after another until we got all the data, and was in any case loading all pages before a result was yielded.

The new method makes an initial call to get the first page of results, then loads all the other pages concurrently (the 1st call tells it how many pages of results there will be). As Guardian Live Events & Masterclasses currently only require 2 pages or less of results each, the performance behaviour of this code should be identical to before, but it's easier to understand and should be faster if either account hits 3 or more pages of data.

The method is now called `getAll` to emphasise the data is not being lazy-loaded as you traverse over the results.

cc @jennysivapalan

### DIFF
--- a/frontend/app/services/EventbriteService.scala
+++ b/frontend/app/services/EventbriteService.scala
@@ -1,25 +1,22 @@
 package services
 
 import com.gu.membership.util.WebServiceHelper
-import play.api.cache.Cache
-
-import scala.concurrent.ExecutionContext.Implicits.global
-import scala.concurrent.duration._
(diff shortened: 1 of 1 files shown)


jennysivapalan:
:+1: :page_facing_up:

rtyley:
Verified in PROD, events and programming partner events are all still there.

PR #373, opened 2015-03-23
Respect "on_sale_status": "SOLD_OUT" from the EventBrite API

rtyley:
We want to show the event as Sold Out on our site if that's what EB will show when the user reaches the ticket selection stage.

This should be merged before we do https://github.com/guardian/membership-frontend/pull/368

cc @davidrapson @jennysivapalan

PR #379, opened 2015-03-23
Display limited availability message on events

davidrapson:
This is an amendment to PR https://github.com/guardian/membership-frontend/pull/368 after discussion with @davidmcdowell and Ben W.

**Dependant on PR https://github.com/guardian/membership-frontend/pull/373**

**Show limited availability messaging in header, like sold out or past events**
screen shot 2015-03-23 at 12 02 09

**Adjust position and messaging in sidebar**
screen shot 2015-03-23 at 12 02 19


frontend/app/model/Eventbrite.scala
```diff
@@ -187,6 +189,10 @@ object Eventbrite {
       case t: InternalTicketing => t
     }
 
+    val limitedAvailabilityText = "Last tickets remaining"
```
feedmypixel: Potentially this text could sit in the fragment even though it is repeated. Would then be more explicit?

feedmypixel:
needs a merge and a small comment bar that :+1:

frontend/app/model/Eventbrite.scala
davidrapson: I think as it's repeated it feels nice to have it in here, I think we do the same for all other status messaging too.

davidrapson:
@jamesoram Mind giving this a once over? I've set the limited availability number to ~150 so it will show on most events. You can change the following value if you want to confirm this in different scenarios https://github.com/guardian/membership-frontend/blob/display-n-tickets-left-new/frontend/app/model/Eventbrite.scala#L193

(This will be set to 15 once approved)

jamesoram:
:+1:

PR #384, opened 2015-03-23
Add Krux tag

davidrapson:
This PR adds Krux tag. Before we get this in I'd like to get a bit more information about this as this snippet requests quite a few different scripts and we already have a number of different third-party scripts on the site so want to make sure that we need everything here.


feedmypixel:
:+1: 
agreed a review of the tracking we have on the site is worth a good look at

PR #385, opened 2015-03-23
Fix omniture pageName

davidrapson:
This PR fixes the value for `pageName` sent to Omniture.

Rather than looking for `document.title` we were looking for the first `<title>` element. The addition of a new `script` tag for social links *above* the page `<title`> element mean that the Omniture setup got the title of the first icon of the SVG sprite…

That was a weird one to track down.

This PR updates the Omniture script to use the standard `document.title`;

// @feedmypixel


feedmypixel:
:+1:

PR #387, opened 2015-03-24
Fix yearly payment period string ongoing payments

feedmypixel:
- There was a bug where the second period string was not being updated. It now reads "taken on or shortly after the 24th every `month`" rather than "taken on or shortly after the 24th every `year`"
- Removed the white space between the £ and amount.

period-string


davidrapson:
:+1:

PR #389, opened 2015-03-24
DRY event-loading tasks

rtyley:
Removing duplication, because duplication is BAAAAAAAAAD. Also small change to use `traverse`:

```
Future.sequence(events.map(mkRichEvent))
```

is equivalent to:

```
Future.traverse(events)(mkRichEvent)
```

:sparkles:


rtyley:
cc @jennysivapalan

jennysivapalan:
nice :+1:

PR #396, opened 2015-03-24
Move benefit icon fragment to fragments/tier

davidrapson:
There was a lonely little benefit item fragment living on its own in a directory in fragments so I thought it would be clearer to move it to live with the other related fragments in `fragments/tier`.


feedmypixel:
:+1:

PR #397, opened 2015-03-25
uri has been removed from the Grid API. 

jennysivapalan:
We need to remove reference to this as it breaks loading of events.

### DIFF
--- a/frontend/app/model/Grid.scala
+++ b/frontend/app/model/Grid.scala
@@ -7,7 +7,7 @@ object Grid {
 
   case class Error(message: String) extends Throwable with GridObject
 
-  case class GridResult(uri: String, data: Data) extends GridObject
+  case class GridResult(data: Data) extends GridObject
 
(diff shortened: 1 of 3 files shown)


rtyley:
:+1:

cc @theefer  & https://github.com/guardian/media-service

theefer:
:+1:

Sorry :cry:

jennysivapalan:
We don't actually use it so was an easy fix.

theefer:
Cool.

FYI I'm working on a [Scala library](https://github.com/guardian/media-service/blob/master/common-lib/src/main/scala/com/gu/mediaservice/lib/argo/model/EntityReponse.scala) to help interact with [argo](https://github.com/argo-rest/spec) hypermedia APIs (argo is the media-type we use on Grid, Composer, Tag API & others).

I've just released it to PROD (in fact that's the PR that removed the `uri` and broke your integration :smile:) and will be doing more work with it until I'm happy enough to break it into a separate library, which others could then start using if they wish.

PR #398, opened 2015-03-25
Fix event vulnerabilitiy to grid parsing errors

rtyley:
Make the `GridService` return no image, rather than die on parse failure (see also #397).

cc @jennysivapalan


jennysivapalan:
:+1: :yen:

PR #399, opened 2015-03-25
Cleaunp event extra bar, fix regression

davidrapson:
This PR fixes a regression on past events.

**Before**

screen shot 2015-03-25 at 15 35 13

**After**

screen shot 2015-03-25 at 15 26 12

I've also taken the opportunity to clean up a bit:

- Extracts the dark status bar out into a separate, simpler, component. Rather than using lots of modifiers.
- Splits the event info panel (sidebar panel) out into a separate `.scss` fragment to make that separation clearer.
- Fix layout regressions

### DIFF
--- a/frontend/app/views/event/page.scala.html
+++ b/frontend/app/views/event/page.scala.html
@@ -56,38 +56,34 @@
 
     @* Limited availability messaging *@
     @if(event.isBookable && event.isLimitedAvailability) {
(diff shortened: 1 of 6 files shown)


feedmypixel:
:+1:

davidrapson:
OK with you @jamesoram ?

jamesoram:
:+1:

PR #401, opened 2015-03-25
Remove KRUX tag

davidrapson:
This PR removes Krux from the site. Currently we are getting some mixed content warnings from some of the tracking pixels requested by Krux, despite the fact that the main script is loaded over HTTPs

// @rtyley


davidrapson:
Spoken to @samdesborough on commercial team and confirmed that Krux is currently only being A/B at the moment and as gu.com.

As gu.com is non-HTTPS they are not affected by mixed content warnings, so we should disable for now.

rtyley:
Yup, we can't have mixed-content warnings :+1:

PR #403, opened 2015-03-25
Only load userzoom for non members

davidrapson:
This PR moves userzoom code to our JavaScript modules so we can control how it loads.

In this case, for the new survey we only want to load userzoom for non-members as the current survey doesn't apply to non-members.

screen shot 2015-03-25 at 17 33 56


frontend/assets/javascripts/src/utils/user.js
```diff
@@ -82,8 +82,13 @@ define([
         };
     })();
 
+    var hasTier = function(memberDetail) {
+        return !!(memberDetail && memberDetail.tier);
+    };
```
feedmypixel: this can be used else where in another PR as it is used in other areas

feedmypixel:
:+1:

frontend/assets/javascripts/src/utils/user.js
davidrapson: Yep, that was my thinking behind adding it, little more concise than doing this same check everywhere.

PR #404, opened 2015-03-25
Refactor Header.js

feedmypixel:
Split Header.js functionality into two files
> `menu.js` - Menu Listeners & Identity sign in returnUrl param population
 `menuDetails.js` Menu population details on signed in status

- clean up of JS to current membership standards
- removal of functionality and classname references no longer being used
- fixed the non working returnUrl param population


frontend/assets/javascripts/src/modules/menu.js
```diff
@@ -0,0 +1,62 @@
...
+            setIdentityCtaReturnUrl();
+        }
+    }
+
+    function addMenuListeners() {
+        bean.on(IDENTITY_MENU_CTA_ELEM, 'click', function (e) {
+            e.preventDefault();
+            e.stopImmediatePropagation();
+
+            IDENTITY_MENU_ELEM.classList.toggle('is-hidden');
```
davidrapson: Maybe turn `is-hidden` and `is-active` into constants?

frontend/assets/javascripts/src/modules/menu.js
```diff
@@ -0,0 +1,62 @@
...
+    function removeDocumentListener() {
+        bean.off(IDENTITY_MENU_ELEM, 'click');
+    }
+
+    function setIdentityCtaReturnUrl() {
+        var windowLocation = window.location;
+        var currentUrl = windowLocation.pathname + windowLocation.search;
+
+        IDENTITY_MENU_CTA_ELEM.setAttribute('href',
+            IDENTITY_MENU_CTA_ELEM.getAttribute('href').replace(/(returnUrl=[^&]+)/g, '$1' + currentUrl)
```
davidrapson: Could we break this down into a testable function, one that takes in the current href and then returns a new one?

frontend/assets/stylesheets/base/_helpers.scss
```diff
@@ -7,6 +7,7 @@
     display: none !important;
 }
 
+.is-visually-hidden,
```
davidrapson: This isn't any better than using `u-h` see: https://github.com/guardian/membership-frontend/blob/master/docs/FRONTEND.md#state-classes

davidrapson:
Would be nice to give these more specific names, e.g., `profileMenu.js` instead of just `menu.js`; could be confused with `navigation.js`.

Other than that and a few minor comments :+1:

PR #405, opened 2015-03-26
Load crazyegg through JS

davidrapson:
Following on from PR https://github.com/guardian/membership-frontend/pull/403 I'm keen to move as many of our third-party scripts to AMD components as possible so we have more control over when they are loaded.

This change also allows us to respect the `ANALYTICS_OFF_KEY` for these tags, as currently they are only limited to PROD.


feedmypixel:
:+1:  much nicer

PR #409, opened 2015-03-26
Put non-private Touchpoint-config into source-control

rtyley:
We're going to have to change a lot of Zuora rate-plan ids very soon, and I get nervous manipulating so many opaque things outside of source control where it's less easy for peopel to check my work.

Zuora rate-plan ids are not private, likewise:

* Stripe public keys - exposed to the user in their browser
* Salesforce API endpoint
* Zuora API endpoint

Usernames should not really have to be private, but I've left them out to attempt to defend against a repeated-bad-password-lockout denial-of-service

Once this goes up, DEVs will need to refresh their config from S3 with new (much shorter) files - and then the private PROD config will get the same treatment.

cc @jennysivapalan

### DIFF
--- a/frontend/conf/application.conf
+++ b/frontend/conf/application.conf
@@ -59,58 +59,10 @@ twitter.username="gdnmembership"
 
 stripe.api.url="https://api.stripe.com/v1"
 
-# Touchpoint-backend environment-specific information
-touchpoint.backend {
-
-    blank {
-        stripe {
-            api.key {
-                secret = ""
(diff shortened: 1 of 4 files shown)


jennysivapalan:
:+1: :no_entry_sign: :key: :lock:

rtyley:
Verified on PROD that I could sign-in as a friend (PROD not UAT, as UAT has broken rateplans right now)

image

PR #411, opened 2015-03-27
added description for Guardin Local

feedmypixel:
event detail local style event the guardian membership


davidrapson:
:+1: (One teamcity build goes through)

PR #414, opened 2015-03-27
Add table modifier into base .table styles

feedmypixel:
All tables were using this modifier so it has been moved into .table base class

### DIFF
--- a/frontend/assets/stylesheets/base/_tables.scss
+++ b/frontend/assets/stylesheets/base/_tables.scss
@@ -44,16 +44,13 @@ td {
         border-bottom: 1px dotted $c-neutral8;
     }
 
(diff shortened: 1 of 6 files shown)


frontend/assets/stylesheets/base/_tables.scss
```diff
@@ -44,16 +44,13 @@ td {
...
     }
 
-}
-
-.table--responsive-font {
     @include mq($from: desktop) {
         th,
         td,
         thead td {
             font-size: rem(14px);
```
davidrapson: Theres probable a font-size mixin for this, so it comes from the font-size scale

davidrapson:
:+1:

PR #415, opened 2015-03-27
remove unused className

feedmypixel:
underline styles are controlled by .copy class no need for u-underline anymore


davidrapson:
:+1:

PR #417, opened 2015-03-27
Move shared links & videos out of application.conf

davidrapson:
After discussion with @rtyley this PR moves static links and video URLs out of `application.conf` and converts them to constants in new standalone config objects. This approach feels cleaner as these links don't change between environments, and now only one file needs to be updated to add new links instead of two.

Planning on doing the same with the email addresses, but will do that in another PR.


rtyley:
:+1:

`Links` is a bit generic, but I think it'll do for now, until we get so many links that some kind of sub-grouping becomes obvious.

davidrapson:
> Links is a bit generic, but I think it'll do for now, until we get so many links that some kind of sub-grouping becomes obvious.

100% agreed.

Look OK to you @jamesoram

jamesoram:
:+1:

mattandrews:
Looks sensible to me :+1:

davidrapson:
https://status.github.com/messages

PR #427, opened 2015-03-30
Remove founding member copy

davidrapson:
Founding membership ends tomorrow (31/03/2015). This PR removes all founding member copy.

PR #428, opened 2015-03-31
Update PROD ids for new non-Founder rate plans

rtyley:
Founders got the guarantee that their subscription charge would never change, non-Founders don't get that guarantee.

https://jira.gutools.co.uk/browse/MEM-697


rtyley:
See also https://github.com/guardian/membership-frontend/pull/427

PR #430, opened 2015-03-31
Revert "Remove founding member copy"

rtyley:
Reverts guardian/membership-frontend#427 - Founding Member sign-up actually ends midnight tonight.

Also tweaked language to be more explicit about the Founder-member deadline. I've had questions from some users showing the old wording was being interpreted differently to the intended deadline, and Charles Minty has signed off on this new wording.

cc @davidrapson @JuliaBellis


rtyley:
Confirmed that the tweaked wording is in place:

image

PR #433, opened 2015-04-01
Add user's Guardian Identity (if available) to their feedback email

rtyley:
Makes it a lot easier to track down issues for users if we have their Guardian Identity Id. This change doesn't require the id, it'll only be included if the user happens to be signed in at this point.

cc @jennysivapalan @JuliaBellis


jennysivapalan:
:+1: :id:

rtyley:
Confirmed, working on PROD, can send feedback signed in or signed out, Identity ID is included as appropriate:

---------- Forwarded message ----------
Date: 1 April 2015 at 10:45
Subject: Membership feedback

Category: Give general feedback
Page: https://membership.theguardian.com/feedback
Comments: Test: Identity Id should be added

Name: Roberto Tyley
Identity user: IdMinimalUser(2342342344324,Some(ewfsdfasdasd))

PR #434, opened 2015-04-01
End of founding member offer

davidrapson:
This PR removes founding member copy (again) and updates rate plan IDs.

@rtyley Mind quickly double checking these rate plan IDs look OK? I re-applied your original update so should be fine.


rtyley:
LGTM. Please confirm with Dale & Charles before merging.

davidrapson:
Have had confirmation from Dale.

davidrapson:
Confirmed on PROD

screen shot 2015-04-01 at 11 06 23
screen shot 2015-04-01 at 11 08 06

PR #435, opened 2015-04-01
Add user id to the HTTP response, for diagnostics

rtyley:
Have had confirmation from Dave Boxhall in InfoSec that it's ok to add this information.

HTTP response headers look like:

```
X-Gu-Identity-Id:176650041
X-Gu-Membership-Test-User:false
```

cc @jennysivapalan @dbgdngit


jennysivapalan:
:+1: :guardsman:

rtyley:
Confirmed, can see these in PROD:

image

PR #436, opened 2015-04-01
Fix bug with submit button disabling after error

davidrapson:
This PR, fixes a bug where the submit button would remain disabled if there was an error with the form, or with stripe payment, which prevents the user from correcting the error and re-submitting.

**Before**

stripe-failed-before mov

**After**

stripe-failed-after mov

@mattandrews @jamesoram

### DIFF
--- a/frontend/assets/javascripts/src/modules/form/helper/loader.js
+++ b/frontend/assets/javascripts/src/modules/form/helper/loader.js
@@ -20,9 +20,13 @@ define(function() {
         LOADER_ELEM.textContent = msg;
     };
 
-    var disableSubmitButton = function (isDisabled) {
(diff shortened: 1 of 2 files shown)


davidrapson:
@jamesoram As discussed added a unit test for the method which toggles the disabled state of the submit button. Have a look and then let me know if you think it's worth adding an extra step to the functional tests as well.

jamesoram:
:+1: nice

davidrapson:
@mattandrews Mind giving this a quick once over?

mattandrews:
:+1: nice fix.

davidrapson:
Confirmed on PROD using test user

PR #438, opened 2015-04-01
Karma coverage

davidrapson:
This PR adds [karma-coverage](https://github.com/karma-runner/karma-coverage) reporting to our test runner.

Current coverage isn't *too* bad but could be better. I'm not aiming for 100% just want to make sure each module has some test coverage.

screen shot 2015-04-01 at 14 08 50

Also cleans up the `.gitignore` file a little bit (ignoring coverage output and removing old `icons.scss` ignore)

@jamesoram @mattandrews


jamesoram:
good idea :+1:

PR #440, opened 2015-04-01
New homepage images

davidrapson:
This PR adds new homepage images. We are replacing all images apart from the pussy riot one and upping the number of images as requested to 7. Currently avg. time on page is ~45 seconds, with each image being shown for 6 seconds it means 7 images is reasonable.

As we now have more images I have decreased the maximum size we request for these images to `1000px`. This means scaling up slightly on desktop (to our `1140px` max-width) but I've run this past Ben W and he is happy with the slight upscaling.

(for reference we currently get 2000, 1000, 500, and 140 px crops back from the grid)

screen shot 2015-04-01 at 15 23 58

screen shot 2015-04-01 at 15 23 37

### DIFF
--- a/frontend/app/controllers/FrontPage.scala
+++ b/frontend/app/controllers/FrontPage.scala
@@ -9,45 +9,52 @@ trait FrontPage extends Controller {
 
     val slideShowImages = Seq(
       ResponsiveImageGroup(
+        altText=Some("RIP Rock and Roll? (Guardian Live event): Emmy the Great"),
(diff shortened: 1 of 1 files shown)


mattandrews:
I wonder if we should take a nod to accessibility here and make the slideshow pausable?

mattandrews:
:+1: to these changes though!

davidrapson:
Confirmed on PROD

screen shot 2015-04-08 at 12 02 04

PR #426, opened 2015-03-30
Cleanup unused images

davidrapson:
Bit of cleanup on images, mostly removing dead images:

- Removes `logo-plain.png` as this is no longer used (was used on previous old browser message)
- Removes `event-placeholder.gif` as this is no longer used (images are optional)
- Removes `wait.gif` as this is inlined in `_loader.scss`
- Moves 2 remaining loose images in root `images` into `images/common`

PR #441, opened 2015-04-01
Re-instate logos to event listings

davidrapson:
Re-instate logos to event listings. Turns out some images weren't so unused after all (See https://github.com/guardian/membership-frontend/pull/426), but they did need to be moved to use new SVGs.

screen shot 2015-04-01 at 16 10 35

screen shot 2015-04-01 at 16 10 39

@jennysivapalan


jennysivapalan:
:+1: :man: :bar_chart:

PR #442, opened 2015-04-01
Add EC2 instance-id to response header, so we can log it in CDN logs

rtyley:
Useful information for diagnosing death, the "X-EC2-instance-id" header. Has a value like "i-6c23702f" or "n/a"

I think it should be possible to remove the header from the response returned to the user by using Fastly Content Header configuration:

https://docs.fastly.com/guides/tutorials/adding-or-modifying-headers-on-http-requests-and-responses

cc @jennysivapalan

### DIFF
--- a/frontend/app/Global.scala
+++ b/frontend/app/Global.scala
@@ -1,17 +1,16 @@
+import configuration.Config
+import controllers.Cached
+import filters.{AddEC2InstanceHeader, CheckCacheHeadersFilter, Gzipper}
+import monitoring.SentryLogging
 import play.api.Application
 import play.api.mvc.Results.{InternalServerError, NotFound}
 import play.api.mvc.{RequestHeader, Result, WithFilters}
 import play.filters.csrf._
-
-import configuration.Config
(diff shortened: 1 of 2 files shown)


jennysivapalan:
:+1: :computer: 
:+1: removing the header from the response returned to the user.

davidrapson:
Jenny has next-level cryptic emoji skills.

rtyley:
I've [updated the Fastly config to remove the header from the response](https://app.fastly.com/#configure/service/1Pvg0VP5lxQCu9LDAb7IAI/diff/29,30) returned to the user. Before the Fastly config update:

```
$ curl --silent -I https://membership.theguardian.com/feedback | grep X-EC2-instance-id
X-EC2-instance-id: i-4dcb8dab
```

...after the Fastly update:

```
$ curl --silent -I https://membership.theguardian.com/feedback | grep X-EC2-instance-id
$ 
```

I think it might have removed it from the logs tho... will 'ave to check tomorrow.

PR #444, opened 2015-04-01
Fix rich data tester warnings

davidrapson:
This PR fixes a few warnings highlighted by [Google's structured data testing tool](https://developers.google.com/structured-data/testing-tool/) 

- Address was incorrectly scoped (and duplicated) so was being counted as `PostalAddress` rather than being considered part of the event
- Low price included 'Partner/Patron`
- Adds [hasMap](http://schema.org/hasMap) for the map URL to note the relation to the address

screen shot 2015-04-01 at 17 47 14

screen shot 2015-04-01 at 17 46 05

screen shot 2015-04-01 at 17 45 57

This PR fixes the immediate issues with our structured data, but I'm going to look into moving towards [JSON-LD](http://json-ld.org/) for metadata.

**Cons of current markup based approach**
- Relies on careful massaging of markup
- We loose metadata if we hide/show various components
- Easy to end up with duplicate schema markup

**Pros of JSON-LD approach**
- Single source of truth for metadata (in the `<head`>)
- Allows us to keep full event metadata in the page even for past events, [as recommended](https://developers.google.com/structured-data/rich-snippets/events#event-specific_usage_guidelines_and_policies)
- Is [Google's recommended approach](https://developers.google.com/structured-data/rich-snippets/events#event-specific_usage_guidelines_and_policies) for events

**Cons of JSON-LD**
Might be harder to handle multiple schema objects on one page (but I might be wrong about this, I've not tried it yet)

I'm going to look at a way to trial JSON-LD in another PR.

@mattandrews

### DIFF
--- a/frontend/app/views/fragments/event/addressSummary.scala.html
+++ b/frontend/app/views/fragments/event/addressSummary.scala.html
@@ -2,14 +2,14 @@
 
 <div class="event-location">
     @if(event.venue.name) {
-        <span class="event-location__detail event-location__name" itemprop="name">@event.venue.name</span>
+        <span class="event-location__detail event-location__name">@event.venue.name</span>
     }
     @for(location <- event.venue.address) {
         @if(location.city) {
-            <span class="event-location__detail event-location__address" itemprop="address" itemscope itemtype="http://schema.org/PostalAddress">
+            <span class="event-location__detail event-location__address">
                 @if(event.venue.name) {,}
-                <span itemprop="addressLocality">@location.city</span>,
-                <span itemprop="postalCode">@location.postal_code</span>
+                @location.city,
+                @location.postal_code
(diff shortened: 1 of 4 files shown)


davidrapson:
CC @mattandrews

mattandrews:
Nice fixes, and good writeup of the pros/cons. I reckon we could go with JSON-LD in the future, as long as we figure out the multiple schema thing.

mattandrews:
:+1:

davidrapson:
Yeah, just not sure what the restrictions are yet, so need to understand this limitations a bit more first. These fixes will do for now though.

davidrapson:
Look OK to you @jamesoram ?

jamesoram:
:+1:

PR #447, opened 2015-04-02
Inform users that their browser doesn't seem able to handle our Join forn

rtyley:
We do client-side validation of form entry, so any error (including the error of not sending the Stripe token) would indicate that the Javascript has failed - probably because the browser has Javascript quirks that we don't see in the modern browsers we test against.

Ideally, we will fix all Javascript issues that unnecessarily stop users from joining, including adapting ourselves to handle the [odd behaviour of Safari 5.1](https://app.getsentry.com/the-guardian/membership/group/56798670/) on `querySelector()` - but if we _don't_ catch those issues in Javascript, we should at least display a helpful message to the user when the server-side see's there's a problem.

image

This text is just copy I put together myself, it's definitely better than the **entirely blank white page** we've been returning to users so far - we should tweak it later.

cc @davidrapson

### DIFF
--- a/frontend/app/controllers/Joiner.scala
+++ b/frontend/app/controllers/Joiner.scala
@@ -7,24 +7,23 @@ import actions._
 import com.gu.membership.salesforce.{PaidMember, ScalaforceError, Tier}
 import com.gu.membership.stripe.Stripe
 import com.gu.membership.stripe.Stripe.Serializer._
-import com.netaporter.uri.Uri
 import com.netaporter.uri.dsl._
+import com.typesafe.scalalogging.LazyLogging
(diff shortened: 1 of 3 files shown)


davidrapson:
I'd move the 'Your browser doesn't seem to be up-to-date...' message inside of `@fragments.page.pageHeader("")` rather than including an empty fragment.

:+1: Other than that.

rtyley:
Confirmed on PROD, there it is:

image

rtyley:
https://github.com/guardian/membership-frontend/pull/448 has the javascript fixes for the specific problems with Safari 5.1.

PR #448, opened 2015-04-02
Safari 5.1 compatibility fixes

davidrapson:
This PR fixes a few issues which were preventing users in Safari 5.1 from signing up as members due to JavaScript errors.

* Bonzo `.hide()` causing a fatal error, switched to toggling class names instead of using show/hide.
* Fixes layout issues due to missing `calc` support.
* Don't use `type=tel` to trigger numeric inputs: causes noise in Safari 5.1, but is also incorrect usage as we were not using for telephone fields. Prefer plain text input with numeric pattern

**Successful payment**

screen shot 2015-04-02 at 17 20 53

**Fixed layout**

safari-after

**Previous broken layout**

screen shot 2015-04-02 at 17 31 20

### DIFF
--- a/frontend/app/views/fragments/form/cardDetail.scala.html
+++ b/frontend/app/views/fragments/form/cardDetail.scala.html
@@ -68,7 +68,7 @@
 
                 <div class="credit-card-container">
                     <i class="credit-card--input-visualisation sprite-card sprite-card--unknown js-credit-card-image"></i>
-                    <input type="tel"
+                    <input type="text" pattern="[0-9]*"
                            size="20"
                            data-stripe="number"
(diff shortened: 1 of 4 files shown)


mattandrews:
All good :+1: 

Nice job figuring out the causes of these issues as well, some smart work here.

davidrapson:
@jamesoram Look OK to you?

davidrapson:
Ran functional test suite and all passed.

screen shot 2015-04-07 at 12 19 33

davidrapson:
Confirmed on PROD using test-user in Safari 5.1

screen shot 2015-04-07 at 13 05 03

PR #449, opened 2015-04-07
Subscribers will be offered 6 months free on the Partner tier

jennysivapalan:
This PR is ready for testing, however from a dev perspective of review this is going to be really hard so I'm happy to do a walk through of the code. @rtyley suggested breaking a few parts out into smaller Pull requests which I will try and do today (this shouldn't affect the testing).

The Partner page now has an optional field for subscriber ID. On entering we perform two checks:
1. [CAS] (https://github.com/guardian/content-authorisation) to identify the user is valid (expiry date of subscription must be after now).  We need to validate the subscriber ID with a postcode or last name @davidmcdowell it would be good to get some input on the messaging around this. I have test users available for both UK and international subscribers.
2. The subscriber ID has not been used on a Membership subscription. We are storing this as a custom field on the Subscription in Zuora. 

If both checks pass the user will be presented with two choices:
1. 6 months free then £15 a month
2. 6 months free then £67.50 for 6 months. After 6 months they will be charged the annual amount (£135).
 
Currently we have hard coded the 6 months but will try to make this configurable for testing.

All flows are affected by this change as we are using the same rate plans but with different dates depending on if they are a subscriber or not. For subscribers we have to send the time now plus 6 months to Zuora for the Contract Acceptance Date (also confusingly occasionally referred to as Customer Acceptance Date) to the existing rate plans in Zuora. Non subscribers have a Contract Acceptance Date of now. We keep other date fields to current date time (now).

As it wasn't clear what to do with upgrade, downgrade and cancellation flow during a subscriber period I've based it on what was easiest to deal with in Zuora. Once a subscriber is out of the free trial the behaviour should be as the same as users that pay on sign up. Below is a list of what we do when a payment is made upfront (existing flows) vs what we do on if the user is in a free period.

Upgrade flow:
User paid up front (existing flow): Make the upgrade straight away.  First payment is new tier charge minus the unused amount of the old tier - I can explain this better in person.
Free trial subscribers: Will not be able to upgrade.

Downgrade flow (users can only downgrade to Friends)
User paid up front (existing flow): Downgrade but it comes into affect on the last day of the existing subscription payment.
Free trial subscriber: Downgrade but comes into affect after the 6 month free trial.

Cancellation flow
User paid up front (existing flow): Downgrade but it comes into affect on the last day of the existing subscription payment.
Free trial subscriber: Downgrade but comes into affect after the 6 month free trial.

Before merging we must ensure:-
- custom field has been added to Zuora instances.
- Zuora has the Customer Require acceptance set to Y.
- Zuora has the rate plan for Partner tier amended to calculate the charges correctly.


frontend/app/model/Zuora.scala
```diff
@@ -55,22 +54,26 @@ object Zuora {
...
+
+    val annual = chargedThroughDate == effectiveStartDate.plusYears(1)
     val paymentPeriodLabel = if (annual) "year" else "month"
   }
 
   object SubscriptionDetails {
-    def apply(ratePlan: RatePlan, ratePlanCharge: RatePlanCharge): SubscriptionDetails = {
-      val endDate = ratePlanCharge.chargedThroughDate.getOrElse(DateTime.now)
+    def apply(subscription: Subscription, ratePlan: RatePlan, ratePlanCharge: RatePlanCharge): SubscriptionDetails = {
+      //val endDate = ratePlanCharge.chargedThroughDate.getOrElse(DateTime.now)
```
dominickendrick: hidden in plain sight

jennysivapalan: Eeek could you tell I was guessing then?!

frontend/app/services/SubscriptionService.scala
```diff
@@ -132,4 +161,46 @@ class SubscriptionService(val tierPlanRateIds: Map[ProductRatePlan, String], val
...
+    }
+
+    def getPaymentSummaryForMembershipsNoFreePeriod(memberId: MemberId) = {
+      for {
+        paymentSummary <- getPaymentSummary(memberId)
+
+      } yield MembershipSummary(paymentSummary.current.serviceStartDate, paymentSummary.current.serviceEndDate,
+        paymentSummary.totalPrice, paymentSummary.current.price, paymentSummary.current.price, paymentSummary.current.nextPaymentDate, initialFreePeriodOffer = false)
+    }
+    for {
```
dominickendrick: A line break here would make it slightly more readable ...

frontend/app/model/Zuora.scala
jennysivapalan: Nothing to see here.

jennysivapalan:
@jamesoram I've put a fix in for https://jira.gutools.co.uk/browse/MEM-713 on this commit https://github.com/guardian/membership-frontend/commit/88444c2e66d7c003db1eb144e2509abdfc2328b9

frontend/assets/javascripts/src/modules/form/subscriber.js
```diff
@@ -0,0 +1,104 @@
...
+            bean.on(SUBSCRIBER_ID_SUBMIT_ELEM, 'click', function(event) {
+
+                event.preventDefault();
+
+                var subscriberId = SUBSCRIBER_ID_INPUT_ELEM.value,
+                    postcode = POSTCODE_ELEM.value,
+                    lastName = LAST_NAME_ELEM.value;
+
+                ajax({
+                    url: '/user/check-subscriber?id='+ subscriberId + '&lastName=' + lastName + buildQueryString(postcode)
```
davidrapson: Looking at this now you've changed the requirements it would probably make sense to move all of this into `buildQueryString` rather than just the last property.

frontend/assets/javascripts/src/modules/form/subscriber.js
```diff
@@ -0,0 +1,104 @@
...
+        if (SUBSCRIBER_ID_INPUT_ELEM && SUBSCRIBER_ID_SUBMIT_ELEM) {
+            bean.on(SUBSCRIBER_ID_SUBMIT_ELEM, 'click', function(event) {
+
+                event.preventDefault();
+
+                var subscriberId = SUBSCRIBER_ID_INPUT_ELEM.value,
+                    postcode = POSTCODE_ELEM.value,
+                    lastName = LAST_NAME_ELEM.value;
+
+                ajax({
```
davidrapson: It's worth putting a check around this to make sure `subscriberId` isn't empty. No point making an XHR request that we no is going to fail. e.g.,

```
if(subscriberId !== '') {
    ajax…
} else {
    display.toggleErrorState({
        isValid: false,
        elem: SUBSCRIBER_ID_INPUT_ELEM,
        msg: 'Please provide your subscriber ID'
    });
}

PR #451, opened 2015-04-08
Add a test metric to isolate tracking issue

mattandrews:
cc @jennysivapalan 

We've noticed that our tracking of event sales is about half the number EventBrite say it is. We reckon this is because of the cookie check here (which, ironically, was added by me to make sure we were tracking Masterclass event sales). This change adds some more tracking independently of the cookie, which will confirm that the numbers are diverging because of the cookie check.

(background: the cookie check was added so that we knew if someone had clicked the "Buy tickets" button on Membership and then completed the purchase via EventBrite – if they did this for Masterclasses they ended up on theguardian.com (not Membership) so we serve them an "ad" on that page which checks for the cookie and if found, tracks the sale)

### DIFF
--- a/frontend/app/controllers/Event.scala
+++ b/frontend/app/controllers/Event.scala
@@ -167,6 +167,7 @@ trait Event extends Controller with ActivityTracking {
       trackAnon(EventActivity("eventThankYou", memberData, EventData(event), order.map(OrderData(_))))(request)
       event.service.wsMetrics.put("user-returned-to-thankyou-page", 1)
     }
+    event.service.wsMetrics.put("user-returned-to-thankyou-page-no-cookie", 1)
   }
 
   def thankyou(id: String, orderIdOpt: Option[String]) = MemberAction.async { implicit request =>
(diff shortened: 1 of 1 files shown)


frontend/app/controllers/Event.scala
```diff
@@ -167,6 +167,7 @@ trait Event extends Controller with ActivityTracking {
       trackAnon(EventActivity("eventThankYou", memberData, EventData(event), order.map(OrderData(_))))(request)
       event.service.wsMetrics.put("user-returned-to-thankyou-page", 1)
     }
+    event.service.wsMetrics.put("user-returned-to-thankyou-page-no-cookie", 1)
```
jennysivapalan: You could also move the `trackAnon` code to here too. You can then limit queries on the dashboard by unique `orderId` in case of the user refresh problem.

mattandrews: Good call – done.

Also, looking at Google Analytics, for the thankyou page we're seeing 325 pageviews, 302 unique – so only 8% of these calls are duplicate, not too many.

davidrapson:
@mattandrews You'll need to pull in the latest changes from master to stop failing build, see PR https://github.com/guardian/membership-frontend/pull/452

mattandrews:
Thanks; just did that (was puzzling me).

@jennysivapalan – this look okay now?

jennysivapalan:
:+1: Let's check the numbers in a day or two and see if it starts tallying!

PR #453, opened 2015-04-08
Add manual Sentry logging to third-party scripts

davidrapson:
This PR aims to give some more visibility to our [network timeout errors](https://app.getsentry.com/the-guardian/membership/group/31302609/) in Sentry. By adding explicit error handlers we can at least flag that an error definitely happened because the script failed to load.

This is definitely a symptom, rather than cause. But is at least one place we can try and get some more information from.

- Add `Raven` error handler to Ophan and Omniture script
- Remove `ophanJsUrl` from config (config was specifying AMD name, feels clearer to just inline URL in JS module)

As this touches how Omniture is loaded, attached is a screenshot of the Omniture Testing Tool confirming settings after change:

screen shot 2015-04-08 at 12 38 31

### DIFF
--- a/frontend/app/configuration/Config.scala
+++ b/frontend/app/configuration/Config.scala
@@ -147,8 +147,6 @@ object Config {
 
   val stage = config.getString("stage")
 
-  val ophanJsUrl = config.getString("ophan.js.url")
-
   val googleAuthConfig = {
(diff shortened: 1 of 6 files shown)


mattandrews:
:+1:

PR #454, opened 2015-04-08
Separate out the joiner thank you flow for members that pay at start and those in offer period.

jennysivapalan:
Separate out the joiner thank you flow from #449
The changes here reflect the different calls to Zuora depending on if the user has paid upfront or not and merging them into one summary class. Only the paid upfront calls should be called from frontend.

@jamesoram The changes here should have no impact on the current tests - would you be able to review?

cc/ @rtyley


jamesoram:
the tests still pass :+1:

jennysivapalan:
@rtyley I've updated this branch to include renewal date so that we can calculate the value more simply. It's based off the changes I made here: https://github.com/guardian/membership-frontend/pull/463

frontend/app/model/MembershipSummary.scala
```diff
@@ -0,0 +1,17 @@
...
+                             firstPaymentEndDate: DateTime,
+                             amountPaidToday: Float,
+                             planAmount: Float,
+                             nextPaymentPrice: Float,
+                             nextPaymentDate: DateTime,
+                             renewalDate: DateTime,
+                             initialFreePeriodOffer: Boolean) {
+
+
+  val annual = startDate.plusYears(1) == renewalDate
```
rtyley: I wonder if we will ever disagree with Zuora what a year is?

For instance, if someone is renewing a year after Feb 29th, will we differ on whether the `renewalDate` is Feb 28th or March 31st? This check could maybe be `math.round(Duration(startDate, renewalDate).millis.toFloat/1.year.millis) ==1`, to allow a tolerance?

jennysivapalan: Feels a bit over complicated, we use this kind of logic in a few places. Both dates are coming from Zuora which I assume is able to account for the leap date so the existing logic should work?

rtyley: > Both dates are coming from Zuora which I assume is able to account for the leap date so the existing logic should work?

I guess I wonder what the right answer is tho'? Is the answer `Feb 28th` or `March 1st` - and how do we know Zuora agrees with Joda on that!?

frontend/app/services/SubscriptionService.scala
```diff
@@ -132,4 +144,41 @@ class SubscriptionService(val tierPlanRateIds: Map[ProductRatePlan, String], val
       invoiceItems <- zuora.query[InvoiceItem](s"SubscriptionId='${subscription.current}'")
     } yield PaymentSummary(invoiceItems)
   }
+
+  def getMembershipSubscriptionSummary(memberId: MemberId): Future[MembershipSummary] = {
+
+    def futureFreeStartingPeriodOffer(memberId: MemberId) =
+      for (subscription <- getLatestSubscription(memberId))
```
rtyley: It looks like `getLatestSubscription(memberId)` gets called more than once in this method (here, and in `getPaymentSummaryForMembershipsWithFreePeriod()`), which is something to watch out for - it calls `getSubscriptions()`, which in turn hits the Zuora API: so performing superfluous duplicate calls to Zuora.

frontend/app/model/MembershipSummary.scala
jennysivapalan: Okay, well let's deal with it as a separate PR as we have at least three places that does this `plusYear` or `plusDay` logic which is not great.

jennysivapalan:
@jamesoram - @rtyley and I have done a bit more refactoring. Would it be okay to run this branch through the existing tests then I'll merge this to PROD if it's all okay.

rtyley:
There are still a few duplicate calls to `getSubscriptions` going on here - each one costing 2 calls to Zuora:

https://github.com/guardian/membership-frontend/blob/cf12aa57/frontend/app/services/SubscriptionService.scala#L87-L88

...we'll try to improve that in a later PR.

PR #455, opened 2015-04-08
Commit bower dependencies?

davidrapson:
This PR tentatively commits bower dependencies, while this is against typical recommendations and is not ideal it is currently the preferred option for Bower:

**Pros**
- Bower does not have a lock file or shrink-wrap which means running bower install on deploy or build can cause different versions to be pulled down, leading to unexpected builds.
- There is an open issue for a Bower lock file, but this has been open for ~2 years https://github.com/bower/bower/issues/505
- This matches what theguardian.com do, for similar reasons
- Committing dependencies is the current recommendation of the Bower team

> It's to lock down your dependencies so to prevent a bad dependency from breaking your app or the remote being down preventing deployment. This could happen even though you have a build step, since you probably don't thoroughly test on every build, and automated tests don't catch everything, especially not visual regressions. Also multiple developers might have different versions of a dependency. By having the dependencies committed you ensure everyone stays on the same version. I also find viewing the diff is a good way to ensure nothing malicious was introduced in the dependency tree.
>
> In the Node world npm shrinkwrap partially solves this, but doesn't yet do checksum matching. Bower currently have an open ticket to implement the same.

http://stackoverflow.com/a/17208939

**Cons**
- Bower lock file is scheduled for a future release https://github.com/bower/bower/pull/1748
- Means committing dependencies…
- OMG the commit ratio :cry: 

screen shot 2015-04-08 at 16 02 26

@rtyley @mattandrews 

Not 100% if this is worth it in our case as we have had very few issues with unexpected dependancy versions but I thought it worth considering.

cc @phamann @OliverJAsh let me know if you disagree with anything here


rtyley:
Quiet glad this didn't in, tho' I can see the motivation. @davidrapson what influenced you to close it in the end?

davidrapson:
Had a look back to see how many times we'd been hit with the different bower dependencies issue when deploying and it was < 5 times so I couldn't justify committing them. Even though it's the recommended approach it just didn't feel right.

Plus selfishly my addition count…

davidrapson:
Also, the Bower lock file feature looks much closer to actually shipping now, could even be in the next dot version so we can wait a little longer.

PR #457, opened 2015-04-09
Fix event conversion tracking

mattandrews:
We had some issues where the number of event ticket purchases we were seeing in Cloudwatch/Elasticsearch didn't tally with the EventBrite numbers (they were under 50% of what they should've been).

I investigated a bit and found that Google Analytics had the correct number of impressions to the `/event/$id/thankyou` page (eg. the same as EventBrite suggested), so we knew people were making it to this page. 

The tracking on that page depended on the user having a cookie of `mem-event-$id` which is set when they click "Buy tickets" on Membership, so I [added a test metric](https://github.com/guardian/membership-frontend/pull/451) which didn't test for this cookie, which revealed this discrepancy:

screen shot 2015-04-08 at 15 13 51

The orange dots show someone making an event purchase (regardless of whether they had the cookie) and the blue dots show people who *do* have the cookie. As you can see, there are quite a few purchases who don't have the cookie.

This PR, then, removes the requirement for the cookie – **except** for the "pixel" tracking, eg. the "ad" served on the NGW thankyou page used for Masterclasses, as we don't have any other way of knowing when a purchase began on Membership. Since we've already proven that the cookie method is unreliable, I've also changed the cookie from a session cookie (immediate expiry) to a 1 hour expiry. My working theory here was that some browsers dropped the session cookie as soon as the user navigated off Membership and onto Eventbrite, so hopefully this will keep tracking those numbers accurately.

Anyway, this is a really long-winded way of saying that I've removed the requirement for a cookie to track event sales. Thanks to @jennysivapalan for helping me with this one!

:cookie:

### DIFF
--- a/frontend/app/controllers/Event.scala
+++ b/frontend/app/controllers/Event.scala
@@ -186,7 +186,7 @@ trait Event extends Controller with ActivityTracking {
     EventbriteService.getEvent(id).map { event =>
       // only log a conversion if the user came from a membership event page
       request.cookies.get(eventCookie(event)).foreach { _ =>
-        trackConversionToThankyou(request, event, None, None, true)
+        trackConversionToThankyou(request, event, None, None)
       }
       NoContent.discardingCookies(DiscardingCookie(eventCookie(event)))
     }.getOrElse(NotFound)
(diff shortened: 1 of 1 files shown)


mattandrews:
Sanity check from @jennysivapalan and @davidrapson appreciated

frontend/app/controllers/Event.scala
```diff
@@ -189,7 +184,10 @@ trait Event extends Controller with ActivityTracking {
 
   def thankyouPixel(id: String) = NoCacheAction { implicit request =>
     EventbriteService.getEvent(id).map { event =>
-      trackConversionToThankyou(request, event, None, None)
+      // only log a conversion if the user came from a membership event page
+      request.cookies.get(eventCookie(event)).foreach { _ =>
```
davidrapson: Looks like you've got too many arguments here

mattandrews: Good spot (had an earlier change which I only partly removed...) – now fixed!

jennysivapalan:
:+1: :cookie: :japanese_ogre:

PR #458, opened 2015-04-09
Use current tier name instead of plan name for summaries

davidrapson:
This PR fixes an issues where "Non Founding Member" was being displayed as the Tier name when downgrading, cancelling or paid to paid upgrades. We don't really want to use the planName here so updating to use the tier name.

## Before

unnamed

screen shot 2015-04-09 at 12 07 58

## After

after

screen shot 2015-04-09 at 12 13 17

@jennysivapalan


mattandrews:
:+1: Looks sensible, but worth holding off until @jamesoram is back? Or could we get Jason to test it...?

davidrapson:
Yep, happy to hold off until @jamesoram is back in.

jamesoram:
:+1: nice

PR #459, opened 2015-04-09
Improve UX of video player

davidrapson:
This PR adds an overlay effect to videos. Uses the YouTube API to start playback of the video when the user clicks on the overlay.

video-2 mov

- Adds new `Video` model and models individual videos in the config
- Add `videoOverlay` component, YT API is lazyloaded only when needed
- Extract out new `videoPlayer` fragment and adds to pattern library

**Caveats**

- We have to load the YouTube iFrame API to trigger video playback
- iOS does not like us triggering playback, so there's a bit of nasty UA detection in here to only start playback on non iOS devices.

This is a medium-term solution until we can use the Guardian player over HTTPS in a couple of months. Are the trade-offs with this approach worth it?

@mattandrews

### DIFF
--- a/frontend/app/configuration/Videos.scala
+++ b/frontend/app/configuration/Videos.scala
@@ -1,6 +1,31 @@
 package configuration
 
+import com.netaporter.uri.dsl._
+import model.{Video, ResponsiveImageGroup, ResponsiveImageGenerator}
+
 object Videos {
-  val whatIsMembership = "https://www.youtube.com/embed/7JnYthFvYEk"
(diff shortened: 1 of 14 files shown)


mattandrews:
Looks nice. :+1:

davidrapson:
@jamesoram Look OK to you?

jamesoram:
:+1:

davidrapson:
Just waiting for a different placeholder image for the Supporter video from Ben W, then this can go out.

davidrapson:
Confirmed on PROD

screen shot 2015-04-13 at 12 24 30

PR #461, opened 2015-04-13
Re-add Krux

davidrapson:
We've had confirmation from Ads team that any non-https tags have been disabled so re-adding Krux. 

@jamesoram Before this goes out I'd appreciate your eyes on it, looking out for any mixed-content warnings. I've enabled Krux in DEV for now so you can test for this. Will put it behind the check again before this goes out.

### DIFF
--- a/frontend/assets/javascripts/src/modules/analytics/krux.js
+++ b/frontend/assets/javascripts/src/modules/analytics/krux.js
@@ -1,0 +1,16 @@
+/*global Raven */
+define(function() {
+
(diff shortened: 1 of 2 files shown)


davidrapson:
@jamesoram As discussed, looking to repeat something like this:

screen shot 2015-04-13 at 10 34 52

jamesoram:
I am also getting mixed-content warnings. :-1: sorry
screen shot 2015-04-13 at 11 16 12 2

davidrapson:
:-1: From me too…

PR #462, opened 2015-04-13
Fix pinch-to-zoom

davidrapson:
Removes maximum-scale=1.0 to fix lack of pinch-to-zoom


mattandrews:
Wow, this line has been around forever... can't believe I hadn't seen it before. Massive accessibility fail! :+1: to removing it.

davidrapson:
:mag: :rocket:

PR #466, opened 2015-04-13
Add new endpoint to return event data JSON

mattandrews:
This is part one of three when it comes to implementing Twitter-style cards for embedding Membership events in Composer:

1. [x] Add endpoint to Membership which returns JSON[P] event data (or an error)
2. [ ] Add support to Composer for querying Membership for valid event URLs `(in progress)`
3. [ ] Add design work to NGW to show Membership widget where embedded `(not yet started)`

One thing I could do with help with here is testing: not really sure where/how to test this. I think it's important we have it covered, as if this endpoint breaks/changes it'll break the above flow in turn. @jamesoram / @jennysivapalan, any time/tips for how/where to write these tests? I basically want to hit the endpoint and confirm the returned JSON structure matches what I expect.

### DIFF
--- a/frontend/app/controllers/Event.scala
+++ b/frontend/app/controllers/Event.scala
@@ -61,19 +61,6 @@ trait Event extends Controller with ActivityTracking {
     eventOpt.getOrElse(Redirect(Links.membershipFront))
   }
 
-  case class EmbedData(title: String,
-                       image: Option[String],
-                       venue: Option[String],
-                       location: Option[String],
-                       price: Option[String],
-                       identifier: String,
-                       start: String,
-                       end: String)
-  case class EmbedResponse(status: String, result: Option[EmbedData])
-
-  implicit val writesEmbedData = Json.writes[EmbedData]
-  implicit val writesResponse = Json.writes[EmbedResponse]
-
   /*
    * This endpoint is hit by Composer when embedding Membership events.
    * Changes here will need to be reflected in the flexible-content repo.
--- a/frontend/app/controllers/Event.scala
+++ b/frontend/app/controllers/Event.scala
@@ -84,8 +84,7 @@ trait Event extends Controller with ActivityTracking {
       end = event.end.toString(standardFormat)
     )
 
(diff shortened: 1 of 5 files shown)


frontend/app/controllers/Event.scala
```diff
@@ -56,6 +61,46 @@ trait Event extends Controller with ActivityTracking {
...
+   * Note that Composer will index this data, which is in turn indexed by CAPI.
+   * (eg. updates to event details will not be reflected post-embed)
+   */
+  def embedData(slug: String, callback: String) = CachedAction { implicit request =>
+  val standardFormat = ISODateTimeFormat.dateTime.withZoneUTC
+
+    val eventDataOpt = for {
+      id <- EBEvent.slugToId(slug)
+      event <- EventbriteService.getEvent(id)
+    } yield EmbedData(
```
davidrapson: This is a bit of a guess regarding testing, but maybe move the actual generation of the JSON object to `EBEvent` and then you can assert against those fields in `EbEventTest`?

jennysivapalan:
@mattandrews I would move the `EmbedResponse` and `EmbedData` to their own class files (perhaps in the `model` package. From here we can unit test the case classes. Happy to help you with this but if you want to try this yourself, we've written a number of deserializers that do the opposite of what you're trying. In the deserializers we create a json as a test example, then run it through the deserializer and check the case class is formed correctly. Here you could do the opposite, create a case class and create an expected json output, run the case class through the serializer and check the output matches the expected json.

mattandrews:
@jennysivapalan this look alright now?

frontend/app/controllers/Event.scala
jennysivapalan: Don't listen to me. Turns out what we were trying wasn't useful.

jennysivapalan:
:+1: :house_with_garden:

mattandrews:
Okay with you @jamesoram?

mattandrews:
Now with a test @jamesoram

jamesoram:
:+1: please make sure to use the same json in the tests for whatever consumes this endpoint

mattandrews:
Will do! thanks.

PR #468, opened 2015-04-14
Fix video overlay firefox bug for blocked players

davidrapson:
This PR fixes a bug in some versions of firefox on the new video player overlay. In new versions of firefox if your version of flash player is out of date it helpfully disables it, requiring you to opt-in to it.

This has the side effect of triggering an error when trying to play a video. By moving the click event to the player `onReady` event we don't do anything if players are blocked.

screen shot 2015-04-14 at 10 34 53

Also removes `loadJs`, should've just used `require` and `js!` plugin (was missing ho to stop it adding the `.js` extension)

Fixes https://app.getsentry.com/the-guardian/membership/group/62535814/

// @mattandrews


mattandrews:
:+1:

davidrapson:
Look OK to you @jamesoram ?

jamesoram:
:+1:

PR #469, opened 2015-04-14
Add `public` to Cache-Control header to encourage browser disk-caching

rtyley:
This follows on from a change in the Grid to encourage browser disk-caching of images:

https://github.com/guardian/grid-infra/pull/32#issuecomment-92754321

Thanks to @sihil, who tells me that (although behaviour varies) there are browsers that _would_ disk-cache an HTTP response with a simple `Cache-Control:max-age=60`, that **wouldn't** disk cache the same response if it came over HTTPS. You have to set 'public' to encourage them to do that.


davidrapson:
:+1: :+1: :+1:

PR #473, opened 2015-04-15
Make SBT resolve dependencies faster with cached resolution

rtyley:
There's a one-off-hit for the first time, after that, the resolve step is much faster.

http://www.scala-sbt.org/0.13/docs/sbt-0.13-Tech-Previews.html#Cached+resolution+%28minigraph+caching%29

http://www.scala-sbt.org/0.13/docs/Cached-Resolution.html

cc @dominickendrick


dominickendrick:
:+1:

PR #474, opened 2015-04-15
Add payment delay and casId if present

jennysivapalan:
In attempt to break out more code from #449 I've separated the code that we need to add the subscription on creating a member. The two fields are a delayedPayment and casId, both are optional so for now I've defaulted them to None. 

@rtyley is this useful? We won't see it be used for real until we add the code to allow users to enter a subscriber ID.


rtyley:
This seems fair enough to me - it's a shame it's not getting exercised, but apart from that :+1:

PR #476, opened 2015-04-15
use play-json serialisation for the SubsCheck result

rtyley:
The play-json serialisation thing allows you to avoid the ` ++ errorMsg.map(msg => Json.obj("msg" -> msg)).getOrElse(Json.obj()` thing.

Also some other minor tweaks.

cc @jennysivapalan


jennysivapalan:
:+1: :taxi:

PR #477, opened 2015-04-15
Add event card endpoint

davidrapson:
This PR adds a new endpoint to an event to return a card for embedding. e.g., https://membership.theguardian.com/event/the-observer-election-countdown-16253236869/card

This allows us to enhance a `richLink` style embed on .com with our own markup.

- Add a new `/card` endpoint
- Add new `inlineResource` method to `Asset` to allow us to inline the contents of an external stylesheet or SVG. @rtyley Appreciate your thoughts on this part in particular
- Add a new `AjaxCachedAction` common action = `Cors + CachedAction`

The final embed will look something like this:

screen shot 2015-04-15 at 16 04 44

But this endpoint returns a JSON response for use on .com:

screen shot 2015-04-15 at 17 02 58

@mattandrews


mattandrews:
Do we need to return an "error" if the event is in the past?

mattandrews:
:+1:

davidrapson:
As discussed we always want to return the event as we are enhancing an existing link, but we might want to handle sold out / past events somehow in the returned template. Will pick up with @davidmcdowell , but this can go out without that.

Look OK to you @jamesoram ?

PR #480, opened 2015-04-16
Update embed endpoint to use CORS

mattandrews:
This (by request from @theefer and @davidrapson, and indeed, the modern HTTP standard...) removes the JSONP callback from the new `/embed` endpoint in favour of CORS, and whitelists the various Composer domains so they can embed our events. Hooray!

### DIFF
--- a/frontend/app/controllers/Event.scala
+++ b/frontend/app/controllers/Event.scala
@@ -66,7 +66,7 @@ trait Event extends Controller with ActivityTracking {
    * Note that Composer will index this data, which is in turn indexed by CAPI.
    * (eg. updates to event details will not be reflected post-embed)
    */
-  def embedData(slug: String, callback: String) = CachedAction { implicit request =>
+  def embedData(slug: String) = Cors.andThen(CachedAction) { implicit request =>
     val standardFormat = ISODateTimeFormat.dateTime.withZoneUTC
 
     val eventDataOpt = for {
@@ -83,7 +83,7 @@ trait Event extends Controller with ActivityTracking {
       end = event.end.toString(standardFormat)
     )
 
-    Ok(Jsonp(callback, eventToJson(eventDataOpt)))
+    Ok(eventToJson(eventDataOpt))
   }
 
   private def eventDetail(event: RichEvent)(implicit request: RequestHeader) = {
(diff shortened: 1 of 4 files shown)


mattandrews:
cc @rtyley / @jennysivapalan / @davidrapson

frontend/app/controllers/Event.scala
```diff
@@ -66,7 +66,7 @@ trait Event extends Controller with ActivityTracking {
    * Note that Composer will index this data, which is in turn indexed by CAPI.
    * (eg. updates to event details will not be reflected post-embed)
    */
-  def embedData(slug: String, callback: String) = CachedAction { implicit request =>
+  def embedData(slug: String) = Cors.andThen(CachedAction) { implicit request =>
```
davidrapson: FYI. When https://github.com/guardian/membership-frontend/pull/477 goes out we can use `AjaxCachedAction` as a shorthand for this (should probably be named `CorsCachedAction` but there was already a `AjaxNoCacheAction` which did `Cors + NoCache`…)

davidrapson:
:+1:

davidrapson:
One minor thing, the `jsonp` import can probably go now too https://github.com/guardian/membership-frontend/blob/ma-add-embed-endpoint/frontend/app/controllers/Event.scala#L16

frontend/conf/application.conf
```diff
@@ -13,7 +13,7 @@ event.discountMultiplier=0.8
 
 content.api.key=""
 
-cors.allow.origin="https://profile.theguardian.com,http://www.theguardian.com"
+cors.allow.origin="https://profile.theguardian.com,http://www.theguardian.com,https://composer.gutools.co.uk,https://composer.code.gutools.co.uk,https://composer.qa.gutools.co.uk,https://composer.release.gutools.co.uk,https://composer.local.dev-gutools.co.uk"
```
theefer: These should be `https://composer.code.dev-gutools.co.uk` (`dev-`) for all the pre-PROD envs.

theefer:
Commented on an error in the config. Thanks for adding this!

frontend/conf/application.conf
mattandrews: Cheers – have a look now.

theefer:
:boom: :+1:

PR #481, opened 2015-04-16
Tweak Event Card styles

davidrapson:
This PR adjusts the inlined styles for the new event card. When embedded in .com we have to contend with a few specificity conflicts as the component is embedded within some `.from-content-api` content so using completely flat selectors won't work.

- Nests all selectors under `.membership-event`
- Removes any layout / margins as these are handled by .com styles

**Before**

screen shot 2015-04-16 at 13 00 19

**After**

screen shot 2015-04-16 at 13 00 07

Note: screenshots are from a local test, nothing embedded yet…

### DIFF
--- a/frontend/assets/stylesheets/event-card.scss
+++ b/frontend/assets/stylesheets/event-card.scss
@@ -9,94 +9,92 @@
    ========================================================================== */
 
 .membership-event {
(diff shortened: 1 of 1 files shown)


mattandrews:
:+1:

PR #482, opened 2015-04-16
Subscriber offer

jennysivapalan:
The final part that has been extracted from #449 is provide subscriber offer to Partners. Note this is also now behind OAuth meaning we can test and release this early but not make it available to the public until the business is ready.

We have already put into production all the work to handle free period offers, the impact on changes to tiers and the validation checks for subscriber offer. None if is used though until now.

@davidrapson I implemented your suggestions for https://github.com/guardian/membership-frontend/pull/449#discussion_r28316830 and https://github.com/guardian/membership-frontend/pull/449#discussion_r28316733 in this PR.

@jamesoram This will need a re-test with Jason. You'll need to hit test-users endpoint to ensure the OAuth is dropped into the cookie but then you should see the subscriber fields.

~~I think we still need to change the rate plan in Zuora production before we can go live with this.~~

### DIFF
--- a/frontend/app/actions/Functions.scala
+++ b/frontend/app/actions/Functions.scala
@@ -90,11 +90,13 @@ object Functions extends LazyLogging {
       override def refine[A](request: AuthRequest[A]) = Future.successful {
         //Copy the private helper method in play-googleauth to ensure the user is Google auth'd
         //see https://github.com/guardian/play-googleauth/blob/master/module/src/main/scala/com/gu/googleauth/actions.scala#L59-60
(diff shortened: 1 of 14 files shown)


jamesoram:
:+1:

davidrapson:
:+1: to the client-side work.

jennysivapalan:
RELAX @prout-bot it's on Prod AND it works :boom: 
cc @rtyley

rtyley:
I hit https://prout-bot.herokuapp.com/api/update/guardian/membership-frontend manually, which will let Prout catch up with things - once something's overdue, it doesn't keep on polling- but it should have got a kick from Riff Raff, hmm...

jennysivapalan:
ahh good to know. Also exciting that this is done! :fireworks:

rtyley:
:bomb:  :boom:  :sparkles: !

PR #483, opened 2015-04-16
Merge cross domain origins

mattandrews:
We've had issues where **local** Guardian frontend can't query our new `/card` endpoint because the CORS headers only allow **prod** Guardian frontend.

This PR merges together all of the CORS domains we support as one global list rather than per-environment.


mattandrews:
cc @rtyley / @jennysivapalan / @davidrapson

davidrapson:
:+1:

mattandrews:
Here it is working (local Identity):

screen shot 2015-04-16 at 16 21 47

rtyley:
This seems fine to me. My only slight reservation is that maybe we should have a shorter Cors list for endpoints that have slightly more sensitive data (`/user/me`) - but none of the domains listed is really problematic in that regard.

PR #484, opened 2015-04-16
Fix staff google-group config so US & Australian staff can join

rtyley:
When we added US and Australian groups in 188b2fb097 we mistakenly gave the Google Group domains as `@theguardian.com` - actually, as far as Google Apps for Business is concerned, the groups are still `@guardian.co.uk` so this would never have worked. Apologies to Ellie and all staff in Guardian America and Australia :earth_americas: :arrows_counterclockwise: :gift_heart: 

This change fixes that issue, but also removes the domain name from the config altogether, to remove the possibility of similar errors in the future. cc @davidrapson

### DIFF
--- a/frontend/app/actions/GuardianDomains.scala
+++ b/frontend/app/actions/GuardianDomains.scala
@@ -1,10 +1,12 @@
 package actions
 
+import configuration.Config.GuardianGoogleAppsDomain
+
 object GuardianDomains {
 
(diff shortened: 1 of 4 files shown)


davidrapson:
:+1: :globe_with_meridians:

jennysivapalan:
:koala: :surfer: :sunny:

PR #485, opened 2015-04-16
Add campaign code and use smallest image for event card

davidrapson:
This PR adjusts the event card embed to add an internal campaign code and use smallest image for event card. As the design for the card is locked to one grid unit wide we can use the smallest image (which is 140px) rather than the default (which is 500px).

event-embed-enhanced

Need to confirm an appropriate campaign code with @JuliaBellis 

// @mattandrews


mattandrews:
Works for me, safe to go out now even if we want to change it later. :+1:

PR #486, opened 2015-04-17
Add QA hook to price info when there is no discount

davidrapson:
Add QA hook to price info when there is no discount @jamesoram


jamesoram:
:+1: :+1: :+1: :+1: :+1:

PR #488, opened 2015-04-17
Update google group checker to use the Directory API

dominickendrick:
This PR switches our google groups checker for test-users and staff membership to use the new Google Directories API

To setup a Google OAuth Service Account I:
- generated a new service account in the Google developer console
- asked Ritesh Patel to add the following scope to the user:
`https://www.googleapis.com/auth/admin.directory.group.readonly`

This requires a Google OAuth Service Account Certificate to be downloaded from s3 to make the API requests work. You'll need to update your `~/.aws/credentials` file to use the following format
```
[membership]
aws_access_key_id = XXX
aws_secret_access_key = XX

```
Note the `[membership]` part. This is the change to make.

You'll also need to update you membership-keys.conf to make this work.

## QA
This can be tested locally on:
https://mem.thegulocal.com/test-users

You should be able to generate test users as normal.


project/Dependencies.scala
```diff
@@ -10,6 +10,8 @@ object Dependencies {
   val scalaUri = "com.netaporter" %% "scala-uri" % "0.4.6"
   val membershipCommon = "com.gu" %% "membership-common" % "0.63"
   val playGoogleAuth = "com.gu" %% "play-googleauth" % "0.1.11"
+  val googleApiClient = "com.google.api-client" % "google-api-client" % "1.19.1"
```
rtyley: Hey, there's a new version:

http://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22com.google.api-client%22%20AND%20a%3A%22google-api-client%22

project/Dependencies.scala
```diff
@@ -10,6 +10,8 @@ object Dependencies {
...
+  val googleAdminService = "com.google.apis" % "google-api-services-admin" % "directory_v1-rev32-1.16.0-rc"
```
rtyley: New dep:

http://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22com.google.apis%22%20AND%20a%3A%22google-api-services-admin-directory%22

project/Dependencies.scala
```diff
@@ -17,10 +19,11 @@ object Dependencies {
   val awsSimpleEmail = "com.amazonaws" % "aws-java-sdk-ses" % "1.9.24"
   val snowPlow = "com.snowplowanalytics" % "snowplow-java-tracker" % "0.5.2-SNAPSHOT"
   val bCrypt = "com.github.t3hnar" %% "scala-bcrypt" % "2.4"
+  val s3 =  "com.amazonaws" % "aws-java-sdk-s3" % "1.9.30"
```
rtyley: Can you extract the awsVersion out into a variable, made common with awsSimpleEmail version?

rtyley:
It looks like the Google API lib logs pretty verbosely at INFO level - I see a lot of this stuff in `logs/frontend.log`:

```
2015-04-17 11:57:35,489 [play-akka.actor.default-dispatcher-7] INFO  c.g.api.client.http.HttpTransport
```

I think you'll have to set the log-level to just 'warn' for c.g.api.client.http.HttpTransport?

frontend/app/services/GoogleDirectoryService.scala
```diff
@@ -0,0 +1,66 @@
...
+    serviceAccountCert: String)
+
+object GoogleDirectoryService {
+
+  val transport = new NetHttpTransport()
+  val jsonFactory = new JacksonFactory()
+
+  val credential = new GoogleCredential.Builder()
+    .setTransport(transport)
+    .setJsonFactory(jsonFactory)
```
rtyley: You might be able to get rid of these two (setTransport & setJsonFactory), as we're using service flow, not using a refreshing token?

dominickendrick: no, this is not possible : /

rtyley:
:+1: - tho I think we should move at least a good chunk of this code to https://github.com/guardian/play-googleauth/, to replace the obsoleted functionality there.

rtyley:
Confirmed can still access test-users on production! https://membership.theguardian.com/test-users

PR #489, opened 2015-04-17
Refactor event image to better handle no image

davidrapson:
Refactor event image to better handle no image case.

- Don't need to pass the the whole `RichEvent` in to the image fragment
- Moves `for` comprehension outside the image fragment to allow us to cope better with no image on event items.

**Before**

screen shot 2015-04-17 at 11 59 30

**After**

screen shot 2015-04-17 at 11 49 17

### DIFF
--- a/frontend/app/views/fragments/event/image.scala.html
+++ b/frontend/app/views/fragments/event/image.scala.html
@@ -1,8 +1,7 @@
 @(img: model.ResponsiveImageGroup, sizes: Option[String] = None, lazyLoad: Boolean = true)
 
 <img
-    @if(!lazyLoad){src="@img.defaultImage"}
-    @if(lazyLoad){data-srcset="@img.srcset"} else {srcset="@img.srcset"}
+    @if(lazyLoad){data-srcset="@img.srcset"} else {src="@img.defaultImage" srcset="@img.srcset"}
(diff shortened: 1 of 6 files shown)


mattandrews:
Personally I preferred the "before" because it drew more attention to @wpf500's Advanced Jazz Hands class.

:+1:

frontend/app/views/fragments/event/image.scala.html
```diff
@@ -1,12 +1,10 @@
...
-        @if(lazyLoad){data-srcset="@img.srcset"} else {srcset="@img.srcset"}
-        sizes="@sizes.getOrElse("100vw")"
-        alt="@img.altText"
-        class="responsive-img @if(lazyLoad) { js-lazyload} "
-        itemprop="image"
-    />
-}
+<img
+    @if(!lazyLoad){src="@img.defaultImage"}
+    @if(lazyLoad){data-srcset="@img.srcset"} else {srcset="@img.srcset"}
```
mattandrews: Could just express this as:

    @if(!lazyLoad){src="@img.defaultImage" srcset="@img.srcset"} else { data-srcset="@img.srcset"}

right?

davidrapson: Yep…good call.

PR #491, opened 2015-04-20
Add CODE and PREVIEW to CORS config

davidrapson:
So we can test the new Membership embed against PREVIEW we need to add this to the CORS config.

// @mattandrews


mattandrews:
:+1:

PR #492, opened 2015-04-20
Use JSON-LD 

davidrapson:
This is first-pass as using [JSON-LD for linked data](https://developers.google.com/structured-data/events/ticketers) over annotated markup / microdata.

Main benefits:

1. Metadata is separate to markup so less likely to be broken due to common markup refactoring
1. As it's modelled can be more easily tested for regressions.
1. Rich data remains even when certain markup is not rendered (useful for past events, different on-sale periods)
1. We can provide additional meta data without having to worry about the markup that's output / using empty inline meta tags

Tested against the [Structured Data Testing Tool](https://developers.google.com/structured-data/testing-tool/)

screen shot 2015-04-20 at 12 18 55

**Note**: We have to remove the markup properties in this same pull request as keeping them results in duplicate data being extracted.

I've got this to the point of feature parity with the markup based solution, but there's definitely more information we could provide using this approach.

// @mattandrews @rtyley

### DIFF
--- a/frontend/app/model/Eventbrite.scala
+++ b/frontend/app/model/Eventbrite.scala
@@ -238,10 +238,6 @@ object Eventbrite {
 
     lazy val memUrl = Config.membershipUrl + controllers.routes.Event.details(slug)
 
-    val c = Json.obj(
-      "name" -> venue.name.mkString,
-      "address" -> venue.addressLine.mkString
-    )
   }
 
   object EBEvent {
--- a/frontend/app/model/Eventbrite.scala
+++ b/frontend/app/model/Eventbrite.scala
@@ -190,6 +190,11 @@ object Eventbrite {
       case t: InternalTicketing => t
     }
 
+    val generalReleasePrice = for {
+      ticketing <- internalTicketing
+      ticket <- ticketing.generalReleaseTicketOpt
+    } yield ticket.priceText
+
     val limitedAvailabilityText = "Last tickets remaining"
     val isLimitedAvailability = internalTicketing.exists(event => event.ticketsNotSold <= Config.eventbriteLimitedAvailabilityCutoff && !event.isSoldOut)
     val ticketsNotSold = internalTicketing.map(_.ticketsNotSold)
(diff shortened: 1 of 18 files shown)


frontend/test/model/SchemaTest.scala
```diff
@@ -0,0 +1,25 @@
...
+class SchemaTest extends PlaySpecification {
+
+  "EventSchema" should {
+
+    "generate JSON-LD schema data for an event" in {
+
+      val event = GuLiveEvent(eventWithName("Test Event"), None, None)
+      val json = Json.toJson(event.schema)
+
+      (json \ "@context").as[String] === "http://schema.org"
```
davidrapson: My tests feel a little awkward here, loading in `GuLiveEvent` to a separate test, feels like I should be able to test this at a lower level.

frontend/app/model/Eventbrite.scala
```diff
@@ -225,6 +237,11 @@ object Eventbrite {
     val slug = slugify(name.text) + "-" + id
 
     lazy val memUrl = Config.membershipUrl + controllers.routes.Event.details(slug)
+
+    val c = Json.obj(
```
jennysivapalan: `c`?

mattandrews: It stands for :cool: 

:sunglasses:

davidrapson: Um, I have no idea. I'm not sure this should be here…

frontend/app/model/Schema.scala
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  // TODO: Generate all offers, not just primary
+  private def offerOpt(event: RichEvent): Option[OfferSchema] = {
+    event.generalReleasePrice.map { price =>
+      OfferSchema(event.memUrl, "primary", price, event.statusSchema)
+    }
+  }
+
+  def from(event: RichEvent): EventSchema = EventSchema(
```
jennysivapalan: This could be an `apply` method (then later you would not need to do `EventSchema.from(event)` but instead `EventSchema(event)`.

frontend/test/model/SchemaTest.scala
jennysivapalan: Yeah I think the implementation of `EventSchema.from(this)` feels a bit strange to me. You could just pass it the fields it needs in each place?

jennysivapalan:
merge conflict :twisted_rightwards_arrows:

jennysivapalan:
:no_entry_sign: :woman: :speech_balloon: :computer: is :+1:

mattandrews:
:+1: to @jennysivapalan's emoji game.

davidrapson:
:twisted_rightwards_arrows: resolved

davidrapson:
Confirmed on PROD, events still have rich data

screen shot 2015-04-23 at 11 27 04

screen shot 2015-04-23 at 11 27 15

PR #493, opened 2015-04-20
No trailing slash on cors config

davidrapson:
Defeated by a trailing `/`

@mattandrews


mattandrews:
D'oh. Sorry, I should've spotted that when approving it.

:+1:

PR #494, opened 2015-04-20
Access-Control-Allow-Origin * on event card endpoint

davidrapson:
Combining Cors / `Access-Control-Allow-Origin` headers on a cached actions results in Fastly caching arbitrary `Access-Control-Allow-Origin` headers depending on the last cached request.

For the new event card endpoint we can serve this with `Access-Control-Allow-Origin *` instead.

We could probably configure something on Fastly to vary based on the `Access-Control-Allow-Origin` but in this case there's no reason the endpoint can't be public.

screen shot 2015-04-20 at 15 27 16

@mattandrews @jennysivapalan @rtyley

### DIFF
--- a/frontend/app/actions/CommonActions.scala
+++ b/frontend/app/actions/CommonActions.scala
@@ -85,8 +85,6 @@ trait CommonActions {
 
   val PaidMemberAction = MemberAction andThen paidMemberRefiner()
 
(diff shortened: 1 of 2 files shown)


mattandrews:
Good catch. :+1:

PR #366, opened 2015-03-20
[Do not merge] Show members area content on joiner thank you page

davidrapson:
Second attempt at this. This PR shows an article snapshot on the joiner page if the user joined via members area content on gu.com.

**Content**
joiner - content

**Event**
joiner - event

**Default**
joiner - default

It also refactors the matching on the event thank you page to take a new `Destination` option which can be one of the following values:

a. `None` => Basic joiner flow (e.g., from pricing page)
b. `EventDestination` => User joined via an event
c `ContentDestination` => User joined via members area content on gu.com

Following the error yesterday I've added a check to ensure that calls to CAPI only happen for gu.com content and added a `recover` block to handle any CAPI errors.

// @rtyley

PR #495, opened 2015-04-20
Added tests around logic for Destination 

jennysivapalan:
This PR  is based off #366 and adds the tests for the Destination decider :sparkle: 

I've created a DestinationService which houses all the logic (created in #366).  The tests are in DestinationServiceTest which does a lot of mocking..sorry. We mock the request*, content API service, eventbrite service, and member service and based on all these return an event or content destination. 

I wrote test for each method in DestinationService but didn't handle cases where values could be None as I wasn't sure it would be useful. Thoughts?

* @rtyley can we pair tomorrow on changing the request work as I couldn't quite get this working.

:rocket: :new_moon_with_face:

### DIFF
--- a/frontend/app/controllers/Joiner.scala
+++ b/frontend/app/controllers/Joiner.scala
@@ -162,32 +162,6 @@ trait Joiner extends Controller with ActivityTracking with LazyLogging {
       }
   }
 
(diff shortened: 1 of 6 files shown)


rtyley:
:+1:

PR #498, opened 2015-04-22
Fix end date for Membership in free period offer 

jennysivapalan:
endDate on `user/me/details` is currently used in Guardian frontend to specify the Next Payment date on Membership details tab and End Date on the cancelled and tier change pages. For those who get a free period offer in Membership the endDate is displaying at 1 January 1970. 

I've updated the endDate on /user/me/details, if a payment has been made it is the charged through date. However if a user is in a free period then we can use the date of when the first payment will be which is the contract acceptance date.

I thought about creating a new field e.g. nextPaymentDate on user/me/details but the logic would be the same as endDate and then I'd have to update frontend too.

Before:
screen shot 2015-04-22 at 14 21 39
screen shot 2015-04-22 at 14 18 05

After:
screen shot 2015-04-22 at 14 24 09
screen shot 2015-04-22 at 14 16 50

:dancer:

### DIFF
--- a/frontend/app/controllers/User.scala
+++ b/frontend/app/controllers/User.scala
@@ -1,9 +1,10 @@
 package controllers
 
 import actions._
-import com.gu.cas.CAS.CASSuccess
+import com.gu.cas.CAS.{CASError, CASSuccess}
 import com.gu.membership.salesforce.{FreeMember, Member, PaidMember}
 import model.PaidTiers
+import model.Zuora.SubscriptionDetails
 import org.joda.time.format.ISODateTimeFormat
 import org.joda.time.{DateTime, Instant}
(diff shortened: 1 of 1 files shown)


frontend/app/controllers/User.scala
```diff
@@ -22,6 +23,7 @@ trait User extends Controller {
   val casService = CASService
 
   def me = AjaxMemberAction { implicit request =>
+    println("hello")
```
mattandrews: :wave:

jennysivapalan: :wave: :earth_asia:

davidrapson:
:no_entry_sign: :man: :date: :+1:

jennysivapalan:
@jamesoram is okay to merge?

jamesoram:
:+1:

PR #502, opened 2015-04-24
Re-add Krux

davidrapson:
Re-add Krux tracking script. Having sat down with Robert Freeman and gone through the Krux configuration we've got a setup which means we can consider using is on Membership. Krux allows all tags to be configured to only support secure connections.

Having set everything to require HTTPS  I am no longer able to repeat mixed-content warnings for first time visitors.

@jamesoram Would you mind looking at this again to confirm you can't repeat any mixed content-warnings?

**Note:** Krux is enabled in Dev to allow QA testing, will be disabled once tested.

### DIFF
--- a/frontend/assets/javascripts/src/modules/analytics/setup.js
+++ b/frontend/assets/javascripts/src/modules/analytics/setup.js
@@ -4,12 +4,14 @@ define([
     'src/modules/analytics/ga',
     'src/modules/analytics/ophan',
     'src/modules/analytics/omniture',
+    'src/modules/analytics/krux',
     'src/modules/analytics/crazyegg'
 ], function (
     cookie,
     googleAnalytics,
     ophanAnalytics,
     omnitureAnalytics,
+    krux,
     crazyegg
 ) {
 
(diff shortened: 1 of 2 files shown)


frontend/assets/javascripts/src/modules/analytics/setup.js
```diff
@@ -20,11 +22,14 @@ define([
             guardian.analyticsEnabled = false;
         }
 
+        guardian.analyticsEnabled = true;
         if (guardian.analyticsEnabled) {
             ophanAnalytics.init();
             omnitureAnalytics.init();
             googleAnalytics.init();
 
+            krux.load();
```
mattandrews: Do we want this in dev?

davidrapson: Yep, meant to mention I'd enabled in Dev so @jamesoram to test.

davidrapson:
@jamesoram Let me know when is a good time to go through this with you.

jamesoram:
:+1: (but we need to monitor mixed-content warnings)

PR #507, opened 2015-04-27
Fixed annual field on user/me/details

jennysivapalan:
This PR is best read with `w=1` parameter.

I've added a call to getMembershipSummary in ZuoraService so that we get the correct value for annual. I've swapped  start date to use membership summary. 

There are now more calls to Zuora which isn't ideal.

### DIFF
--- a/frontend/app/controllers/User.scala
+++ b/frontend/app/controllers/User.scala
@@ -38,6 +38,13 @@ trait User extends Controller {
         Future.successful(Json.obj())
     }
 
+    def endDate(subscriptionDetails: SubscriptionDetails) = {
+      subscriptionDetails.chargedThroughDate.orElse {
+        if (subscriptionDetails.inFreePeriodOffer) Some(subscriptionDetails.contractAcceptanceDate)
+        else None
+      }
+    }
+
     val futurePaymentDetails = for {
       cardDetails <- futureCardDetails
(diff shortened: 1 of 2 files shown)


jennysivapalan:
I've added this commit https://github.com/guardian/membership-frontend/commit/6b69e74e09bee4689e667d1e50e3b269423d9db5 which adds in the next payment price, next payment date and the renewal date (similar to the thankyou pages) so that we can make this information available on the Membership profile tab.
Once we update the Membership profile tab, I think we should get rid of endDate. Although I'm not sure if /user/me/details is used anywhere else.

@rtyley what do you think? I realise by adding these fields to the endpoint we're going to have to support this in the long run so not sure if it's the right thing to do.

rtyley:
:+1: to this fix going out before @jennysivapalan goes on holiday! The increasing number of Zuroa calls is problematic though- these are literally duplicate calls, it's crazy to be doing multiple identical hits during a single transaction. So this will be tech-debt to address once the fix goes out.

jennysivapalan:
:+1: Agreed - I will document today how many calls we do on sign up flow and any thoughts on if this can be improved.

jennysivapalan:
Just add this is change on Profile Membership tab (I'm working on a PR in frontend to add in next payment price and date if they next payment date and renewal date are different).

Before:
screen shot 2015-04-28 at 11 30 11

After:
screen shot 2015-04-28 at 11 27 31

PR #508, opened 2015-04-27
Update offers and competitions tag

davidrapson:
- Update CAPI query for offers & comps
- Refactor content models, separate out Offer from base content, replace `MembersOnlyContent` with `RichContent` trait and base `ContentItem` case class with a `ContentItemOffer` extension
- Some clean-up of names: rename `membersOnly*` in `GuardianContentService` to `offersAndCompetitions*` as the query is not specifically to do with members only content.

screen shot 2015-04-27 at 14 06 17

Only one item added so will likely need to wait for more before making this public.

// @jennysivapalan


jennysivapalan:
Looks good :+1: :trophy:

davidrapson:
Confirmed:

screen shot 2015-04-28 at 17 21 38

PR #509, opened 2015-04-28
Only call method if required

jennysivapalan:
I've inlined the method so that we don't call it for users that pay at start of Membership.
cc @rtyley


rtyley:
:+1:

PR #510, opened 2015-04-28
Add Dom and Raul to list of users that can access boxes.

jennysivapalan:
@dominickendrick @tudorraul you can now access boxes ....once this is merged, artifact is uploaded to S3 and err new membership deploy.
:key:


rtyley:
:+1:

PR #499, opened 2015-04-23
Update Friend rate plans in DEV & UAT for one-off zero-payment

rtyley:
These new one-off zero-payment rate plans should create fewer 'noise' invoices in salesforce (see message below). Updating in DEV & UAT for testing.

cc @jennysivapalan

---------- Forwarded message ----------
From: Renaldo
Date: 23 April 2015 at 10:35
Subject: Friend Memberships - ProductRate Plan Id change

Hi Roberto,

We're changing the product rate plan for the Friends membership.  

The reason for this is due to the large number of zero amount invoices being created in Salesforce.  In order to reduce the number of invoices, we are switching the product rate plan from a recurring monthly zero charge, to a one-time zero charge.

I've attached the new Friend Product Rate Plan Id (highlighted in Green).  

Can you please change this Product Rate Plan Id in UAT so that we can test?

PR #511, opened 2015-04-28
Update Friend rate plans in PROD for one-off zero-payment

rtyley:
These new one-off zero-payment rate plans should create fewer 'noise' invoices in salesforce - DEV & UAT already updated.

Follows on from https://github.com/guardian/membership-frontend/pull/499


rtyley:
The radiator shows that we've had Friend sign-ups complete since this went live, so looks ok!

PR #512, opened 2015-04-29
Use config code from common

rtyley:
Follows on from https://github.com/guardian/membership-common/pull/66

...because SWF needs the exact same swathe of config.

The Stripe URL is the same in all environments, so I've removed it from config, and made it a constant.


rtyley:
Verified on PROD, https://profile.theguardian.com/membership/edit looks good, config seems to be parsed fine.

PR #513, opened 2015-04-30
removing auth for 6 month subscribers

dominickendrick:
This removes google auth from the offer for subscribers.
****This should only be released on the 12th May****


davidrapson:
:+1:

PR #514, opened 2015-05-05
Add a new provider (PNG only)

mattandrews:
This provider doesn't have an SVG (!) and this needs to go out today.

I spent some time trying to model an `EventProviderLogo` case class which could indicate the filetype/filename but gave up, realising this is what the event admin tool should be doing. This, then, is a quick workaround we can throw out when have proper metadata.

Looks like this:

screen shot 2015-05-05 at 14 46 48

cc @davidrapson (who tests these PRs now @jamesoram has left us? :cry: )


frontend/app/views/event/page.scala.html
```diff
@@ -49,7 +49,12 @@ <h1 class="event-masthead__name qa-event-detail-name">@event.name.text</h1>
                     }
                 </div>
             }
-            <img class="event-header__logo" src="@Asset.at("images/providers/" + event.providerOpt.getOrElse(event.metadata.identifier) + ".svg")">
+            @if(event.providerOpt.getOrElse(event.metadata.identifier) == "5x15") {
```
davidrapson: :crying_cat_face:

mattandrews: I know, I know. Open to better suggestions!

(also, there's another place this SVG logic exists, `seriesDetail.scala.html`, but I don't think it's needed here as the entire thing is wrapped in `@if(event.providerOpt.isEmpty) {`)

davidrapson:
Reluctant :+1:

Sean (@TesterSpike) will start picking up QA testing. Tagging him in here but there's still a little setup left so we can probably merge this one in without as it's a small change.

davidrapson:
Incentive to get the admin tool off the ground…

frontend/app/views/event/page.scala.html
davidrapson: Yeah, only thing I could think of would be returning a tuple of the name + extension, rather than going full case class.

mattandrews: That was where I was going, but wasn't sure how to turn this line:

    if EBEvent.providerWhitelist.contains(provider)

... into something that would work with a tuple/class.

davidrapson: IDK, some kind of find or filter maybe `EBEvent.providerWhitelist.find(_._1 == provider)`. Probably OK to leave it as is for now.

mattandrews: Life's too short. Gonna merge it, and never look back. 

giphy

PR #515, opened 2015-05-05
Remove htmlClass, main template cleanup

davidrapson:
Bit of template cleanup:

- Removes `htmlClass` option from templates, no longer used and a bit of an anti-pattern anyway.
- Removes `@* ===== Anything in here will be hidden for old browsers ===== *@` message in `main.scala.html` as this is now a lie, we still show the page.
- Whitespace cleanup, because reasons.

Probably best viewed with https://github.com/guardian/membership-frontend/pull/515/files?w=1

@mattandrews


mattandrews:
:+1:

PR #516, opened 2015-05-05
Use checking card details instead of making payment

davidrapson:
Use **"checking card details"** instead of **"making payment"** for card processing message. Avoids explicit mention of payment which allows us to use the same message for processing subscriber offers (where they are not actually paying yet) but avoids needing to have two separate messages.

screen shot 2015-05-05 at 17 12 26

// @mattandrews

### DIFF
--- a/frontend/assets/javascripts/src/modules/form/validation/listeners.js
+++ b/frontend/assets/javascripts/src/modules/form/validation/listeners.js
@@ -28,7 +28,6 @@ define([
      */
     var addSubmitListener = function () {
         bean.on(SUBMIT_ELEM, 'click', function (e) {
-            var processingMessage = 'Processing...';
             e.preventDefault();
 
             form.elems.map(function (elem) {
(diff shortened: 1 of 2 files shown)


frontend/assets/javascripts/src/modules/form/validation/listeners.js
```diff
@@ -36,12 +35,8 @@ define([
             });
 
             if (!form.errs.length) {
-                if (form.hasPayment) {
-                    processingMessage.replace('...', ' payment...');
```
davidrapson: This is bit janky so have removed, doesn't need to replace anything just "Processing…" is fine at this stage.

davidrapson:
@TesterSpike Might be a good first item for you to test?

mattandrews:
Works for me: :+1:

mattandrews:
:credit_card: :clock11:

PR #519, opened 2015-05-07
Offers & Competitions refinements

davidrapson:
Fixes a couple of UX issues with offers & competitions flow:

**1. Allow customSignInUrl to override default Sign In returnUrl**

This allows users to click on the main Sign In link when being redirected from .com and be returned to the original article (currently they get dumped back on the current page). e.g., `https://profile.theguardian.com/signin?returnUrl=http://www.theguardian.com/membership/2015/may/06/win-a-luxurious-cream-blanket-made-from-100-cashmere-mop-2&skipConfirmation=true`

---

**2. Don't show event when on choose tier page with `?membershipAccess query string present`**

This prevents users being shown a 50/50 content/event  page if they've been via an event page before.

**Before**:

screen shot 2015-05-07 at 15 18 43

**After**:

screen shot 2015-05-07 at 15 18 52

// @mattandrews @davidmcdowell

### DIFF
--- a/frontend/assets/javascripts/src/modules/identityPopup.js
+++ b/frontend/assets/javascripts/src/modules/identityPopup.js
@@ -11,7 +11,8 @@ define([
 
     var IS_HIDDEN = 'is-hidden';
     var IS_ACTIVE = 'is-active';
-    var IDENTITY_MENU_CTA_ELEM = document.querySelector('.js-identity-menu-cta');
+    var IDENTITY_MENU_CTA_ELEM = document.querySelector('.js-identity-menu-toggle');
+    var IDENTITY_MENU_CTA_URL = document.querySelector('.js-identity-menu-url');
     var IDENTITY_MENU_ELEM = document.querySelector('.js-identity-menu');
     var HTML_ELEM = document.documentElement;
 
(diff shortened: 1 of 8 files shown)


frontend/assets/javascripts/src/modules/identityPopup.js
```diff
@@ -53,9 +54,11 @@ define([
         var windowLocation = window.location;
         var currentUrl = windowLocation.pathname + windowLocation.search;
 
-        IDENTITY_MENU_CTA_ELEM.setAttribute('href',
-            populateReturnUrl(IDENTITY_MENU_CTA_ELEM.getAttribute('href'), currentUrl)
-        );
+        if(IDENTITY_MENU_CTA_URL) {
```
davidrapson: Split the selector used for URL replacing from the toggle selector so I can just opt-out of replacing the url.

(As an aside if we actually passed `pageInfo` to all pages we could avoid needing to do this with JavaScript. Says the client-side developer)

frontend/assets/javascripts/src/modules/identityPopupDetails.js
```diff
@@ -9,7 +9,7 @@ define(['src/utils/user'], function (userUtil) {
 
     var IS_HIDDEN = 'is-hidden';
     var SIGNED_IN_TEXT = 'You are signed in as';
-    var MENU_CTA_TEXT_ELEM = document.querySelector('.js-identity-menu-cta-text');
+    var MENU_TEXT_ELEM = document.querySelector('.js-identity-menu-text');
```
davidrapson: Renamed this just to avoid some naming conflicts / confusion.

mattandrews:
Nice, looks much cleaner. :+1:

PR #521, opened 2015-05-08
Offers refinements update

davidrapson:
This a resubmission of PR https://github.com/guardian/membership-frontend/pull/519 which we took down after the spike in traffic/latency yesterday. This spike was unrelated to the changes so this is safe to go back out again.

It also cleans up an bug in the original PR with double Sign In / username wording.

unnamed

screen shot 2015-05-08 at 14 15 00

Removed the complexity by just using `js-user-displayname` to replace the contents of the Sign In link, rather than manually showing/hiding two separate containers.

---

Fixes a couple of UX issues with offers & competitions flow:

**1. Allow customSignInUrl to override default Sign In returnUrl**

This allows users to click on the main Sign In link when being redirected from .com and be returned to the original article (currently they get dumped back on the current page). e.g., `https://profile.theguardian.com/signin?returnUrl=http://www.theguardian.com/membership/2015/may/06/win-a-luxurious-cream-blanket-made-from-100-cashmere-mop-2&skipConfirmation=true`

---

**2. Don't show event when on choose tier page with `?membershipAccess query string present`**

This prevents users being shown a 50/50 content/event  page if they've been via an event page before.

**Before**:

screen shot 2015-05-07 at 15 18 43

**After**:

screen shot 2015-05-07 at 15 18 52

// @mattandrews

### DIFF
--- a/frontend/app/controllers/Joiner.scala
+++ b/frontend/app/controllers/Joiner.scala
@@ -47,21 +47,25 @@ trait Joiner extends Controller with ActivityTracking with LazyLogging {
   }
 
   def tierChooser = NoCacheAction { implicit request =>
-    val eventOpt = PreMembershipJoiningEventFromSessionExtractor.eventIdFrom(request).flatMap(EventbriteService.getBookableEvent)
-    val pageInfo = PageInfo(
-      CopyConfig.copyTitleChooseTier,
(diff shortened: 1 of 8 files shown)


mattandrews:
:+1:

PR #522, opened 2015-05-08
Switch to node-sass

davidrapson:
After chatting to @sndrs I wanted to see what improvements we'd get from switching to `libsass`.

Fairly painless switch with the main benefit being shaving off a chunk of time from our build.

## Build Time

**Before**
```
grunt compile --dev  15.33s user 0.81s system 166% cpu 9.688 total
```

**After**
```
grunt compile --dev  8.54s user 0.52s system 124% cpu 7.299 total
```

## Diff

**Spacing order:**
```scss
html input[type="button"],
input[type="reset"]
```
after:
``` scss
html input[type="button"], input[type="reset"]
```

**Units**
`padding: .3em 1em;` (ruby-sass) vs `padding: 0.3em 1em;` (libsass)

**Webfont URLs**
This PR also adds explicit versioning of `guss` components, rather than just including **all** of guss. Only real change is upgrading the version of `guss-webfonts` but in terms of actual changes to the diff this just means we now use a versioned URL.

```
pasteup.guim.co.uk/fonts/webfonts/hinting-on_kerning-on/original/GuardianTextEgyptianWeb/GuardianTextEgyptianWeb-Regular.woff2
```
after:
```
pasteup.guim.co.uk/fonts/1.0.0/hinting-cleartype/kerning-on/original/GuardianTextEgyptianWeb/GuardianTextEgyptianWeb-Regular.woff2
```

[Full diff available in this gist](https://gist.github.com/davidrapson/7c7a2f874e71ab5fecc6)

---

~~This change also means our only Ruby dependancy is `scss-lint`. It might be worth making the call to remove Ruby from our TeamCity builds as we probably don't need to run `scss-lint` as part of a TC build.~~

**This PR also removes the Ruby dependency as `scss-lint` has since been removed**

Thoughts?

@mattandrews


mattandrews:
Nice. :clock11: 

:+1:

davidrapson:
Thanks, diff looks tame but I'd like to get @TesterSpike to give it a run through before I put this out anyway.

davidrapson:
Added `time-grunt` to make the build time a little clearer too:

screen shot 2015-05-08 at 15 34 12

davidrapson:
@TesterSpike Let me know if you'd like to me to run you through this one.

davidrapson:
Verbal :+1: from @TesterSpike

TesterSpike:
:+1:

PR #523, opened 2015-05-08
Add config switch for optimizely

davidrapson:
As we've not being running optimizely tests for some time and we're unlikely to run any more in the near future it seems a waste to load it all the time.

This puts optimizely behind a config switch which is defaulted to `false`. Need to confirm with @JuliaBellis but it seems sensible to do this for now.

@mattandrews


mattandrews:
Nice.

:+1:

davidrapson:
Confirmed that Optimizely can be disabled for now.

PR #524, opened 2015-05-08
Remove unneeded schema

davidrapson:
Removes some unneeded schema:

- Remove `organization` schema from header as we already have it in a `json-ld` block
- Remove schema markup from event listings as it's only really meant for full event details, and results in a [boat load of schema validation errors](https://developers.google.com/structured-data/testing-tool?url=https%253A%252F%252Fmembership.theguardian.com%252Fevents) at the moment anyway.

There's now no `itemscope` / `itemprop` declarations in the markup. Everything we need is handled by `JSON-LD`.

### DIFF
--- a/frontend/app/views/embeds/eventCard.scala.html
+++ b/frontend/app/views/embeds/eventCard.scala.html
@@ -10,7 +10,7 @@
         @Html(embedCss)
     </style>
 }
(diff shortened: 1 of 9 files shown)


davidrapson:
// @mattandrews

mattandrews:
:+1:

PR #529, opened 2015-05-13
Add eventhawk

mattandrews:
Makes the effective frontend of https://github.com/guardian/membership-event-tool available within this app (so that repo should be deleted once this is merged). 

Available on `/staff/admin` (I think this is a bit simpler to remember/explain than "event tool").

Most of this code is taken directly from `membership-event-tool` but the template/grunt stuff is new.


mattandrews:
cc @davidrapson (you can ignore most of the CSS/HTML)

mattandrews:
Also, bizarrely, this build failed several times due to `scsslint`, but those errors didn't come up when I ran that task locally. Any ideas on this?

frontend/Gruntfile.js
```diff
@@ -114,6 +119,18 @@ module.exports = function (grunt) {
                     preserveLicenseComments: false,
                     out: '<%= dirs.publicDir.javascripts %>/tools.js'
                 }
+            },
+            compileAdmin: {
+                options: {
+                    name: 'src/modules/admin/admin',
```
davidrapson: It would be nice if we could separate the admin JS out too, so not part of our core app modules. Could just mean putting stuff in `src/admin` rather than in `src/modules/admin`.

frontend/assets/javascripts/src/modules/admin/admin.js
```diff
@@ -0,0 +1 @@
+// hello, world!
```
davidrapson: See comment on `Gruntfile`

davidrapson:
:+1: Aside from where we put the JS.

frontend/Gruntfile.js
mattandrews: Done!

mattandrews:
Reckon this is good to go in terms of strategy (eg. embedding the tool *inside* Membership) or should we wait for @rtyley's return? Any thoughts @dominickendrick?

dominickendrick:
Looks good to me :+1:

PR #531, opened 2015-05-13
Remove scss-lint

davidrapson:
**Update**:

After doing some digging it turns out that this line `exclude: 'assets/stylesheets/components/*'` in our `scss-lint.yml` has meant that we've never actually been running `scss-lint` against the bulk of our styles. The majority of our styles live in `assets/stylesheets/components/`.

Given this, and given the number of people that actually touch our CSS (~2) we can say we're not getting any real benefit from `scss-lint` so we should remove it.

@mattandrews


mattandrews:
:+1: 

Where was this change a few hours ago...

davidrapson:
@mattandrews Might want to give this a double look as I've removed `scss-lint` completely now.

mattandrews:
I'm happy with this. Especially since it wasn't running since last year...

mattandrews:
:+1:

PR #537, opened 2015-05-15
Sanitise whitespace for CAS lookup

davidrapson:
This PR fixes an issue where subscribers were submitting subscriber IDs with additional whitespace causing the CAS lookup to fail.

- Subscriber ID can have all whitespace stripped
- CAS requires whitespace in postcode so only trim whitespace for either end

**Before**
Submitting a valid CAS id with extra whitespace causes the lookup to fail

screen shot 2015-05-15 at 12 56 39

**After**
Whitespace is trimmed and the lookup is successful. Not here that the CAS ID is valid but the offer has already been reedemed.

screen shot 2015-05-15 at 12 55 28

@nlindblad @emma-p Either of you mind taking a look at this one as there is no-one else in today on my team?


tudorraul:
LGTM :+1:

PR #542, opened 2015-05-18
Use guardian.co.uk address for new test users group

davidrapson:
@mattandrews


mattandrews:
Worth adding a comment here too to remind people to use the canonical (eg. `guardian.co.uk`) domain?

mattandrews:
:+1: in any case.

PR #544, opened 2015-05-20
Calendar grid

davidrapson:
This adds an initial version of the calendar grid.

- Adds a new `WhatsOn` controller for the new pages.
- Adds initial version of calendar grid view, just basic list grouped by month.
- Refactors the secondary navigation styling and adds the ability to make an item private so we can preview the secondary nav on the new pages, but not show what's on in the main navigation.

List icon is a placeholder for now as Ben provided duplicate icons by mistake.

Page is behind staff auth so we can ship this in stages.

calendar

@mattandrews

PR #545, opened 2015-05-21
Refactor popup navigation styles

davidrapson:
Bit of refactoring following on from PR https://github.com/guardian/membership-frontend/pull/544 in the spirit of treating navigation items as discreet components.

- Merges popup styles into one module: we had `nav-popup` `nav-popup--small` `nav-popup--profile` and an unused `nav-popup--sections` yet only one usage of the popup. (I imagine this is just a hold over from .com styles)
- Removes the nested `nav/` directory and simplifies styles down to a single `navigation.scss` file and moves the unrelated `controls` into the main `components/` directory too.

screen shot 2015-05-21 at 11 27 15

screen shot 2015-05-21 at 11 27 19

// @mattandrews @TesterSpike


mattandrews:
:+1: nice. You're right, I think most of this complexity came from a copy/paste job on NGW.

PR #547, opened 2015-05-21
Calendar list

davidrapson:
Initial version of the calendar list view.

- Like  https://github.com/guardian/membership-frontend/pull/544 this is behind google auth
- Adds text label to view switch icon as @davidmcdowell and I felt it lacked enough context without the label.

Small screen is a little cramped but we only have desktop designs at the moment so haven't made too many assumptions for how that should look.

In discussion with @davidmcdowell as to whether we just have the list view as there is currently little distinction between grid and list (other than hiding the price in the grid). Putting both up so we can review them.

screen shot 2015-05-21 at 14 38 38

@mattandrews

### DIFF
--- a/frontend/app/controllers/WhatsOn.scala
+++ b/frontend/app/controllers/WhatsOn.scala
@@ -4,7 +4,6 @@ import com.github.nscala_time.time.Imports._
 import configuration.CopyConfig
 import model.RichEvent.RichEvent
 import model.{EventGroup, ContentItem, EventCollections, PageInfo}
-import org.joda.time.DateMidnight
 import play.api.mvc.Controller
 import services._
 import tracking.ActivityTracking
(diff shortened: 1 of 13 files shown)


frontend/app/model/Nav.scala
```diff
@@ -10,13 +10,8 @@ object Nav {
...
 
-  def fetchNav(url: String) = {
-    val navigationURLs = navigation.find { navItem =>
-      // Return the parent item if URL is a subnav
-      navItem.href == url || navItem.subNav.map(_.href).contains(url)
-    }
-    navigationURLs
-  }
+  def fetchNav(url: String) =
+    navigation.find(navItem => url.contains(navItem.href))
```
davidrapson: Previous logic prevented us showing the secondary nav in the list view as it doesn't need to be in the nav.

This version just matches against the url. In this case anything under `/whats-on` will return the what's on parent nav.

frontend/app/views/whatson/calendarList.scala.html
```diff
@@ -0,0 +1,43 @@
+@(groupedEvents: scala.collection.SortedMap[org.joda.time.LocalDate, scala.collection.SortedMap[org.joda.time.LocalDate, Seq[model.RichEvent.RichEvent]]], pageInfo: model.PageInfo)
```
davidrapson: This gets pretty crazy when nested like this, I'm sure there's a terser way of modelling this, cc/ @rtyley

frontend/app/controllers/WhatsOn.scala
```diff
@@ -21,6 +20,23 @@ trait WhatsOn extends Controller with ActivityTracking {
...
+  private def groupEventsByDayAndMonth(events: Seq[RichEvent]):  SortedMap[LocalDate, SortedMap[LocalDate, Seq[RichEvent]]] = {
+    val unsortedMap = groupEventsByDay(events).groupBy(_._1.withDayOfMonth(1))
+    SortedMap(unsortedMap.toSeq :_*)
+  }
+
+  private def groupEventsByMonth(events: Seq[RichEvent]): SortedMap[LocalDate, Seq[RichEvent]] = {
+    val unsortedMap = events.groupBy(_.start.toLocalDate.withDayOfMonth(1)).map {
+      case (monthDate, eventsForMonth) => monthDate -> eventsForMonth
+    }
+    SortedMap(unsortedMap.toSeq :_*)
```
mattandrews: Is this a fairly standard Scala thing? What's it doing?

frontend/app/views/whatson/calendarList.scala.html
mattandrews: Any way to break it into multiple lines, at least?

mattandrews:
:+1:

frontend/app/controllers/WhatsOn.scala
davidrapson: Supposedly yes, the `:_*` is the way of converting a sequence into variable arguments. I guess a little like converting an array for use in `.call()` in JS.

davidrapson:
Confirmed on PROD

screen shot 2015-05-22 at 11 18 44

PR #548, opened 2015-05-21
Remove unused config

davidrapson:
Removes a line of unused config for the old image tool. Not to be confused with the remaining [ordering.json](https://github.com/guardian/membership-frontend/blob/master/frontend/conf/application.conf#L30) from that tool.

@mattandrews


rtyley:
:+1:

PR #549, opened 2015-05-21
TODON'T

davidrapson:
Removes some old `TODOS` that we don't need anymore.

### DIFF
--- a/frontend/assets/javascripts/src/modules/toggle.js
+++ b/frontend/assets/javascripts/src/modules/toggle.js
@@ -9,9 +9,6 @@
 //     * data-toggle-label is optional.
 //     * data-toggle-hidden should be added to toggle elements which should be hidden on pageload
 
-/**
- * TODO: Use pure vanlilla JS for entire module
- */
 define(['$', 'bean'], function ($, bean) {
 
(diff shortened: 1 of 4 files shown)


functional-tests/src/test/scala/MasterclassTests.scala
```diff
@@ -63,21 +63,4 @@ class MasterclassTests extends BaseMembershipTest {
     }
   }
 
-  // TODO this is disabled as the feature is not yet complete
-//  feature("A user can search Masterclasses") {
-//
-//    scenarioWeb("51. A non-member can search Masterclasses") {
```
davidrapson: We've got by so far without this test, so should remove it.

frontend/assets/javascripts/src/modules/toggle.js
```diff
@@ -9,9 +9,6 @@
 //     * data-toggle-label is optional.
 //     * data-toggle-hidden should be added to toggle elements which should be hidden on pageload
 
-/**
- * TODO: Use pure vanlilla JS for entire module
```
davidrapson: Probably not worth worrying about

PR #550, opened 2015-05-21
UserDetails doesn't need constructor

davidrapson:
Bit of cleanup to `UserDetails`, doesn't need to be a constructor method (final module that was using this pattern).

@jennysivapalan


davidrapson:
cc/ @rtyley as there is no more @mattandrews :(

davidrapson:
@afiore Hey Andrea, would you mind giving this a look and giving a :+1: if it looks OK to you?

afiore:
:+1: Looks fine to me. Just out of curiosity, why do you export `_.CLASS_NAMES` instead of `CLASS_NAMES` directly? Does `_` have a specific meaning in CommonJS?

davidrapson:
@afiore Nope, it's a convention to signify that this is being exported just for tests, rather than a public API method.

afiore:
Cool, thanks

davidrapson:
Merged in latest changes a fixed some lint errors from new `eslint` setup.

screen shot 2015-06-18 at 12 07 41
---
screen shot 2015-06-18 at 12 07 55

PR #553, opened 2015-05-26
Add two providers

mattandrews:
Adds support for two new external event providers:

screen shot 2015-05-26 at 15 39 51

(for local testing: http://mem.thegulocal.com/event/member-book-event-15404175303)

screen shot 2015-05-26 at 15 39 58

(for local testing: http://mem.thegulocal.com/event/testing-comment-tags-15874749804)

Could you give this a quick check, @TesterSpike?


TesterSpike:
:+1:

PR #555, opened 2015-05-28
Fix event provider order confirmation bug

mattandrews:
Our code assumed that when a user hits the `thankyou` endpoint, they made an order via a Guardian Live event. Now we have Guardian Local events, we need to distinguish between them otherwise the call to EventBrite's order endpoint fails (auth error).

This method isn't beautiful: it requires an extra `?source=gu-local` parameter on the EventBrite return URL. @jennysivapalan spent some time trying to do this in code (eg. stepping through each EventBrite account and trying to request the order) but when this failed (with the auth error, eg. querying a Local order ID against the Live EB account) it threw an exception and the async nature of the calls made this tricky to work with.

Would appreciate your input on this change @rtyley!

### DIFF
--- a/frontend/app/controllers/Event.scala
+++ b/frontend/app/controllers/Event.scala
@@ -212,20 +212,25 @@ trait Event extends Controller with ActivityTracking {
     trackAnon(EventActivity("eventThankYou", memberData, EventData(event), order.map(OrderData(_))))(request)
   }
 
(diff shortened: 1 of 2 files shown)


mattandrews:
Closing in favour of the much simplified https://github.com/guardian/membership-frontend/pull/556

PR #556, opened 2015-05-28
Use the appropriate event provider to check order details

mattandrews:
cc @rtyley / @jennysivapalan


rtyley:
:+1:, I'm glad we don't need to add the new parameter like in #555! The underlying code in the EventServices is a bit yuck, but I don't think we really mind here.

jennysivapalan:
Ooh nice spot :+1: .. except for the build fail

mattandrews:
@TesterSpike are you okay with this going out? Happy to run through the repro steps (but it does require buying a £15 ticket, which can be refunded, if you want to test on PROD not DEV...)

PR #557, opened 2015-05-28
Removing email address from event order confirmation page

dominickendrick:
This removes the users email address from being displayed on the confirmation page, as this can be used to extract user data from our site :cry: 

**Before:** 
image

**After:**
image

The text wrapping is intentional


mattandrews:
Might be worth changing the wording to something like "your registered EventBrite email address" in case they use a different one from their Guardian one?

(also, merge in `master` and the build should pass!)

dominickendrick:
cc @davidmcdowell

mattandrews:
:+1:

PR #559, opened 2015-06-02
Add campaign tracking

mattandrews:
This PR (done in partnership with @dominickendrick) extends our ElasticSearch tracking to recognise when a user has entered the site with a `?CMP=foo` or `?INTCMP=bar` marketing campaign ID in the URL. If this string is present on any pageload event, Omniture will add it to its `s_sess` cookie. We take advantage of this and read back that cookie and parse the campaign ID. We then send this campaign code (in a field called, er, `campaignCode`) with several events in ElasticSearch: signup, cancel, upgrade, downgrade and event ticket purchase. This means we can start to graph campaigns on the dashboard in terms of revenue and conversions.

Would appreciate some of your time, @jennysivapalan and @rtyley, in just looking over this.

### DIFF
--- a/frontend/app/controllers/TierController.scala
+++ b/frontend/app/controllers/TierController.scala
@@ -14,6 +14,7 @@ import services.{IdentityApi, IdentityService, MemberService}
 import play.api.mvc.{AnyContent, Result, Controller, DiscardingCookie}
 import services.{SubscriptionService, IdentityApi, IdentityService, MemberService}
 import tracking.ActivityTracking
+import utils.CampaignCode.extractCampaignCode
 
 import scala.concurrent.Future
 
--- a/frontend/app/controllers/TierController.scala
+++ b/frontend/app/controllers/TierController.scala
@@ -11,10 +11,13 @@ import play.api.libs.concurrent.Execution.Implicits.defaultContext
 import play.api.libs.json.Json
 import play.api.mvc.{Controller, DiscardingCookie, Result}
 import services.{IdentityApi, IdentityService, MemberService}
+import play.api.mvc.{AnyContent, Result, Controller, DiscardingCookie}
+import services.{SubscriptionService, IdentityApi, IdentityService, MemberService}
+import tracking.ActivityTracking
 
 import scala.concurrent.Future
 
-trait DowngradeTier {
+trait DowngradeTier extends ActivityTracking {
   self: TierController =>
 
   def downgradeToFriend() = PaidMemberAction { implicit request =>
@@ -23,7 +26,7 @@ trait DowngradeTier {
 
   def downgradeToFriendConfirm() = PaidMemberAction.async { implicit request => // POST
     for {
-      cancelledSubscription <- request.touchpointBackend.downgradeSubscription(request.member, request.user)
+      cancelledSubscription <- request.touchpointBackend.downgradeSubscription(request.member, request.user, extractCampaignCode(request))
     } yield Redirect(routes.TierController.downgradeToFriendSummary)
   }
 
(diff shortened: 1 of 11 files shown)


mattandrews:
(if you want to test this locally, it looks like Omniture doesn't create the `s_sess` cookie in that environment, so you can create it yourself with the following value: `%20s_campaign%3Dfoo%3B`, set against `.thegulocal.com`)

mattandrews:
Oh! And: this PR also contains this branch (https://github.com/guardian/membership-frontend/pull/556) which we needed to test tracking – we need to merge that PR first, so you can ignore those commits here.

jennysivapalan:
@mattandrews might as well merge #556?

frontend/app/controllers/TierController.scala
```diff
@@ -11,10 +11,11 @@ import play.api.libs.concurrent.Execution.Implicits.defaultContext
 import play.api.libs.json.Json
 import play.api.mvc.{AnyContent, Result, Controller, DiscardingCookie}
 import services.{SubscriptionService, IdentityApi, IdentityService, MemberService}
+import tracking.ActivityTracking
 
 import scala.concurrent.Future
 
-trait DowngradeTier {
+trait DowngradeTier extends ActivityTracking {
```
jennysivapalan: No need to extend `ActivityTracking`.

jennysivapalan: Ignore me - just seen where `extractCampaignCode` method is defined.

jennysivapalan: Me again. I think the `extractCampaignCode` should be defined in this file. It's the only place that uses it? 
Unless you plan to use it somewhere else? In that case I recommend a separate file to `ActivityTracking`.

mattandrews: Any thoughts on this @dominickendrick?

mattandrews:
@jennysivapalan is this okay besides comment?

frontend/app/controllers/TierController.scala
dominickendrick: It's used in other places besides this controller. Specifically in `ActivityTracking` itself

mattandrews:
@jennysivapalan have a look now!

jennysivapalan:
:eyes: :+1:

mattandrews:
Ran the automated tests on this @dominickendrick but most of them failed (either timeouts or "element not found" errors) which I suspect wasn't down to these changes but to just test flakiness. Happy to sit and manually test flows to make sure they still work, eg:

[ ] sign up flow
[ ] upgrade flow
[ ] downgrade flow
[ ] cancel flow
[ ] ticket purchase flow

(eg. all the things this PR tracks).

Does this sound adequate to you @TesterSpike? Let me know ASAP if you can, I leave in two days!

TesterSpike:
Sounds adequate to me :)

davidrapson:
Confirmed all flows in Matt's list in DEV. @TesterSpike @jennysivapalan Happy with all of this?

davidrapson:
Ran functional tests locally, getting some failures but the failures are similar when testing against `master` so unsure whether this is just down to local idiosyncrasies or existing failures rather than changes in this branch.

**This branch**

screen shot 2015-06-17 at 12 42 26

**Master**

screen shot 2015-06-17 at 12 52 27

davidrapson:
@TesterSpike @dominickendrick I've updated a broken test due to the recent changes to the Join page, which fixes a number of functional tests but I still get a few failures locally waiting for the Membership tab on Profile. I'll need some dedicated time to look into the functional tests as a whole which feels outside the scope of this PR. Are we comfortable releasing this change with the functional tests in the their current state?

dominickendrick:
I'm happy with that.

davidrapson:
Successfully went through sign up process (as Supporter) and downgrade (to Friend) using test users.

screen shot 2015-06-17 at 15 13 08

---

screen shot 2015-06-17 at 15 13 31

PR #560, opened 2015-06-02
[Deploy on 29/06/2015] Remove 'Event Detail | ' from event pages

davidrapson:
Remove 'Event Detail | ' from event pages so we're not starting all event page titles with the same text. Affects Omniture tracking so need to check with @JuliaBellis before this goes out.

screen shot 2015-06-02 at 11 13 29

### DIFF
--- a/frontend/app/views/event/page.scala.html
+++ b/frontend/app/views/event/page.scala.html
@@ -6,7 +6,7 @@
 @import views.support.Dates._
 @import views.support.Social.eventDetail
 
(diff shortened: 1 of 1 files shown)


mattandrews:
:+1:

TesterSpike:
:+1:

davidrapson:
Have notified team about change to page titles, as it affects analytics. Releasing on Monday 29 June.

PR #562, opened 2015-06-02
Reduce z-index on roundel

davidrapson:
Stops roundel appearing over navigation on small screens. @mattandrews 

Before:
screen shot 2015-06-02 at 12 32 14

After:
screen shot 2015-06-02 at 12 32 37


mattandrews:
Although I kind of like the abrupt nature of it sitting atop the menu, this should probably go out. :+1:

PR #563, opened 2015-06-02
Improve handling of long running events

davidrapson:
Improves handling of multi-day events. Previously events were only sellable when the eventbrite status was `live`. This causes issues with pseudo multi-day events where an event may be classed as `started` but there are still some ticket types available to buy (due to them being used to model future dates).

Having spoken to @rtyley a suitable fix for this is to use the `sales_end` dates from the events `ticket_classes`. If a ticket type is still for sale then it's OK to show the booking button.

screen shot 2015-06-02 at 15 45 18

This also updates the sales end date to work with multi-day events. Previously this would show the sales end of the first ticket type.

screen shot 2015-06-02 at 15 48 37

@mattandrews

### DIFF
--- a/frontend/app/model/Eventbrite.scala
+++ b/frontend/app/model/Eventbrite.scala
@@ -150,7 +150,7 @@ object Eventbrite {
 
     val salesDates = TicketSaleDates.datesFor(eventTimes, primaryTicket)
 
-    val salesEnd = allTickets.map(_.sales_end).sorted.last
+    val salesEnd = allTickets.map(_.sales_end).max
 
     val isCurrentlyAvailableToPaidMembersOnly =
       generalReleaseTicketOpt.map(!TicketSaleDates.datesFor(eventTimes, _).tierCanBuyTicket(Tier.Friend)).getOrElse(true)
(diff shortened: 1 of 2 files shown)


frontend/app/model/Eventbrite.scala
```diff
@@ -150,7 +150,7 @@ object Eventbrite {
 
     val salesDates = TicketSaleDates.datesFor(eventTimes, primaryTicket)
 
-    val salesEnd = primaryTicket.sales_end
+    val salesEnd = allTickets.map(_.sales_end).sorted.last
```
davidrapson: Not 100% sure about this change so would appreciate some :eyes:  on it.

mattandrews:
Looks sensible enough to me but I think @rtyley or @jennysivapalan should give it a check too.

frontend/app/model/Eventbrite.scala
```diff
@@ -207,10 +207,17 @@ object Eventbrite {
     val limitedAvailabilityText = "Last tickets remaining"
     val isLimitedAvailability = internalTicketing.exists(event => event.ticketsNotSold <= Config.eventbriteLimitedAvailabilityCutoff && !event.isSoldOut)
     val ticketsNotSold = internalTicketing.map(_.ticketsNotSold)
-
     val isSoldOut = internalTicketing.exists(_.isSoldOut)
-    val isBookable = status == "live" && !isSoldOut
-    val isPastEvent = status != "live" && status != "draft"
+
+    val isBookable = {
+      val hasBookableTicketClasses = ticket_classes.exists(_.sales_end < DateTime.now)
```
davidrapson: @rtyley Updated based on our conversation, reckon we still need to compare against `sales_start` too?

frontend/app/model/Eventbrite.scala
```diff
@@ -207,10 +207,17 @@ object Eventbrite {
...
+      (status == "live" || (status == "started" && hasBookableTicketClasses)) && !isSoldOut
+    }
+
+    val isPastEvent = {
+      val conditions = Set("cancelled", "ended", "completed")
```
rtyley: Random comment: I'm not sure if a cancelled event _is_ a past event... (obviously, it probably gets cancelled before it was supposed to happen). It's like what we mean is 'an event which we are no longer going to attempt to sell tickets for'.

frontend/app/model/Eventbrite.scala
rtyley: Could we rename this variable to `isStartedAndHasBookableTicketClasses`, making it equal `status == "started" && ticket_classes.exists(_.sales_end < DateTime.now)`? The fact that it's only true when `status == "started"` lowers it's possibly disruptive impact, and the ugly name makes it clear people shouldn't mess with it.

As for comparing with `sales_start`, you can if you like, but I think it's probably not essential given we check against  `status == "started"`.

davidrapson: Sounds sensible, I've made that change.

davidrapson:
// @TesterSpike

rtyley:
:+1:

PR #564, opened 2015-06-02
Remove unused and effectively blank test

rtyley:
This class doesn't actually run any tests at all, as far as I can see.


davidrapson:
:+1: You've just invented NTDD (no test driven development)

PR #565, opened 2015-06-02
Filter out offers and competitions which start with EXPIRED

davidrapson:
Filter out offers and competitions which start with EXPIRED. @mattandrews 

**Before**

screen shot 2015-06-02 at 17 52 53

**After**

screen shot 2015-06-02 at 17 52 59

// @mattandrews

### DIFF
--- a/frontend/app/controllers/OffersAndCompetitions.scala
+++ b/frontend/app/controllers/OffersAndCompetitions.scala
@@ -6,10 +6,12 @@ import model.ContentItemOffer
 
 trait OffersAndCompetitions extends Controller {
   def list = CachedAction { implicit request =>
+
     val results = GuardianContentService.offersAndCompetitionsContent.map(ContentItemOffer)
-      .filter { item =>
-        item.content.fields.map(_("membershipAccess")).isEmpty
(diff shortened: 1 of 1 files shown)


frontend/app/controllers/OffersAndCompetitions.scala
```diff
@@ -6,10 +6,12 @@ import model.ContentItemOffer
 
 trait OffersAndCompetitions extends Controller {
   def list = CachedAction { implicit request =>
+
     val results = GuardianContentService.offersAndCompetitionsContent.map(ContentItemOffer)
-      .filter { item =>
-        item.content.fields.map(_("membershipAccess")).isEmpty
-      }.filter(_.imgOpt.nonEmpty)
+      .filter(_.content.fields.map(_("membershipAccess")).isEmpty)
+      .filter(!_.content.webTitle.startsWith("EXPIRED"))
```
davidrapson: This is the current approach (events already have EXPIRED in the title) but I don't love it. Would be ideal if there's something from CAPI that can appropriately flag that this is expired/out of date.

mattandrews:
This is pretty horrible (from a metadata POV) but the only real option here that I can see. :+1:

PR #569, opened 2015-06-08
Use eslint in place of jshint and jscs

davidrapson:
Uses `eslint` in place of `jshint` and `jscs`. Replaces three dependencies with one and removes the ambiguity between `jshint` and `jscs`.

// @rtyley


rtyley:
:+1:

Hope that linting is providing value for you!

davidrapson:
@rtyley Yep, JS linting is actually usefull…

PR #571, opened 2015-06-10
Add British Council provider

davidrapson:
screen shot 2015-06-10 at 13 51 09

// @rtyley


frontend/app/model/Eventbrite.scala
```diff
@@ -227,7 +227,7 @@ object Eventbrite {
 
     val providerOpt = for {
       desc <- description
-      m <- "<!-- provider: (\\w+) -->".r.findFirstMatchIn(desc.html)
+      m <- "<!-- provider: (.*?) -->".r.findFirstMatchIn(desc.html)
```
davidrapson: Expanded to allow for `-` in `british-council` name.

rtyley: My paranoia would prefer `\\S+`

rtyley:
:+1: , with the regex fix

PR #574, opened 2015-06-10
Add analytics event tracking to new Friend link

davidrapson:
So we can explicitly track how many people click on this through analytics @rtyley


jennysivapalan:
:+1: :chart_with_upwards_trend:

PR #578, opened 2015-06-11
Pass the PaymentGateway to the account. 

jennysivapalan:
For subscriptions we want to use Zuora and will be introducing GoCardless.  Before we can do that we have been advised to start sending through the Stripe payment gateway for Membership.

Digging into this it looks like Stripe is the default gateway so not really sure we need to make this change. However if the default does get changed in the future it's probably best we do start sending the payment gateway through.

Zuora Payment Gateway settings:
screen shot 2015-06-11 at 14 45 57

Before change:
screen shot 2015-06-11 at 14 47 00

After change:
screen shot 2015-06-11 at 14 45 44

Note I used UAT to test this as it has the same settings as PROD. In DEV the gateway is called Stripe Test Gateway 1. I've asked this to match UAT and PROD so we don't have to do any config changes. I don't have the correct credentials to make this change myself. We should hold off merging this until DEV has been updated.

cc @davidrapson @rtyley @TesterSpike


rtyley:
:+1:

jennysivapalan:
Works fine with test-users for Friend and Partner. Will keep an eye on radiator stats for Prod.

PR #580, opened 2015-06-11
Make statusText an Option

davidrapson:
Make `statusText` an `Option` as it was bugging me that one of the values was an empty string. Am I doing this Scala thing right? @rtyley


rtyley:
:+1: :+1: :+1: :+1:

davidrapson:
Confirmed on PROD

screen shot 2015-06-11 at 15 49 57

PR #581, opened 2015-06-11
Provider refactor

davidrapson:
This PR refactors how we model external providers, to remove some unpleasant forking logic and make them a little easier to work with.

- Providers are now modelled as case classes rather than strings
- Logo images are defined as part of the object rather than generated
- Providers have an additional title just to aid with readability compared to the slug/id

This PR also replaces the Idler SVG (previously 900k :scream:) with a much smaller PNG, made simple due to these changes. Ran all provider images through image optimisation again too just to squeeze them down as much as possible.

screen shot 2015-06-11 at 17 33 46

screen shot 2015-06-11 at 16 43 51

**Bonus UX improvement on event overview tool**

screen shot 2015-06-11 at 16 44 17
screen shot 2015-06-11 at 16 44 07

@jennysivapalan @rtyley


frontend/app/model/Eventbrite.scala
```diff
@@ -12,6 +12,7 @@ import play.api.data.validation.ValidationError
 import play.api.libs.functional.syntax._
 import play.api.libs.json._
 import utils.StringUtils._
+import views.support.Asset
```
davidrapson: Considering moving this out of views support as we use it in a few places outside views now…

jennysivapalan:
:+1: :haircut:

PR #584, opened 2015-06-12
Use identity-play-auth-lib extracted from membership-frontend

rtyley:
I'll probably extract a bit more code out of Membership, this is just a first chunk.

https://github.com/guardian/identity-play-auth

cc @jennysivapalan


jennysivapalan:
:+1: :id: :electric_plug:

rtyley:
Confirmed, can still sign up as a test-user - and non test-users are still correctly identified as such.

PR #585, opened 2015-06-16
Increase Eventbrite API refresh time to 15 min

davidrapson:
Increase Eventbrite API refresh time to 15 min to reduce calls due to current rate limiting. @jennysivapalan


jennysivapalan:
:+1: :hourglass_flowing_sand:

PR #588, opened 2015-06-18
Configure live-reload so it doesn't reload entire page.

davidrapson:
We previously had live-reload configured for source assets, rather than the built files which causes full page reloads. This updates the configuration to only reload the actual stylesheet that's changed (we don't need live-reload on JS).

Also removes linting on the `Gruntfile` because it's just annoying in this case.

// @rtyley

### DIFF
--- a/frontend/Gruntfile.js
+++ b/frontend/Gruntfile.js
@@ -200,21 +200,26 @@ module.exports = function (grunt) {
          ***********************************************************************/
 
         watch: {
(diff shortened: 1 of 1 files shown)


davidrapson:
@afiore Mind giving this a look?

afiore:
:+1: It sounds sensible, also fine with not linting the GruntFile, as it's only run via NodeJS.

PR #589, opened 2015-06-19
Tag tests core

TesterSpike:
I've added some new test tags and added them to the tests.  If only the core tests are run then the test suite takes about 3 minutes to run.

I've also moved the tests and steps into packages of their own to make the separation more obvious.

When running the core tests some tests were failing because of changes made to the tier prices (test prices in line with production prices now) and a field that had a css selector updated because of the supporter tier changes.

### DIFF
--- a/functional-tests/src/main/scala/com/gu/membership/tags/Tags.scala
+++ b/functional-tests/src/main/scala/com/gu/membership/tags/Tags.scala
@@ -2,7 +2,23 @@ package com.gu.membership.tags
 
 import org.scalatest.Tag
 
+/* Usage: The tag is added as a comma separated list for each scenario after the scenario name
+e.g. scenarioWeb("Example test that is unstable but quick", Small, Unstable)
+  Running tests locally:  The Run/Debug configuration can specify which tags you want to run or ignore using the tag
+  text e.g. to run only core tests but ignore unstable or optional tests you would use the following in Test Options
+  -n CoreTest -l OptionalTest -l Unstable
+  Running tests TeamCity: In Edit Configuration Settings -> Build Step -> SBT Commands enter the following
+  "project functional-tests" clean compile "test-only -- -l unstable"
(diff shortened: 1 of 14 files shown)


functional-tests/src/main/scala/com/gu/membership/tags/Tags.scala
```diff
@@ -2,7 +2,23 @@ package com.gu.membership.tags
 
 import org.scalatest.Tag
 
+/* Usage: The tag is added as a comma separated list for each scenario after the scenario name
```
davidrapson: :tada: Thanks for adding these steps! Mind adding it into the `README` in `functional-tests/` as well?

davidrapson:
Ran `test-only -n CoreTest -l OptionalTest -l Unstable`. Getting a few failing tests, particularly around change tier.

Seems to be this selector failing on identity: `https://github.com/guardian/membership-frontend/blob/eb49e2d8d2f9ea09add0c451f90aa94da61bf5bd/functional-tests/src/main/scala/com/gu/membership/pages/IdentityEditPage.scala#L28`. 

As this is on the Membership tab, it's a little unstable. Might be worth us having a look to make sure this runs as expected.

Core ran quickly though so :clap:

functional-tests/src/main/scala/com/gu/membership/tags/Tags.scala
TesterSpike: will do :-)

TesterSpike:
I'll have a look at that selector and see if we can improve it in some way

davidrapson:
Success! I had something misconfigured with my profile setup which is why they were failing before. All passing now, ran in < 2 minutes using phantomjs for core tests which is great. :+1: from me.

screen shot 2015-06-24 at 14 10 23

davidrapson:
@TesterSpike Are you OK to merge and deploy this, or would you prefer I did?

davidrapson:
Going to merge this in this afternoon

PR #590, opened 2015-06-22
turn of async tracking

dominickendrick:
This stops async requests for membership event tracking. It seems to be causing file handle limit issues so this is to fix that in the short term.


jennysivapalan:
If it is short term then I'm okay with this. We'll need to keep an eye on cloudwatch metrics etc :+1:

davidrapson:
@jennysivapalan @dominickendrick Is this something we still need to do?

dominickendrick:
I think the issue was less problematic that I originally thought. So we can close this PR

PR #591, opened 2015-06-22
A/B test challenger template

davidrapson:
This PR adds the new A/B test challenger page behind Google auth.

sco mov

join the guardian membership

*Some slight weirdness in that screenshot due to the sticky header*

**Note**  This test is for desktop only so the this is not polished on small screens. It largely looks OK with existing patterns, just will need refinement if we go with this page.

@jennysivapalan @afiore


jennysivapalan:
I only looked at Scala and it looks good :+1: . Why is the endpoint behind Google auth and disallow on the robots.txt?

jennysivapalan:
The build was still running this morning so I just stopped it and ran a new build.

davidrapson:
@jennysivapalan It's behind google auth because I'd like to get the page out today but we're not going to run the test just yet (so just get it up for review). robots.txt is because this is a page for an A/B test and I guess we don't really want it indexed.

PR #597, opened 2015-06-24
Challenger test setup

davidrapson:
Some setup required for the new A/B test:

- Uses new optimizely account
- Sets up custom bucketing code (hide navigation items if user is in variation bucket)

@jennysivapalan


jennysivapalan:
:+1: :game_die:

PR #599, opened 2015-06-24
Challenger page refinements

davidrapson:
Challenger page refinements:

- Updated copy
- Some design / padding tweaks from Ben W
- Add smooth scrolling / animation to sections navigation
- Refactor sections navigation to avoid layout jump (there's a couple of magic number that will need refactoring, the CSS for this page in general is rougher than it could be; but OK for a test).
- **Make page public so we can validate optimizely setup**

smooth-scroll mov

:scroll: :scroll: :scroll: 

@jennysivapalan


jennysivapalan:
:scroll: :+1:

davidrapson:
Waiting…

screen shot 2015-06-25 at 10 33 29

PR #600, opened 2015-06-24
Remove old nav-popup file

davidrapson:
Removes an unused sass partial that was refactored out in https://github.com/guardian/membership-frontend/pull/545. File is not `@import`ed at all so is safe to remove.

@afiore


afiore:
:+1:

PR #601, opened 2015-06-25
Minor amends to challenger page

davidrapson:
- Adjust positioning of header text
- Correct a more link label & jump position
- Fix display issue with more link hovers

screen shot 2015-06-25 at 11 00 34

@afiore


afiore:
:+1:

PR #602, opened 2015-06-25
Setup test bucket variations

davidrapson:
This PR runs the variation code needed to hide navigation items if you are in a specific A/B test bucket.

@afiore

### DIFF
--- a/frontend/assets/javascripts/src/modules/analytics/optimizely.js
+++ b/frontend/assets/javascripts/src/modules/analytics/optimizely.js
@@ -1,6 +1,7 @@
 define(function() {
 
     var EXPERIMENT_ID = 3089970060;
+    var VARIATION_INDEX = 1;
 
     function variantSteps() {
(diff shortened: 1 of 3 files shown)


frontend/assets/javascripts/src/modules/analytics/optimizely.js
```diff
@@ -20,8 +16,8 @@ define(function() {
     }
 
     function init() {
-        var userVariations = window.optimizely.data.state.variationMap || false;
-        if (userVariations && userVariations[SETTINGS.experimentId] === SETTINGS.variationId) {
+        var userVariations = window.optimizely.variationMap || false;
+        if (userVariations && userVariations[EXPERIMENT_ID] === 1) {
```
afiore: Don't know much about the context (i.e. what a variation and a variation id is...), but I think it would read better if you were to replace the `1` with something like `VARIATION_ID`.

davidrapson:
@afiore Should be a bit clearer now

PR #604, opened 2015-06-25
Switch to px-to-rem

davidrapson:
This PR switches from `guss-rem` to a grunt task for converting px values to `rem` units. Means we no longer have to wrap everything in `rem()`. Removes a few instances of `rem()`, another PR will follow removing the rest if this goes well.

Frontend have already made this move and as `guss-rem` is unlikely to get much love in future it makes sense to move to this as well.

Tested a key pages across a few browsers and all looks good:

screen shot 2015-06-25 at 14 18 31

screen shot 2015-06-25 at 14 23 18

screen shot 2015-06-25 at 14 18 46

cc/ @sndrs In the absence of another client-side developer on this project, mind giving this the once over?

@TesterSpike It would be great if you're able to give this a run through as well. Expectation is that everything renders the same as PROD.

### DIFF
--- a/frontend/Gruntfile.js
+++ b/frontend/Gruntfile.js
@@ -50,10 +50,7 @@ module.exports = function (grunt) {
             options: {
                 outputStyle: 'compressed',
                 sourceMap: isDev,
-                precision: 5,
-                includePaths: [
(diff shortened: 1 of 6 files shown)


TesterSpike:
:+1:  everything looks the same in all browsers/mobile

PR #610, opened 2015-06-29
Don't show pricing for Shubbak

davidrapson:
Shubbak: A Window on Contemporary Arab Culture is a multi-event multi-day event with complex uses of ticket types so we have no way of knowing what price to report.

For this case the only way around this I can think of is to special-case the event and not show pricing for it. Yes it's unpleasant. This will be reverted on 30th July.

screen shot 2015-06-29 at 17 50 13
screen shot 2015-06-29 at 17 50 19
screen shot 2015-06-29 at 17 50 16

@jennysivapalan


jennysivapalan:
I don't want to plus 1 this so I'll just say :ok_hand:

PR #616, opened 2015-07-02
Fix SVG image scaling issues in IE

davidrapson:
> Internet Explorer 9–11 suffer from various bugs that prevent proper scaling of inline SVG’s.
> [http://nicolasgallagher.com/canvas-fix-svg-scaling-in-internet-explorer/]()

As spotted by @TesterSpike there are some scaling/aspect-ratio issues with SVGs as images in IE (all versions 9-11), thankfully there's an approach for fixing it. 

This PR models SVG images to encapsulate Nicolas Gallagher's SVG scaling fix. Only being used on the event listing headers for now as this is the most prominent place there's an issue, but will also be used on other provider logos in future (we're currently doing some manual massaging on provider logos so the issues aren't as noticeable there but this will still improve things).

screen shot 2015-07-02 at 10 54 05
screen shot 2015-07-02 at 10 53 39
screen shot 2015-07-02 at 11 00 18
screen shot 2015-07-02 at 11 00 01

*Masterclasses is particularly noticeable*

@TesterSpike Mind giving this the once over?

### DIFF
--- a/frontend/app/model/ResponsiveImage.scala
+++ b/frontend/app/model/ResponsiveImage.scala
@@ -1,6 +1,7 @@
 package model
 
 import com.gu.contentapi.client.model.Content
+import views.support.Asset
 import model.RichEvent.GridImage
 
 object ResponsiveImageGenerator {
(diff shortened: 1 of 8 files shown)


TesterSpike:
:+1:

PR #617, opened 2015-07-02
Prevent dual sentry logging

davidrapson:
Remove `Raven.captureMessage` for third-party JS as `Raven.captureException(err);` is enough. The former just gives an unnecessary line of logging without any context.


jennysivapalan:
:+1: :arrow_down:  :bird:

PR #623, opened 2015-07-14
Improve handling of long running events

davidrapson:
**Revised version of https://github.com/guardian/membership-frontend/pull/563 now we have  a live example of this** 

I've added a JSON mock of that event and added some test cases around it. Will compare against [this production](https://membership.theguardian.com/event/shubbak-a-window-on-contemporary-arab-culture-17514481285) with @TesterSpike and then we can release this.

---

Improves handling of multi-day events. Previously events were only sellable when the eventbrite status was `live`. This causes issues with pseudo multi-day events where an event may be classed as `started` but there are still some ticket types available to buy (due to them being used to model future dates).

Having spoken to @rtyley a suitable fix for this is to use the `sales_end` dates from the events `ticket_classes`. If a ticket type is still for sale then it's OK to show the booking button.

**Before:**

screen shot 2015-07-14 at 13 08 29

**After**

screen shot 2015-07-14 at 13 10 42

---

This also updates the sales end date to work with multi-day events. Previously this would show the sales end of the first ticket type.

screen shot 2015-06-02 at 15 48 37

@jennysivapalan @TesterSpike

### DIFF
--- a/frontend/app/model/Eventbrite.scala
+++ b/frontend/app/model/Eventbrite.scala
@@ -151,7 +151,7 @@ object Eventbrite {
 
     val salesDates = TicketSaleDates.datesFor(eventTimes, primaryTicket)
 
-    val salesEnd = primaryTicket.sales_end
+    val salesEnd = allTickets.map(_.sales_end).sorted.last
 
     val isCurrentlyAvailableToPaidMembersOnly =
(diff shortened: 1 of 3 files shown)


TesterSpike:
:+1:

davidrapson:
Confirmed on PROD

screen shot 2015-07-14 at 16 10 06

PR #624, opened 2015-07-16
Fix zxcvbn version

davidrapson:
zxcvbn, the password strength component we use has released a [number of broken versions](https://github.com/dropbox/zxcvbn/releases) which change the way it works with AMD. This PR locks down the version to an explicit hash for the version prior to the recent releases. Not ideal, but OK for now until I can update properly to the latest version.

See also https://github.com/guardian/subscriptions-frontend/pull/111

@afiore


davidrapson:
@afiore ping

davidrapson:
TeamCity backed up :disappointed:

PR #627, opened 2015-07-20
Simplify ticket sales dates. Remove toggle

davidrapson:
As we have recently simplified out priority booking rules, this also means we can simplify how we display ticket dates on event detail pages. This PR removes the toggle and shows a consistent label for all cases. Less code, clearer UX, winner.

screen shot 2015-07-20 at 16 54 33

**Before**

screen shot 2015-07-20 at 16 52 35

**After**

screen shot 2015-07-20 at 16 52 23

@afiore @tudorraul


afiore:
:+1:

PR #628, opened 2015-07-21
Use thumbnail image, not main

davidrapson:
We should be using the thumbnail crop for offers/masterclass images, not main. Solves this problem:

**Before**

screen shot 2015-07-21 at 11 50 57

**After**

screen shot 2015-07-21 at 11 53 57

Still looks correct on Masterclasses

screen shot 2015-07-21 at 11 53 13
screen shot 2015-07-21 at 11 53 18

@afiore @TesterSpike


TesterSpike:
:+1:

PR #629, opened 2015-07-21
Abstract default provider logo

davidrapson:
Abstract default provider logo so we model it rather than having conditional statements.

screen shot 2015-07-21 at 14 21 38
screen shot 2015-07-21 at 14 21 47

@afiore As you are now working on Membership, mind giving this a look? :smile:


afiore:
:+1:  Looks good. `frontend/app/model/Eventbrite.scala` is huge! perhaps it would be good to split it out into a proper package with multiple files, but that is outside of the scope of this PR.

davidrapson:
@afiore Yep it is :smile: , likewise for `RichEvent`. Going to have a little refactor soon.

PR #630, opened 2015-07-22
Fix event overview layout

davidrapson:
Before:

screen shot 2015-07-22 at 13 04 21

After:

screen shot 2015-07-22 at 13 09 10

@afiore @tudorraul


afiore:
:+1: Big tabs look much nicer!

PR #631, opened 2015-07-22
De-remify base and layout styles

davidrapson:
As we now [use a postprocessor](https://github.com/guardian/membership-frontend/blob/master/frontend/Gruntfile.js#L72) to handle converting all px values to rem, we can remove the direct use of `rem()` This PR cleans up base and layout styles. Another PR to follow which cleans up components.

@afiore @tudorraul


afiore:
:+1:

PR #632, opened 2015-07-22
Extract Metadata from RichEvent

davidrapson:
- Extract Metadata from RichEvent
- Move large image and priority booking settings *out* of Metadata
- Move social hashtag *in* to Metadata

@afiore You'll like this one :smile:


afiore:
:+1:

PR #633, opened 2015-07-22
Port password strength refinements from Subscriptions

davidrapson:
Port password strength refinements from Subscriptions

screen shot 2015-07-22 at 16 54 03


tudorraul:
:+1: LGTM

PR #635, opened 2015-07-23
Rationalise colour variables

davidrapson:
Rationalise colour variables, notably renames `mem-brandBlue` variables to be more generic and consistent with variable naming convention. Particularly helpful as the Membership brand colour is about to not be blue…

@afiore @tudorraul


afiore:
:+1:

PR #637, opened 2015-07-24
Implement Content Security Policy

davidrapson:
This PR implements a [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/Security/CSP)

Known third-party sources have been white-listed in the policy. We are running this in report only mode for now, which means that no content will be blocked but any violations will be reported to and endpoint and logged to Sentry. 

The reporting should help us audit for any third-parties we don't know about as well as reporting any malicious third-party scripts. Testing:

- [x] Check functional tests pass
- [x] Check membership tab on Profile

@rtyley @afiore


rtyley:
:+1:

davidrapson:
@TesterSpike Would you mind giving this a once over?

TesterSpike:
:+1:

PR #638, opened 2015-07-27
Only show Programming Partner Events if we have some

davidrapson:
Only show Programming Partner Events if we have some

screen shot 2015-07-27 at 11 45 35
screen shot 2015-07-27 at 12 08 07

@afiore


afiore:
:+1:

PR #639, opened 2015-07-27
Revert Shubbak special-case code

davidrapson:
Reverts special-case code from https://github.com/guardian/membership-frontend/pull/610. Ended on 26th, not 29th so we can revert this a little early.

@afiore @ostapneko


afiore:
:+1: good to see the spacial case is going

PR #640, opened 2015-07-28
Priority booking: No longer need start of day logic

davidrapson:
General release for Alex Ferguson event was set to midnight, instead of specified time of 9am. This was due to some holdover logic from the previous priority booking rules. We no longer need to adjust the times to the start of the day in any case.

@rtyley


rtyley:
:+1: tho' we should consequently start displaying time on the ticket release times.

PR #636, opened 2015-07-23
Guardian Members Redesign

davidrapson:
This is now in a position to review. I've pulled out all non-visual changes I can, but the PR is still quite large. Apologies.

Some cross-browser notes: The angled-cuts across the site should be supported in Chrome, Safari and Opera (both desktop and covering iOS/Android). On Firefox/IE it is expected that these fallback to rectangular shapes.

# Screenshots

homepage
masterclasses
events
event-detail-small
brochure-page
event-details

#  QA

Just adding notes from Sean here with clarifications and/or fixes where appropriate:

- [x] ~~Are the competition pages not meant to be rebranded/redesigned?~~ Referring to pages on theguardian.com, this is OK.
- [x] ~~Should the home page have consistent treatment of the images?  Large image has angle effect but the smaller images do not~~ Intentional
- [x] Should the image on book event/masterclass page have the same treatment as the image else where?
- [x] **IE9** The 'Member' text in the logo partially falls off the edge of the banner and is not visible
- [x] ~~**IE9** The 'G Live' logo at the bottom of the page is stretched~~ This a known issue with the provider logos (also applies to third-party logos); can be fixed using the same technique that fixes the logo stretching in the header but I'll fix this in a later PR as it's an issue with the current design too.
- [x] **IE9** Optional text on Feedback page is a little close to the other elements
- [x] ~~**Safari Windows 8.1** About page White button styling inconsistent with coloured button styling~~ Safari on Windows was deprecated a number of years ago (locked to 5.1, current version 9) so a couple of visual bugs are fine.

# Release

- [ ] Awaiting final approved logo before this can go live. 

@afiore @tudorraul @rtyley

PR #641, opened 2015-07-28
[Redesign prep] Simplify Masterclasses event listings logic

davidrapson:
This has come out of the work on the rebrand https://github.com/guardian/membership-frontend/pull/636. Simplifies how Masterclass event listings are handled, no need to use an empty `EventPortfolio` here. The goal is to use the new fragment to simplify the main events page too.

@afiore @rtyley


afiore:
:+1:

PR #642, opened 2015-07-28
[Redesign prep] We don't need Modernizr anymore

davidrapson:
This is an effort to remove any non-visual changes from https://github.com/guardian/membership-frontend/pull/636. Turns out we don't actually need Modernizr any more, so we can remove it.

@rtyley @afiore


afiore:
:+1: One less dependency!

PR #643, opened 2015-07-28
[Redesign prep] Fallback to events list instead of front

davidrapson:

It makes more sense, and is less frustrating in dev, to make the event detail fallback link go to the events list instead of the Membership front.

@rtyley @afiore


afiore:
:+1: For internal links I think is generally always better to use the routes module and avoid code duplication.

PR #645, opened 2015-07-28
[Redesign prep] Extract SVGImage out from ResponsiveImage model

davidrapson:

This extracts SVGImage out from ResponsiveImage model, as this model is likely to get larger as part of the redesign; and it was a bit lazy of me to dump it here in the first place.

@rtyley @afiore

### DIFF
--- a/frontend/app/model/SVG.scala
+++ b/frontend/app/model/SVG.scala
@@ -1,0 +1,16 @@
+package model
+
+import views.support.Asset
+
+object SVG {
+
+  case class SVGImage(
+    label: String,
+    path: String,
+    width: Int,
+    height: Int
+  ) {
+    val src = Asset.at(path)
+  }
+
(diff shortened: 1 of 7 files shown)


rtyley:
It was just occupying the same .scala file, which doesn't bother me too much, but :+1: if it makes things happier for you :sparkles:

davidrapson:
@rtyley Yeah, this model is going to expand a bit, so it makes things a bit clearer for me.

frontend/app/model/SVG.scala
```diff
@@ -0,0 +1,16 @@
+package model
+
+import views.support.Asset
+
+object SVG {
+
+  case class SVGImage(
```
afiore: Minor naming tip: there's no need to repeat the `SVG` prefix here

PR #646, opened 2015-07-28
[Redesign prep] Use sass-map for brand colours

davidrapson:

As we will soon have a larger set of Membership brand colours, it's helpful to model them in a similar way to frontend using sass maps. This PR maintains the existing brand colours but introduces the map config ahead of the redesign colour palette changes.

@afiore


afiore:
:+1:

PR #647, opened 2015-07-28
Allow partners to select free books or events as a membership perk

afiore:
Benefit selection

- Allows new partners to chose between books or events
- Automatically setup patrons with both perks

This implementation relies on `model.FeatureChoice.id` in order for the code to synchronise with Zuora product features. We are disabling this functionality on production, but it will still be accessible in development mode.

*Coding todo*
- [  ] ~~Add appropriate copy for form dialogs~~
- [x] Produce an error when the expected product features cannot be found in Zuora 

**Config todo**  
- [x] Add `Events` and `Books` product features to Zuora Dev
- [x] Make sure that the *patron* and *partner* product have the allow *Allow Feature Changes in Subscriptions*  option set to `yes` 
- [x] repeat steps above for UAT
- [x] repeat steps above for PROD

@rtyley @davidrapson @TesterSpike

### DIFF
--- a/frontend/app/configuration/Config.scala
+++ b/frontend/app/configuration/Config.scala
@@ -164,6 +164,4 @@ object Config {
   val bcryptPepper = config.getString("activity.tracking.bcrypt.pepper")
 
   val casServiceConfig = config.getString("cas.url")
-
-  val tierFeatures = config.getStringList("tierFeatures").toList
 }
--- a/frontend/app/configuration/Config.scala
+++ b/frontend/app/configuration/Config.scala
@@ -5,7 +5,6 @@ import com.amazonaws.auth.{AWSCredentialsProviderChain, InstanceProfileCredentia
 import com.gu.googleauth.{GoogleAuthConfig, GoogleServiceAccount}
 import com.gu.identity.cookie.{PreProductionKeys, ProductionKeys}
 import com.gu.membership.salesforce.Tier
-import com.gu.membership.touchpoint.TouchpointBackendConfig
 import com.netaporter.uri.dsl._
 import com.typesafe.config.ConfigFactory
 import model.Eventbrite.EBEvent
@@ -13,8 +12,8 @@ import net.kencochrane.raven.dsn.Dsn
 import play.api.Logger
 import services._
 
+import scala.collection.JavaConversions._
 import scala.util.Try
-import java.net.URLEncoder
 
 object Config {
   val logger = Logger(this.getClass())
@@ -166,4 +165,5 @@ object Config {
 
   val casServiceConfig = config.getString("cas.url")
 
+  val tierFeatures = config.getStringList("tierFeatures").toList
 }
--- a/frontend/app/controllers/Healthcheck.scala
+++ b/frontend/app/controllers/Healthcheck.scala
@@ -1,39 +1,33 @@
 package controllers
 
+import com.github.nscala_time.time.Imports._
+import com.gu.monitoring.CloudWatchHealth
 import play.api.Logger
 import play.api.mvc.{Action, Controller}
-import services.{TouchpointBackend, GuardianLiveEventService}
-import com.gu.monitoring.CloudWatchHealth
-import com.github.nscala_time.time.Imports._
-import play.api.libs.concurrent.Execution.Implicits._
+import services.{GuardianLiveEventService, TouchpointBackend}
 
-import scala.concurrent.Future
-
-case class Test(name: String, result: () => Future[Boolean])
+case class Test(name: String, ok: () => Boolean)
 
 object Healthcheck extends Controller {
   val zuoraService = TouchpointBackend.Normal.zuoraService
 
   val tests = Seq(
-    Test("Events", () => Future { GuardianLiveEventService.events.nonEmpty }),
-    Test("CloudWatch", () => Future { CloudWatchHealth.hasPushedMetricSuccessfully }),
-    Test("Zuora", () => zuoraService.pingSupplier.get().map(t => t > DateTime.now - 2.minutes))
+    Test("Events", () => GuardianLiveEventService.events.nonEmpty),
+    Test("CloudWatch", () => CloudWatchHealth.hasPushedMetricSuccessfully),
+    Test("Zuora", () => zuoraService.lastPingTimeWithin(2.minutes))
   )
 
-  def healthcheck() = Action.async { req =>
-    Future.sequence(tests.map(_.result())).map { results =>
-      val failedTests =
-        results.zip(tests.map(_.name)).filterNot { case (ok, _) => ok }
+  def healthcheck() = Action { req =>
+    val failedTests = tests.filterNot(_.ok())
 
-      Cached(1) {
-        if (failedTests.nonEmpty) {
-          failedTests.foreach { case (_, test) =>
-            Logger.warn(s"Test $test failed, health check will fail")
-          }
-          ServiceUnavailable("Service Unavailable")
-        } else {
-          Ok("OK")
+    Cached(1) {
+      if (failedTests.nonEmpty) {
+        failedTests.foreach { test =>
+          Logger.warn(s"Test ${test.name} failed, health check will fail")
         }
+        ServiceUnavailable("Service Unavailable")
+      } else {
+        Ok("OK")
       }
     }
   }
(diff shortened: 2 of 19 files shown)


frontend/app/forms/MemberForm.scala
```diff
@@ -47,6 +53,25 @@ object MemberForm {
 
   case class FeedbackForm(category: String, page: String, feedback: String, name: String, email: String)
 
+
+  implicit val productFeaturesFormatter: Formatter[FeatureChoice] = new Formatter[FeatureChoice] {
```
tudorraul: type hint seems redundant

frontend/app/model/FeatureChoice.scala
```diff
@@ -0,0 +1,38 @@
...
+  override val codes = Seq("Books")
+}
+case object Events extends FeatureChoice {
+  override val id    = "Events"
+  override val label = "6 free events"
+  override val codes = Seq("Events")
+}
+
+case object BooksAndEvents extends FeatureChoice {
+  override val id = "Both"
```
tudorraul: Using `BooksAndEvents` as the `id` might be more intuitive here

frontend/app/forms/MemberForm.scala
```diff
@@ -109,7 +135,7 @@ object MemberForm {
     )(StaffJoinForm.apply)(StaffJoinForm.unapply)
   )
 
-  val paidMemberJoinForm: Form[PaidMemberJoinForm] = Form(
+  def paidMemberJoinForm: Form[PaidMemberJoinForm] = Form(
```
tudorraul: any benefit of using `def`?

frontend/app/model/FeatureChoice.scala
afiore: Well spotted!

frontend/app/forms/MemberForm.scala
afiore: Unfortunately it isn't. If I do not set it, when I inspect the type in Intellij I end up with a one meter long string. Blame it on the compiler...

frontend/app/forms/MemberForm.scala
afiore: Well spotted once again! This was a left over from some changes I eventually haven't committed!

frontend/conf/dev.conf
```diff
@@ -18,7 +18,7 @@ include "PROD"
 
 identity.webapp.url="https://profile.thegulocal.com"
 identity.production.keys=false
-identity.api.url="https://idapi.thegulocal.com"
+identity.api.url="https://idapi.code.dev-theguardian.com"
```
tudorraul: was this meant to be committed?

afiore: I think that hitting identity code a better default setting

TesterSpike:
:+1:  tested using zuora uat

frontend/conf/dev.conf
rtyley: Although we originally developed Membership with everything local, including Identity API, we're now recommending that the Identity version can be the one hosted on CODE: https://docs.google.com/a/guardian.co.uk/spreadsheets/d/1EBU0VrRZCslq_A_mKlocxOhAtCGXI3Ye-KJlKVCIVi0/edit?usp=sharing

frontend/app/configuration/Config.scala
```diff
@@ -5,7 +5,6 @@ import com.amazonaws.auth.{AWSCredentialsProviderChain, InstanceProfileCredentia
 import com.gu.googleauth.{GoogleAuthConfig, GoogleServiceAccount}
 import com.gu.identity.cookie.{PreProductionKeys, ProductionKeys}
 import com.gu.membership.salesforce.Tier
-import com.gu.membership.touchpoint.TouchpointBackendConfig
```
rtyley: This file only seems to have import changes! This PR is pretty big - try to exclude unnecessary changes if possible.

frontend/app/controllers/Healthcheck.scala
```diff
@@ -5,27 +5,36 @@ import play.api.mvc.{Action, Controller}
 import services.{TouchpointBackend, GuardianLiveEventService}
 import com.gu.monitoring.CloudWatchHealth
 import com.github.nscala_time.time.Imports._
+import play.api.libs.concurrent.Execution.Implicits._
 
-case class Test(name: String, result: () => Boolean)
+import scala.concurrent.Future
+
+case class Test(name: String, result: Future[Boolean])
```
rtyley: There's been a big change in semantics here - the tests are now only ever evaluated once!  :-1:

frontend/app/forms/MemberForm.scala
```diff
@@ -19,22 +21,26 @@ object MemberForm {
...
+    override val plan = FriendTierPlan
+    override val featureChoice = None
   }
 
   case class StaffJoinForm(name: NameForm, deliveryAddress: Address, marketingChoices: MarketingChoicesForm,
-                            password: Option[String] ) extends JoinForm {
-    val plan = StaffPlan
+                            password: Option[String]) extends JoinForm {
+    override val plan = StaffPlan
+    override val featureChoice = None
```
rtyley: Have you clarified if Staff members will not be getting the free tickets?

frontend/app/model/FeatureChoice.scala
```diff
@@ -0,0 +1,42 @@
...
+  override val label = "4 free books"
+  override val codes = Set(id)
+}
+case object Events extends FeatureChoice {
+  override val id    = "Events"
+  override val label = "6 free events"
+  override val codes = Set(id)
+}
+
+case object BooksAndEvents extends FeatureChoice {
```
rtyley: Is it necessary to create a special BooksAndEvents feature?! Surely this should be represented as a set that can contain either `Books` or `Events` or both?

afiore:
@rtyley I have updated the PR and refactored FeatureChoice to behave like a set. Additionally, I have also simplified the Healthcheck code so that it doesn't use futures

frontend/app/model/FeatureChoice.scala
```diff
@@ -0,0 +1,32 @@
...
+    Books
+  )
+
+  val codes = all.map(_.id)
+
+  def fromString(id: String): Set[FeatureChoice] = id match {
+    case Books.id => Set(Books)
+    case Events.id => Set(Events)
+    case _ =>
+      if (id == Events.id + Books.id || id == Books.id + Events.id) Set(Books, Events)
```
rtyley: I think this parsing code would probably be better written to just split a comma-separated string, then individually map each element to a `FeatureChoice` in a Set

afiore:
@rtyley renamed `model.Events` to `model.FreeEventsTickets`. I have also pushed a few more tweaks to the parsing logic. As for the staff members, I believe they will have no access to the events/books product features (find the implementation details [here](https://github.com/guardian/membership-frontend/blob/books-or-events-choice/frontend/app/services/SubscriptionService.scala#L88))

frontend/app/forms/MemberForm.scala
```diff
@@ -119,7 +138,8 @@ object MemberForm {
       "marketingChoices" -> marketingChoicesMapping,
       "password" -> optional(nonEmptyText),
       "casId" -> optional(nonEmptyText),
-      "subscriberOffer" -> default(boolean, false)
+      "subscriberOffer" -> default(boolean, false),
+      "featureChoice" -> productFeature
```
rtyley: An interesting thing that's going on here is that I think that there's no server-side validation that a legitimate productFeature selection has been made. Specifically, a user signing up as a 'Partner' could manipulate the form so that it posted _both_ product features (Books *and* Free event tickets) - and they would get them, despite not signing up as a Patron.

rtyley:
:+1: I think this PR is good to merge (I know it's lived a long time!) - but it would be good to address the server-side validation concern in a subsequent PR:

https://github.com/guardian/membership-frontend/pull/647/files#r36313333

...probably enough to validate against a Map[Tier, Int] that specifies how many Product Features a user is allowed to select at a given tier.

afiore:
@rtyley We are actually doing this check, but not in the form class (see

frontend/app/forms/MemberForm.scala
afiore: I believe that tampering with the form would not be that easy, as we [keep the member tier into account](https://github.com/guardian/membership-frontend/blob/books-or-events-choice/frontend/app/services/SubscriptionService.scala#L88) before assigning product features to   a Zuora subscription. Is just that we do not perform that check as part of the form validation but directly within the `SubscriptionService`

PR #650, opened 2015-07-29
Update external actions to use inline svg icons

davidrapson:
Update external actions to use the preferred new inline SVG icons instead of background images. No visual changes; just cleanup.

screen shot 2015-07-29 at 14 26 31

@afiore


afiore:
:+1:

PR #652, opened 2015-07-29
Use strict for all JS modules; generally tighten up validation

davidrapson:
JS `strict` mode is generally worth defaulting to. We we're doing this in some modules, but as it wasn't enforced a lot were missing it. This PR adds `'use strict';` where needed; and generally tightens up our `eslint` config where possible.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -15,17 +15,12 @@
 
   "rules": {
     "camelcase": 0,
-    "strict": 0,
     "curly": [2, "all"],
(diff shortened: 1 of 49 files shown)


afiore:
Is `eslint` now going to complain if we have no 'use strict;' in a module?

davidrapson:
Yep, `eslint` will now enforce this for all new modules.

afiore:
:+1:

PR #653, opened 2015-07-30
Refactor zuora service implementation

afiore:
The current implementation of `ZuoraService` initialises a number of `ScheduledTasks` with a "blank" authentication value. This is not ideal, as it means that for a small time window, the application is not capable to talk to Zuora. In this specific case `ScheduledTask` also makes composing async operations in the right order harder than necessary, forcing us to use fiddly callbacks to ensure that operations happen in the right order.

Following a suggestion from @rtyley, I have repurposed a pattern from [Prout](github.com/guardian/prout). This consists in wrapping a future value into an Akka agent, which in turn gets periodically refreshed by the Actor system's scheduler (see `utils.FutureSupplier`).

**Note**: This PR branches off #647

@tudorraul @rtyley

### DIFF
--- a/frontend/app/services/zuora/ZuoraService.scala
+++ b/frontend/app/services/zuora/ZuoraService.scala
@@ -71,7 +71,7 @@ class ZuoraService(val apiConfig: ZuoraApiConfig) extends LazyLogging {
 
   def getAuth = authSupplier.get()
 
-  def authRequest[T <: ZuoraResult](action: => ZuoraAuthAction[T])(implicit reader: ZuoraReader[T]): Future[T] = {
+  def authRequest[T <: ZuoraResult](action: => ZuoraAction[T])(implicit reader: ZuoraReader[T]): Future[T] = {
     getAuth.flatMap(auth => request(action, Some(auth)))
   }
 
@@ -94,7 +94,9 @@ class ZuoraService(val apiConfig: ZuoraApiConfig) extends LazyLogging {
     val url = apiConfig.url.toString()
 
     Timing.record(metrics, action.getClass.getSimpleName) {
-      WS.url(url.toString).post(action.xml(authOpt))
+      WS.url(url.toString)
+        .withHeaders("ContentType" -> "text/plain")
+        .post(action.xml(authOpt))
     }.map { result =>
       metrics.putResponseCode(result.status, "POST")
       reader.read(result.body) match {
--- a/frontend/app/services/zuora/ZuoraService.scala
+++ b/frontend/app/services/zuora/ZuoraService.scala
@@ -20,6 +20,7 @@ import scala.concurrent.duration._
 
 import play.api.libs.concurrent.Execution.Implicits._
 import scala.concurrent.Future
+import scala.util.{Failure, Success}
 
 case class ZuoraServiceError(s: String) extends Throwable {
   override def getMessage: String = s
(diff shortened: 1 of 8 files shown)


frontend/app/utils/FutureSupplier.scala
```diff
@@ -0,0 +1,19 @@
...
+import play.api.libs.concurrent.Execution.Implicits._
+
+class FutureSupplier[T](f: => Future[T]) {
+  private val agent = Agent(f)
+
+  def refresh(): Future[T] = for {
+    refreshAlteringFuture <- agent.alter { currentRefresh =>
+      if (currentRefresh.isCompleted) f else currentRefresh
+    }
+    refreshFuture <- refreshAlteringFuture
```
tudorraul: misleading name here I think; the type of `refreshFuture` is not `Future[T]`

frontend/app/utils/FutureSupplier.scala
```diff
@@ -0,0 +1,19 @@
...
+
+import scala.concurrent.Future
+import play.api.libs.concurrent.Execution.Implicits._
+
+class FutureSupplier[T](f: => Future[T]) {
+  private val agent = Agent(f)
+
+  def refresh(): Future[T] = for {
+    refreshAlteringFuture <- agent.alter { currentRefresh =>
+      if (currentRefresh.isCompleted) f else currentRefresh
```
tudorraul: can we avoid reassigning the same value back to the agent, as it might cause an additional blocking of the agent while this happens, which is slightly suboptiomal?

frontend/app/services/zuora/ZuoraService.scala
```diff
@@ -36,78 +38,72 @@ object ZuoraServiceHelpers {
...
-  def authTaskWithCallbacks(callbacks: Seq[() => Unit]) =
-    ScheduledTask(s"Zuora ${apiConfig.envName} auth", Authentication("", ""), 0.seconds, 30.minutes) {
-      val eventualAuthentication = request(Login(apiConfig))
-      eventualAuthentication.filter(_.token.nonEmpty).foreach(_ => callbacks.foreach(_()))
-      eventualAuthentication
-    }
+  def getAuth = authSupplier.get()
 
-  val authTask = authTaskWithCallbacks(Seq(() => featuresSchedule))
+  def authRequest[T <: ZuoraResult](action: => ZuoraAction[T])(implicit reader: ZuoraReader[T]): Future[T] =
```
tudorraul: The name `authRequest` read authenticat**ion** request at first, but than I figured out it was authenticat**ed** request. I would use something that suggests this better this better, ie `authenticatedRequest` or `secureRequest`

frontend/app/utils/FutureSupplier.scala
afiore: Not sure that would bring a performance improvement (i.e. two transactions instead of one). And even if so, my guess is that it would be pretty negligible. But yeah, you could write something like:

```scala
currentRefresh <- agent.get
if (currentRefresh.isCompleted) {
  agent.send(_ => f)
}
currentFuture
```

Though I find it more concise and robust as it is to be honest

tudorraul: I was thinking something along the lines of
```scala
val currentRefresh = agent.get()
if(currentRefresh.isCompleted) currentRefresh
else agent.alter(_ => f).flatMap(a => a)
```

tudorraul:
:+1: bar the comment

PR #655, opened 2015-07-31
Fix Local events not displaying in past events

davidrapson:
This PR fixes a bug with Local events not displaying in past events. It looked like they were being included, but they actually weren't.

- Fix issue with Local events not displaying in past events container
- Increase the number of past events we show
- Don't use EventPortfolio for patterns; overcomplicated, we only need a sample list of events

**Before**
screen shot 2015-07-31 at 13 57 20

**After**
screen shot 2015-07-31 at 13 56 59

@rtyley

### DIFF
--- a/frontend/app/controllers/Event.scala
+++ b/frontend/app/controllers/Event.scala
@@ -143,21 +143,19 @@ trait Event extends Controller with ActivityTracking {
   }
 
   def list = CachedAction { implicit request =>
-    val pageInfo = PageInfo(
-      CopyConfig.copyTitleEvents,
-      request.path,
-      Some(CopyConfig.copyDescriptionEvents)
-    )
-    val pastEvents = (guLiveEvents.getEventsArchive ++ localEvents.getEventsArchive).headOption
-      .map(chronologicalSort(_).reverse)
     Ok(views.html.event.guardianLive(
       EventPortfolio(
         guLiveEvents.getFeaturedEvents,
         chronologicalSort(guLiveEvents.getEvents ++ localEvents.getEvents),
-        pastEvents,
+        chronologicalSort(guLiveEvents.getEventsArchive.toList.flatten ++ localEvents.getEventsArchive.toList.flatten).reverse,
         guLiveEvents.getPartnerEvents
       ),
-      pageInfo))
+      PageInfo(
+        CopyConfig.copyTitleEvents,
+        request.path,
+        Some(CopyConfig.copyDescriptionEvents)
+      ))
+    )
   }
 
   def listFilteredBy(urlTagText: String) = CachedAction { implicit request =>
@@ -171,7 +169,7 @@ trait Event extends Controller with ActivityTracking {
       EventPortfolio(
         Seq.empty,
         chronologicalSort(guLiveEvents.getTaggedEvents(tag) ++ localEvents.getTaggedEvents(tag)),
-        None,
+        Seq.empty,
         None
       ),
       pageInfo))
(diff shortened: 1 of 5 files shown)


frontend/app/views/fragments/event/listing.scala.html
```diff
@@ -68,34 +68,30 @@ <h3 class="listing__title h-intro">
         </section>
     }
 
-    @if(showPastEvents && eventPortfolio.pastEvents) {
```
rtyley: quite surprised that used to compile... `eventPortfolio.pastEvents` is an Option behaving like a boolean? Wonder if play templates is getting involved there.

davidrapson: Yeah, it surprised me a little too. These changes look OK though?

frontend/app/controllers/Event.scala
```diff
@@ -143,21 +143,19 @@ trait Event extends Controller with ActivityTracking {
...
-      Some(CopyConfig.copyDescriptionEvents)
-    )
-    val pastEvents = (guLiveEvents.getEventsArchive ++ localEvents.getEventsArchive).headOption
-      .map(chronologicalSort(_).reverse)
     Ok(views.html.event.guardianLive(
       EventPortfolio(
         guLiveEvents.getFeaturedEvents,
         chronologicalSort(guLiveEvents.getEvents ++ localEvents.getEvents),
-        pastEvents,
+        chronologicalSort(guLiveEvents.getEventsArchive.toList.flatten ++ localEvents.getEventsArchive.toList.flatten).reverse,
```
afiore: I think it would be worth creating an intermediate value here for readability purposes.
```scala
val archivedEvents = 
  (guLiveEvents.getEventsArchive ++ localEvents.getEventsArchive).flatten.toList.reverse 
```
Notice that the same expression can be written in a slightly more concise way by concatenating two options into an iterable.

afiore: Forget about my comment on `.reverse`, I hadn't noticed that is called on the output of `chronologicialSort`

frontend/app/views/fragments/event/listing.scala.html
```diff
@@ -68,34 +68,30 @@ <h3 class="listing__title h-intro">
...
-                            </li>
-                        }
-                    </ul>
-                    <button class="action action--secondary u-no-margin js-toggle" data-toggle="js-toggle-past-events" data-toggle-label="Less" data-toggle-icon="minus">
-                        @fragments.actionIcon("plus", leftIcon=true)
-                        <span class="action__label js-toggle-label">More past events</span>
-                    </button>
-                }
+                <ul class="grid grid--bordered grid--4up">
+                    @for(event <- eventPortfolio.pastEvents.slice(0, 4)) {
```
afiore: Why not `pastEvents.take(4)`?

davidrapson: Good call, this was already like this (just changing the check above / indentation) but this is nicer.

frontend/app/views/fragments/event/listing.scala.html
```diff
@@ -68,34 +68,30 @@ <h3 class="listing__title h-intro">
...
+                        <li class="grid__item">
+                            @fragments.event.itemMinimal(event, isCard=false)
+                        </li>
+                    }
+                </ul>
+                <ul class="grid grid--bordered grid--2up" id="js-toggle-past-events" data-toggle-hidden>
+                    @for(event <- eventPortfolio.pastEvents.slice(4, 36)) {
```
afiore: Perhaps replacing these offsets with constant names would make this code a bit more intuitive. Something like:

```scala
eventPortfolio.pastEvents.slice(featuredEventsCount, whatsOnCount)
```

afiore:
:+1:

PR #657, opened 2015-07-31
Unset CSP header, logging too noisy in Sentry

davidrapson:
CSP logging is currently very noisy in Sentry, messages seem to get truncated so we're missing context and where there is context there's a lot of noise from browser extensions (e.g., Safari Pinterest extension). Going to review this next week to see if there's a clearer way we can report this without overwhelming Sentry with more noise than signal. For now, let's unset the header.

@rtyley


tudorraul:
can we not turn notifications off in Sentry for this type of errors until we start the investigation?

davidrapson:
@tudorraul I think it's preferable to just remove it for now; we're not enforcing the CSP header so I'd rather just nuke the noise particularly over the weekend.

davidrapson:
@tudorraul Yay or nay?

tudorraul:
:+1: go for it

PR #659, opened 2015-08-03
Fix toggle icons

davidrapson:
I removed this by mistake :(

@afiore @tudorraul

### DIFF
--- a/frontend/assets/javascripts/src/modules/toggle.js
+++ b/frontend/assets/javascripts/src/modules/toggle.js
@@ -20,6 +20,7 @@ define(['$', 'bean'], function ($, bean) {
         TOGGLE_LABEL_SELECTOR = '.js-toggle-label',
         TOGGLE_DATA_ELM = 'toggle',
         TOGGLE_DATA_LABEL = 'toggle-label',
+        TOGGLE_DATA_ICON = 'toggle-icon',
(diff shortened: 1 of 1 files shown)


frontend/assets/javascripts/src/modules/toggle.js
```diff
@@ -33,6 +34,15 @@ define(['$', 'bean'], function ($, bean) {
         }
     };
 
+    var toggleIcon = function($elem) {
+        var iconName = $elem.data(TOGGLE_DATA_ICON);
+        var iconRef = $elem[0].querySelector('svg > use');
+        if (iconName && iconRef) {
+            $elem.data(TOGGLE_DATA_ICON, iconRef.getAttribute('xlink:href').replace('#icon-', ''));
```
afiore: Maybe it would be worth moving the '#icon' into a constant

davidrapson:
@afiore OK otherwise?

frontend/assets/javascripts/src/modules/toggle.js
davidrapson: @afiore OK otherwise?

afiore: :+1: ok otherwise

PR #660, opened 2015-08-04
Re-model CSP message to send formatted version

davidrapson:
Messages were getting truncated in Sentry, this formats the CSP report into something more readable, and removes `original-policy` from the reported message.

@rtyley

### DIFF
--- a/frontend/app/controllers/Security.scala
+++ b/frontend/app/controllers/Security.scala
@@ -39,6 +39,8 @@ object Security extends Controller with LazyLogging {
       |Violated Directive: $violatedDirective
       |Referrer: $referrer
       |""".stripMargin
+
+    val directiveTag = violatedDirective.split(" ").head
   }
 
   implicit val reads: Reads[CSPReport] = (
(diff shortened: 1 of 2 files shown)


frontend/app/controllers/Security.scala
```diff
@@ -2,24 +2,70 @@ package controllers
...
 
     SentryLogging.ravenOpt.fold {
-      logger.error(message)
+      logger.error(report.message)
     } {
-      _.sendMessage(message)
+      _.sendEvent(new EventBuilder()
+        .withMessage(report.message)
+        .withLevel(RavenEvent.Level.INFO)
+        .withTag("CSP Directive", report.directiveTag)
```
davidrapson: @rtyley Added a tag for the violated directive type e.g., `img-src`, `script-src` which feels like the most appropriate thing to tag on.

rtyley:
@davidrapson I added rate-limiting with 068a056, apart from that I think this PR is good to go :+1:

davidrapson:
@rtyley Thanks for adding that!

PR #661, opened 2015-08-05
Update Optimizely experiment ID

davidrapson:
In order to push the winning Optimizely variant to 100% of users we need to restart the experiment (to reset old visitors that might already be bucketed). This means we need to update the experiment ID in our code.

@afiore @tudorraul


afiore:
:+1:

PR #664, opened 2015-08-06
USA Supporter page

davidrapson:
This PR adds the USA Supporter page.

- USD prices are shown approximately and rounded up. Stakeholders understand that these are hard-coded and approximate prices.
- Promos now have the option to display an image caption (used for this page)
- URL for the page will be `/us/supporter`. Note, the current UK page is `/about/supporter`. The reason we are not using `/us/about/supporter` is **a)** the About page is about to go (replaced by long homepage) and **b)** there is no equivalent `/us/about` page.

**Wide**

wide

**Narrow**

narrow

@afiore @tudorraul @TesterSpike


frontend/app/views/fragments/promos/promoPrimary.scala.html
```diff
@@ -12,16 +13,26 @@
     <div class="promo-primary__header @toneClass.mkString(" ")">
         <h2 class="promo-primary__title h-promo@if(isLead){ h-promo--primary}">@title</h2>
     </div>
-    <div class="promo-primary__media@if(stampImageOpt){ stamped-image}">
+    <figure class="promo-primary__media@if(stampImageOpt){ stamped-image}@if(showImageCaption) { captioned-image}">
```
tudorraul: `if` on an `Option` looks a bit unsafe. Would `map` work just as well?

frontend/app/views/fragments/promos/promoPrimary.scala.html
```diff
@@ -12,16 +13,26 @@
...
         <div class="promo-primary__image">
-            <img src="@img.defaultImage" srcset="@img.srcset" sizes="100vw" alt="@img.altText" class="responsive-img" />
+            <img src="@img.defaultImage" srcset="@img.srcset" sizes="100vw" alt="@img.metadata.map(_.description).orElse(img.altText).getOrElse("")" class="responsive-img" />
```
tudorraul: I would extract this in a method

frontend/app/views/fragments/promos/promoSecondary.scala.html
```diff
@@ -12,13 +13,23 @@
         <div class="promo-secondary__header">
             <h2 class="promo-secondary__title h-promo">@title</h2>
         </div>
-        <div class="promo-secondary__image@if(stampImageOpt){ stamped-image}">
-            <img src="@img.defaultImage" srcset="@img.srcset" sizes="100vw" alt="@img.altText" class="responsive-img" />
+        <div class="promo-secondary__image@if(stampImageOpt){ stamped-image}@if(showImageCaption) { captioned-image}">
```
tudorraul: this seams to repeat itself. Maybe a method would make it more readable too.

frontend/app/views/info/supporterUSA.scala.html
```diff
@@ -0,0 +1,82 @@
...
+                </div>
+                @fragments.tier.packagePromoUSASupporter(isReversed=false)
+            }
+        </div>
+    }
+
+    @* ===== Support open, independent journalism ===== *@
+    <div id="supporter-video" class="page-slice page-slice--slim l-constrained">
+        @fragments.promos.promoVideoSecondary(
+            Html("Support open, independent journalism"),
```
tudorraul: can the use of `Html` be removed here? all clients of this template use only a `String`.

tudorraul:
:+1: bar the comments

PR #667, opened 2015-08-10
Record free event ticket usage in Zuora

afiore:
After a quick investigation that ruled out the possibility of using EventBrite webhooks, @rtyley and I agreed on recording event usage within the thank you page the user lands on after buying an event.
Assuming that the user is signed into the app, we can link the newly created usage record in Zuora with the member `salesforceContactId` and its corresponding subscription id.

screen shot 2015-08-10 at 18 04 37

**Rollout plan**
- [x] Create a UOM in Zuora DEV and UAT for _Free event tickets_ and get rid of the 'events' prefix in the usage description.
- [x] Ensure the UOM configuration is reproduced to the other tiers
 
@rtyley @tudorraul

### DIFF
--- a/frontend/app/controllers/Event.scala
+++ b/frontend/app/controllers/Event.scala
@@ -204,6 +204,7 @@ trait Event extends Controller with ActivityTracking {
         val eventUrl = ebCode.fold(Uri.parse(event.url))(c => event.url ? ("discount" -> c.code))
         val memberData = MemberData(request.member.salesforceContactId, request.user.id, request.member.tier.name, campaignCode = extractCampaignCode(request))
         track(EventActivity("redirectToEventbrite", Some(memberData), EventData(event)))(request.user)
+
         Found(eventUrl)
           .withCookies(Cookie(eventCookie(event), "", Some(3600)))
       }
--- a/frontend/app/controllers/Event.scala
+++ b/frontend/app/controllers/Event.scala
@@ -197,7 +197,6 @@ trait Event extends Controller with ActivityTracking {
     event.internalTicketing.exists(_.salesDates.tierCanBuyTicket(tier))
 
   private def eventCookie(event: RichEvent) = s"mem-event-${event.id}"
-  private val isComplimentaryKey = "is-complimentary-ticket"
 
   private def redirectToEventbrite(request: AnyMemberTierRequest[AnyContent], event: RichEvent): Future[Result] =
     Timing.record(event.service.wsMetrics, s"user-sent-to-eventbrite-${request.member.tier}") {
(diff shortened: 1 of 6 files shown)


frontend/app/controllers/Event.scala
```diff
@@ -218,7 +226,14 @@ trait Event extends Controller with ActivityTracking {
       } yield {
         event.service.getOrder(oid).map { order =>
           trackConversionToThankyou(request, event, Some(order), Some(request.member))
-          Ok(views.html.event.thankyou(event, order)).discardingCookies(DiscardingCookie(eventCookie(event)))
+
+          request.session.get(isComplimentaryKey) foreach ( _ =>
```
rtyley: :-1:  This looks wrong to me - it looks like you're assuming that because we showed complimentary tickets to the user (`session.get(isComplimentaryKey)`), the user chose to use them? I believe we should instead be using the Attendee information from the EventBrite APi data for the Order https://www.eventbrite.com/developer/v3/formats/attendee/#ebapi-std:format-attendee - look at the ticket classes consumed to see if a compilmentary ticket was used.

afiore: You are right, displaying the free ticket doesn't actually imply that one uses it. I am going to change the implementation and try to record the free event ticket usage based on the attendee data

afiore:
@rtyley I have updated the implementation so that it uses the `EBAttendee` object already returned as part of the EventBrite Order response. By extending `EBAttendee` with a `ticket_class_id` field we can easily verify that the purchase has actually used a complimentary ticket.

Due to the way in which the data is modelled in the EventBrite API, [this code](https://github.com/guardian/membership-frontend/pull/667/files#diff-abb6c51f77945af0f3a1eca1252ea19bR130) caters for the possibility of multiple attendees, potentially buying more than one ticket. Happy to revisit this choice if you see any possible issues with it.

frontend/app/controllers/Event.scala
```diff
@@ -217,8 +223,15 @@ trait Event extends Controller with ActivityTracking {
         event <- EventbriteService.getEvent(id)
       } yield {
         event.service.getOrder(oid).map { order =>
+          val count = memberService.countComplimentaryTicketsUsed(event, order)
+          if (count > 0) {
```
rtyley: This will currently always be 0 or 1, right?

afiore: Yep, maybe this method should be a boolean afterall

frontend/app/services/zuora/ZuoraActions.scala
```diff
@@ -352,3 +352,25 @@ case class UpgradePlan(subscriptionId: String, subscriptionRatePlanId: String, n
     </ns1:amend>
   }
 }
+
+case class CreateFreeEventUsage(accountId: String, description: String, quantity: Int, subscriptionId: Option[String]) extends ZuoraAction[CreateResult] {
```
rtyley: I think the `subscriptionId` probably shouldn't be an option - if we can't find a subscription for the user, something is very wrong!

rtyley:
:+1:

PR #671, opened 2015-08-12
Remove optimizley and crazyegg [Depends on #674]

davidrapson:
This PR can only be merged once https://membership.theguardian.com/join-challenger has been upgraded to the main homepage. Removes Optimizley and Crazyegg as we will no longer need them.

@afiore @tudorraul


afiore:
:+1:

PR #672, opened 2015-08-12
Prepare Join Challenger for launch

davidrapson:
Refactors join challenger template out into generalised components rather than test/template-specific styling.

Redesign on comparison table to work repsonsively, this is the minimum set of changes needed to launch the page as this will be partially revisited as part of the redesign anyway.

@afiore @tudorraul 

screen shot 2015-08-11 at 17 28 36
screen shot 2015-08-11 at 17 29 13
screen shot 2015-08-11 at 17 28 48
screen shot 2015-08-11 at 17 29 01

---
screen shot 2015-08-12 at 12 14 22


frontend/app/model/Nav.scala
```diff
@@ -10,6 +10,8 @@ object Nav {
     subNav: Seq[NavItem] = Nil
   )
 
+  case class NavAnchor(href: String, title: String)
```
tudorraul: this case class is only used in the `views` package, and should be contained within it.

frontend/assets/stylesheets/base/_tables.scss
```diff
@@ -55,3 +55,86 @@
         background-color: darken($c-neutral8, 4%);
     }
 }
+
+/* Comparison Table
+   ========================================================================== */
+
+$_table-tier-height: 150px;
```
tudorraul: is this an exception to the guss grid?

davidrapson: Yep, it's mostly a magic number, for the height of the tier-headers on small screens.

frontend/app/model/Nav.scala
davidrapson: Not sure I agree, it's common to the other Navigation code in here, given it's a small case class I don't quite see the benefit of creating something dedicated in the `views` package for it.

davidrapson:
@TesterSpike Mind giving this a quick once over?

tudorraul:
:+1:

PR #673, opened 2015-08-12
CSP: Don't log w/empty message and tag with UA

davidrapson:
Getting some weird Sentry logging, according to the Sentry source this is due to the message being empty. This PR filters for empty messages, and also logs the UA for the CSP report to aid with debugging. @rtyley 

screen shot 2015-08-12 at 14 36 33

### DIFF
--- a/frontend/app/controllers/Security.scala
+++ b/frontend/app/controllers/Security.scala
@@ -56,13 +56,20 @@ object Security extends Controller with LazyLogging {
 
     SentryLogging.ravenOpt.fold {
       logger.error(report.message)
-    } {
-      _.sendEvent(new EventBuilder()
-        .withMessage(report.message)
-        .withLevel(RavenEvent.Level.INFO)
-        .withTag("CSP Directive", report.directiveTag)
-        .build()
-      )
+    } { raven =>
+      if(report.message.nonEmpty) {
+
+        val eventBuilder = new EventBuilder()
+          .withMessage(report.message)
+          .withLevel(RavenEvent.Level.INFO)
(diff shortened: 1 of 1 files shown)


frontend/app/controllers/Security.scala
```diff
@@ -56,13 +56,20 @@ object Security extends Controller with LazyLogging {
...
       logger.error(report.message)
-    } {
-      _.sendEvent(new EventBuilder()
-        .withMessage(report.message)
-        .withLevel(RavenEvent.Level.INFO)
-        .withTag("CSP Directive", report.directiveTag)
-        .build()
-      )
+    } { raven =>
+      if(report.message.nonEmpty) {
```
rtyley: I don't see how the message _can_ be empty - tho' we do see it as empty in Sentry - so I would not expect this `if` condition to stop the weird reports we're seeing in Sentry.

frontend/app/controllers/Security.scala
```diff
@@ -56,13 +56,20 @@ object Security extends Controller with LazyLogging {
...
+
+        val eventBuilder = new EventBuilder()
+          .withMessage(report.message)
+          .withLevel(RavenEvent.Level.INFO)
+          .withTag("CSP Directive", report.directiveTag)
+
+        for(ua <- request.headers.get(USER_AGENT)) {
+          eventBuilder.withTag("CSP User Agent", ua)
```
rtyley: I would guess that this will be more helpful, but I guess we'll see...

rtyley:
:+1:

PR #675, opened 2015-08-12
Enforce free ticket allowance

afiore:
Keep free event tickets usage into account when issuing partners and patrons with complimentary ticket codes. We hardcode the allowance to 6, and check that the total number of relevant usage records in Zuora is lower or equal to this value. The query used to fetch Usage records from zuora is scoped with the following parameters:

- `Usage.StartDateTime >= subscription.termStartDate`
- `Usage.subscriptionId == subscription.id`

**Note:** Depends on #667


afiore:
cc @rtyley @tudorraul

frontend/app/services/SubscriptionService.scala
```diff
@@ -155,6 +156,17 @@ class SubscriptionService(val tierPlanRateIds: Map[ProductRatePlan, String],
     } yield subscriptionDetails
   }
 
+  def getUsageCountWithinTerm(memberId: MemberId, unitOfMeasure: String): Future[Int] = (for {
+      subscriptions <- getSubscriptions(memberId)
+      details <- getSubscriptionDetails(subscriptions.head.id) if subscriptions.nonEmpty
+      startDate = formatDateTime(details.contractAcceptanceDate)
+      endDate = formatDateTime(details.effectiveStartDate + 1.year)
```
rtyley: Lose the endDate (the current term should be ongoing) - and see if you can get term start and end date out of the REST API https://knowledgecenter.zuora.com/BC_Developers/REST_API/B_REST_API_reference/Subscriptions/4_Get_subscriptions_by_account

frontend/app/model/FeatureChoice.scala
```diff
@@ -30,6 +32,8 @@ case object Books extends FeatureChoice {
   override val label = "4 free books"
 }
 case object FreeEventTickets extends FeatureChoice {
+  val allowance = zuoraFreeEventTicketsAllowance
   override val zuoraCode = "Events"
-  override val label = "6 free events"
+  override val label = s"$allowance free events"
+  val uom = "Events"
```
tudorraul: what does `uom` mean?

frontend/app/model/Zuora.scala
```diff
@@ -43,6 +44,50 @@ object Zuora {
     override def getMessage: String = s"$code: $message"
   }
 
+  object Rest {
```
tudorraul: What's the purpose of this object?

frontend/app/model/FeatureChoice.scala
afiore: This is [fully documented](https://knowledgecenter.zuora.com/BC_Developers/SOAP_API/E1_SOAP_API_Object_Reference/Usage) Zuora jargon, it stands for unit of measure

frontend/app/model/Zuora.scala
afiore: It models the different types of REST API responses into Scala classes

frontend/app/model/FeatureChoice.scala
tudorraul: Is it worthwhile putting a comment or renaming the val to smth like `unitName` or `unitsToMeasure`?

frontend/app/services/SubscriptionService.scala
```diff
@@ -155,6 +156,17 @@ class SubscriptionService(val tierPlanRateIds: Map[ProductRatePlan, String],
     } yield subscriptionDetails
   }
 
+  def getUsageCountWithinTerm(memberId: MemberId, unitOfMeasure: String): Future[Int] = (for {
+    account <- getAccount(memberId)
+    subscriptions <- zuoraRestService.subscriptionsByAccount(account.id)
+    subscription = subscriptions.head if subscriptions.nonEmpty
+    startDate = formatDateTime(subscription.termStartDate)
+    whereClause = s"StartDateTime >= '$startDate' AND SubscriptionID = '${subscription.id}'"
```
tudorraul: I would create a new method in the Zuora service to contain this logic

frontend/app/services/SubscriptionService.scala
```diff
@@ -155,6 +156,17 @@ class SubscriptionService(val tierPlanRateIds: Map[ProductRatePlan, String],
     } yield subscriptionDetails
   }
 
+  def getUsageCountWithinTerm(memberId: MemberId, unitOfMeasure: String): Future[Int] = (for {
+    account <- getAccount(memberId)
+    subscriptions <- zuoraRestService.subscriptionsByAccount(account.id)
```
tudorraul: A `zuoraRestService.lastSubscriptionsByAccount` method might be useful

frontend/app/services/SubscriptionService.scala
afiore: What Zuora service? We have both the REST and the SOAP one... This method uses both, so I have placed it in SubscriptionService

frontend/app/services/SubscriptionService.scala
afiore: It might, but we first need to get right the logic whereby we filter and disambiguate zuora subscription records (e.g. by productRatePlanId). The next story in the backlog is about that, so I would do it then

frontend/app/services/SubscriptionService.scala
tudorraul: I think only the SOAP service supports where clauses

frontend/app/services/zuora/ZuoraRestService.scala
```diff
@@ -32,41 +31,79 @@ class ZuoraRestService(config: ZuoraApiConfig) {
...
+    Timing.record(metrics, "subscriptionsByAccount") {
       get(s"subscriptions/accounts/$accountKey").map { response =>
         metrics.putResponseCode(response.code, "GET")
-        Try(extractFeatures(response.body().string())) match {
-          case Success(x) => x
-          case Failure(e) =>
-            metrics.recordError()
-            throw e
+        parseResponse[List[Rest.Subscription]](response) match {
+          case Rest.Success(subcriptions) => subcriptions
```
tudorraul: typo

frontend/app/services/SubscriptionService.scala
tudorraul: The same logic is applied to both clients of existing `subscriptionsByAccount` clients, and this is getting the last subscription. We can be DRYer at this point.

frontend/app/services/SubscriptionService.scala
afiore: As it is, this methods depends on the `ZuoraRestService` in order to fetch the subscription record. I am not really keen on moving this into the SoapClient and ending up with a soap client that somehow depends on the rest client...

tudorraul:
:+1:

chrisjowen:
Just a quick point I can't see the usage of the unitOfMeasure param in 

SubscriptionService:getUsageCountWithinTerm -- line 159

PR #678, opened 2015-08-13
Remove unnecessary reference to ~/.aws/config

rtyley:
The AWS SDK for Java reads from only `~/.aws/credentials`, not  `~/.aws/config` so far as I can see (corrected thanks to @davidrapson !):

http://docs.aws.amazon.com/AWSSdkDocsJava/latest/DeveloperGuide/credentials.html

(introduced with https://github.com/guardian/membership-frontend/commit/7ccfe6df)

We use the credentials to download a key file from S3 to connect to the Google Directory API.

cc @davidrapson

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,14 +80,6 @@ region = eu-west-1
 
 ```
 
-In `~/.aws/config` add the following:
-
-```
-[default]
-output = json
-region = eu-west-1
-```
-
 ### Ubuntu
 
 In an ideal world, your Ubuntu package install would be:


davidrapson:
@rtyley Definitely this way around? " typically located at `~/.aws/credentials`" in that article.

rtyley:
> Definitely this way around? " typically located at ~/.aws/credentials" in that article.

Yes! I will update the description.

davidrapson:
:shipit:

PR #679, opened 2015-08-13
Display event/books choice and complimentary event tickets to test users only

afiore:
Wrap relevant code blocks in `isTestUser(user)` if statements so that upcoming features can be QA'ed but are not accessible to end users.

cc @rtyley


tudorraul:
:+1: LGTM

rtyley:
Cool :+1:  As discussed, we could probably look into some kind of feature-enabling switch cookie, but this does at least put us in a place where the events team can start adding the complimentary ticket category.

PR #683, opened 2015-08-17
Display event/books choice and complimentary event tickets to test users only

afiore:
This is a revisited attempt at triggering the event/book choice and the complimentary tickets functionality  to test users only (previous failed attempt here: #679). The root cause of the problem was the call to `TouchPointBackend.forUser(IdMinimalUser(member.identityId, None))` which was resulting in test users never being picked up, and produced a runtime error whereby the app could not find a Zuora account while rendering in the new joiner thank you page.

cc @tudorraul


frontend/app/services/MemberService.scala
```diff
@@ -54,6 +56,9 @@ class FrontendMemberRepository(salesforceConfig: SalesforceConfig) extends Membe
 }
 
 trait MemberService extends LazyLogging with ActivityTracking {
+  implicit def memberToIdMinimalUser(member: Member): IdMinimalUser =
+    IdMinimalUser(member.identityId, member.firstName)
```
tudorraul: please add a comment to specify the importance of `firstName` here

frontend/app/services/MemberService.scala
```diff
@@ -125,7 +130,7 @@ trait MemberService extends LazyLogging with ActivityTracking {
 
   def retrieveComplimentaryTickets(member: Member, event: RichEvent): Future[Seq[EBTicketClass]] = {
     for {
-      memberTierFeatures <- TouchpointBackend.forUser(IdMinimalUser(member.identityId, None))
+      memberTierFeatures <- TouchpointBackend.forUser(member)
```
tudorraul: I wouldn't use implicit conversion here

tudorraul:
:+1: bar the comments

PR #685, opened 2015-08-17
Move JavaScript tests into assets/javascripts

davidrapson:
Move JavaScript tests into `assets/javascripts`. I think it's clearer to keep JavaScript tests alongside the rest of the JS rather than in with mainly scala tests.

@afiore


afiore:
:+1: Agreed

PR #689, opened 2015-08-19
Bump maxDiscountQuantityAvailable to 3

afiore:
As we now offer 1 complimentary ticket and 2 discounted ones,
we need to increase this limit to 3 instead of 2.

cc @tudorraul


davidrapson:
:+1:

PR #691, opened 2015-08-20
Remove 'Join' from homepage title

davidrapson:
One minor detail missed when migrating join challenger page to home page. Removes 'Join' from the homepage title.

screen shot 2015-08-20 at 12 27 52
screen shot 2015-08-20 at 12 28 00

@afiore

### DIFF
--- a/frontend/app/views/index.scala.html
+++ b/frontend/app/views/index.scala.html
@@ -8,7 +8,7 @@
 @import views.support.NavAnchor
 @import views.support.Asset
 
-@main("Join", pageInfo) {
+@main("", pageInfo) {
 
     @* ===== Intro ===== *@
     <div class="hero-banner" style="background-image: url(@Asset.at("images/join-challenger/event-grid.jpg"))">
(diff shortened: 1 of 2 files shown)


frontend/app/views/index.scala.html
```diff
@@ -8,7 +8,7 @@
 @import views.support.NavAnchor
 @import views.support.Asset
 
-@main("Join", pageInfo) {
+@main("", pageInfo) {
```
davidrapson: Ideally I'd make title an `Option` in `main.scala.html`, but this is the minimum change needed to get this out.

afiore:
:+1:

frontend/app/views/index.scala.html
afiore: Yep, I would make it option and move title into `pageInfo` if possible...

davidrapson: @afiore Weirdly `title` is **already** in PageInfo, so not sure why we include it here too tbh.

PR #694, opened 2015-08-20
Simplify benefits ahead of tier recalibration

davidrapson:
This PR simplifies how we describe benefits ahead of the tier recalibration launch. We can release these changes ahead of time as we're not including the new benefits here.

## Supporter pages

Before/After
screen shot 2015-08-20 at 17 46 41
screen shot 2015-08-20 at 17 40 33

## Choose tier

Before/After
choose-tier-before
after

## Change tier

Before/After
change-tier-before
screen shot 2015-08-20 at 17 44 14

## Cancel

Before/After
cancel-before
cancel-after

@afiore


frontend/app/model/Benefits.scala
```diff
@@ -58,57 +55,64 @@ object Benefits {
 
   val friendBenefitsList = benefitsFilter(
     "book_tickets",
-    "digital_digest",
-    "video_highlights"
+    "offers_competitions",
```
tudorraul: Relying on strings might cause issues when typos occur. On the other hand this is contained in the same file, and the risks are small.

tudorraul:
:+1: bar the comment

frontend/app/model/Benefits.scala
davidrapson: @tudorraul I'd definitely prefer not to use strings, how would you approach this?

frontend/app/model/Benefits.scala
```diff
@@ -24,17 +24,14 @@ object Benefits {
   )
 
   val allBenefits = Seq(
```
tudorraul: This looks to be used only in the `views` package and would be better located there, I think.

frontend/app/model/Benefits.scala
tudorraul: name each `BenefitItem` object and use its name onwards

PR #697, opened 2015-08-24
Disambiguating Zuora subscription data from Membership vs Subscriptions

afiore:
_PR description by @rtyley :_

These are some fairly intense changes to the `SubscriptionService` to ensure that it correctly disambiguates between Zuora data for Membership and for Subscriptions (now the Digital Pack 2.0 is soon to launch).

The key thing to know about Zuora data:

* **Zuora accounts** do **not** directly relate to Membership vs Subscriptions (you can not say 'this Zuora Account is a Membership Account'). A Zuora account only has to exist to provide a default payment method - so if a user is only paying with credit card, they should only have one account, but if they have both direct debit AND credit card, they will need _two_ Zuora accounts. PayPal, etc, would require a third.
* **Zuora Subscriptions** are only _loosely_ associated with Membership or Subscriptions. You could have a Zuora Subscription which had both membership.theguardian.com & subscribe.theguardian.com products on it - we don't currently create those, but there is no guarantee that they won't occur in future.
* **Zuora Products** _are_ associated with either Membership or Subscriptions. A Product might be 'Supporter' membership, or 'Digital Pack' for subscribe.theguardian.com.

So when it comes to working out which _Zuora Subscription_ you want to work with, you have to see which Products it has. If it has Membership Products, then it's a Membership subscription. The Zuora account that the subscription falls under could then be considered a 'Membership' account for some purposes - but might also include Zuora Subscriptions for subscribe.theguardian.com.

We are taking advantage of the Guardian custom field `ProductType` which already exists on Zuora Products. For Partner, Patron, Supporter, Staff Membership, this is correctly set to 'Membership'. For 'Digital Pack' the `ProductType` is 'Digital Pack' which I think may not be a good value for it - it should probably be 'Subscriptions'.

cc @tudorraul @vivekkr 

https://trello.com/c/ltEhCsjV/37-filter-non-membership-zoura-subscriptions-to-avoid-clash-with-subscribe-theguardian-com

### DIFF
--- a/frontend/app/model/Zuora.scala
+++ b/frontend/app/model/Zuora.scala
@@ -55,9 +55,9 @@ object Zuora {
   }
   case class InternalError(code: String, message: String) extends Error
 
-  case class SubscriptionStatus(current: Subscription, future: Option[Subscription], amendType: Option[String]) {
-    val currentId = current.id
-    val futureIdOpt = future.map(_.id)
+  case class SubscriptionStatus(currentVersion: Subscription, futureVersionOpt: Option[Subscription], amendType: Option[String]) {
+    val currentVersionId = currentVersion.id
+    val futureVersionIdOpt = futureVersionOpt.map(_.id)
     val cancelled = amendType.contains("Cancellation")
   }
 
--- a/frontend/app/model/Zuora.scala
+++ b/frontend/app/model/Zuora.scala
@@ -55,8 +55,10 @@ object Zuora {
   }
   case class InternalError(code: String, message: String) extends Error
 
-  case class SubscriptionStatus(current: String, future: Option[String], amendType: Option[String]) {
-    val cancelled = amendType.exists(_ == "Cancellation")
+  case class SubscriptionStatus(current: Subscription, future: Option[Subscription], amendType: Option[String]) {
+    val currentId = current.id
+    val futureIdOpt = future.map(_.id)
+    val cancelled = amendType.contains("Cancellation")
   }
 
   case class SubscriptionDetails(planName: String, planAmount: Float, effectiveStartDate: DateTime,
--- a/frontend/app/services/SubscriptionService.scala
+++ b/frontend/app/services/SubscriptionService.scala
@@ -24,10 +24,10 @@ case class SubscriptionServiceError(s: String) extends Throwable {
 trait AmendSubscription {
   self: SubscriptionService =>
 
-  private def checkForPendingAmendments(memberId: MemberId)(fn: String => Future[AmendResult]): Future[AmendResult] = {
+  private def checkForPendingAmendments(memberId: MemberId)(fn: SubscriptionStatus => Future[AmendResult]): Future[AmendResult] = {
     getSubscriptionStatus(memberId).flatMap { subscriptionStatus =>
-      if (subscriptionStatus.future.isEmpty) {
-        fn(subscriptionStatus.current)
+      if (subscriptionStatus.futureIdOpt.isEmpty) {
+        fn(subscriptionStatus)
       } else {
         throw SubscriptionServiceError("Cannot amend subscription, amendments are already pending")
       }
@@ -35,11 +35,12 @@ trait AmendSubscription {
   }
 
   def cancelSubscription(memberId: MemberId, instant: Boolean): Future[AmendResult] = {
-    checkForPendingAmendments(memberId) { subscriptionId =>
+    checkForPendingAmendments(memberId) { subscriptionStatus =>
+      val currentSubscriptionVersion = subscriptionStatus.current
       for {
-        subscriptionDetails <- getSubscriptionDetails(subscriptionId)
+        subscriptionDetails <- getSubscriptionDetails(currentSubscriptionVersion)
         cancelDate = if (instant) DateTime.now else subscriptionDetails.chargedThroughDate.getOrElse(DateTime.now())
-        result <- zuoraSoapService.authenticatedRequest(CancelPlan(subscriptionId, subscriptionDetails.ratePlanId, cancelDate))
+        result <- zuoraSoapService.authenticatedRequest(CancelPlan(currentSubscriptionVersion.id, subscriptionDetails.ratePlanId, cancelDate))
       } yield result
     }
   }
@@ -51,12 +52,13 @@ trait AmendSubscription {
     def effectiveFrom(subscriptionDetails: SubscriptionDetails) = subscriptionDetails.chargedThroughDate.getOrElse(subscriptionDetails.contractAcceptanceDate)
 
 
-    checkForPendingAmendments(memberId) { subscriptionId =>
+    checkForPendingAmendments(memberId) { subscriptionStatus =>
+      val currentSubscriptionVersion = subscriptionStatus.current
       for {
-        subscriptionDetails <- getSubscriptionDetails(subscriptionId)
+        subscriptionDetails <- getSubscriptionDetails(subscriptionStatus.current)
         dateToMakeDowngradeEffectiveFrom = effectiveFrom(subscriptionDetails)
 
-        result <- zuoraSoapService.authenticatedRequest(DowngradePlan(subscriptionId, subscriptionDetails.ratePlanId,
+        result <- zuoraSoapService.authenticatedRequest(DowngradePlan(currentSubscriptionVersion.id, subscriptionDetails.ratePlanId,
           tierPlanRateIds(newTierPlan), dateToMakeDowngradeEffectiveFrom))
       } yield result
     }
@@ -65,7 +67,8 @@ trait AmendSubscription {
   def upgradeSubscription(memberId: MemberId, newTierPlan: TierPlan, preview: Boolean, featureChoice: Set[FeatureChoice]): Future[AmendResult] = {
     import SubscriptionService._
 
-    checkForPendingAmendments(memberId) { subscriptionId =>
+    checkForPendingAmendments(memberId) { subscriptionStatus =>
+      val subscriptionId = subscriptionStatus.currentId
       for {
         zuoraFeatures <- zuoraSoapService.featuresSupplier.get()
         ratePlan <- zuoraSoapService.queryOne[RatePlan](s"SubscriptionId='$subscriptionId'")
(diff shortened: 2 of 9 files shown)


frontend/app/services/zuora/ZuoraRestService.scala
```diff
@@ -32,12 +33,16 @@ class ZuoraRestService(config: ZuoraApiConfig) extends LazyLogging {
     }
   }
 
+  val productCatalogSupplier = new FutureSupplier[ProductCatalog](productCatalog)
```
rtyley: @vivekkr @tudorraul this is where you can start grabbing the Zuora Product Catalog now

frontend/app/model/Zuora.scala
```diff
@@ -55,8 +55,10 @@ object Zuora {
   }
   case class InternalError(code: String, message: String) extends Error
 
-  case class SubscriptionStatus(current: String, future: Option[String], amendType: Option[String]) {
-    val cancelled = amendType.exists(_ == "Cancellation")
+  case class SubscriptionStatus(current: Subscription, future: Option[Subscription], amendType: Option[String]) {
```
tudorraul: This looks to suggest a subscription _change_ rather than a subscription status.

tudorraul: Also, the wording here can be a bit confusing, `future` suggests a Scala `Future`. Perhaps `newVersion` might hold the meaning better.

frontend/app/services/SubscriptionService.scala
```diff
@@ -105,76 +106,74 @@ object SubscriptionService {
...
+    amendments <- zuoraSoapService.query[Amendment](where)
+  } yield {
+      val latestSubscription = subscriptionVersions.maxBy(_.version)
+      sortAmendments(subscriptionVersions, amendments)
         .find(_.contractEffectiveDate.isAfterNow)
-        .fold(SubscriptionStatus(latestSubscriptionId, None, None)) { amendment =>
-          SubscriptionStatus(amendment.subscriptionId, Some(latestSubscriptionId), Some(amendment.amendType))
-        }
+        .fold(SubscriptionStatus(latestSubscription, None, None)) { futureAmendment =>
+          logger.info(s"subscription = ${subscription.id}, latestSubscriptionId ${latestSubscription.id}")
```
tudorraul: what is this logging statement helping us with?

frontend/app/services/SubscriptionService.scala
```diff
@@ -105,76 +106,74 @@ object SubscriptionService {
...
+  def memberTierFeatures(memberId: MemberId): Future[Seq[Rest.Feature]] = for {
+    (_, subscription) <- accountWithLatestMembershipSubscription(memberId)
+    catalog  <- zuoraRestService.productCatalogSupplier.get
+  } yield subscription
+      .latestWhiteListedRatePlan(catalog.productIdsOfType(MembershipProductType)).toSeq
+      .flatMap(_.subscriptionProductFeatures)
+
+  /**
+   * @return the current subscription version of the user, and the future subscription version, if
+   *         they have a pending downgrade.
```
tudorraul: Does this method work only for downgrades, or does it for upgrades too?

frontend/app/services/SubscriptionService.scala
```diff
@@ -105,76 +106,74 @@ object SubscriptionService {
...
+   */
+  def getSubscriptionStatus(memberId: MemberId): Future[SubscriptionStatus] = for {
+    (_, subscription) <- accountWithLatestMembershipSubscription(memberId)
+    subscriptionVersions <- subscriptionVersions(subscription.subscriptionNumber)
+    where = subscriptionVersions.map { sub => s"SubscriptionId='${sub.id}'" }.mkString(" OR ")
+    amendments <- zuoraSoapService.query[Amendment](where)
+  } yield {
+      val latestSubscription = subscriptionVersions.maxBy(_.version)
+      sortAmendments(subscriptionVersions, amendments)
```
tudorraul: a bit unclear why sorting is necessary here?

frontend/app/services/SubscriptionService.scala
```diff
@@ -105,76 +106,74 @@ object SubscriptionService {
...
-
-      where = subscriptions.map { sub => s"SubscriptionId='${sub.id}'" }.mkString(" OR ")
-      amendments <- zuoraSoapService.query[Amendment](where)
-    } yield {
-      val latestSubscriptionId = sortSubscriptions(subscriptions).last.id
+  import SubscriptionService._
+  val MembershipProductType = "Membership"
+
+  private def subscriptionVersions(subscriptionNumber: String): Future[Seq[Subscription]] = for {
+    subscriptions <- zuoraSoapService.query[Subscription](s"Name = '$subscriptionNumber'")
```
tudorraul: no need for a `for` comprehension here

frontend/app/services/SubscriptionService.scala
```diff
@@ -105,76 +106,76 @@ object SubscriptionService {
...
+  } yield subscriptions
+
+  def accountWithLatestMembershipSubscription(member: MemberId): Future[(Account, Rest.Subscription)] = for {
+    accounts <- zuoraSoapService.query[Account](s"crmId='${member.salesforceAccountId}'")
```
tudorraul: I am in favour of concealing this ZQL logic in the `ZuoraSoapService`

frontend/app/services/zuora/Rest.scala
```diff
@@ -31,6 +31,16 @@ object Rest {
   case object Cancelled extends SubscriptionStatus
   case object Expired extends SubscriptionStatus
 
+  /*
+   * Zuora product catalogue entry: https://knowledgecenter.zuora.com/BC_Developers/REST_API/B_REST_API_reference/Catalog
+   */
+  case class ProductCatalog(products: List[Product]) {
```
tudorraul: should the return generic type be uniformed in these 2 methods? i.e. both `Seq` or bothe `Set`

frontend/app/services/SubscriptionService.scala
afiore: Agreed, but out of the scope of this PR really, as it is something we do in a number of places

afiore: Also, notice that `query` and `queryOne` are pretty much the only public method in that class

frontend/app/services/SubscriptionService.scala
afiore: This currently works only for upgrades (which was the case also for the previous versions of this code). We plan on handling downgrades after we close this PR

rtyley:
cc @TesterSpike

frontend/app/services/SubscriptionService.scala
```diff
@@ -185,46 +186,44 @@ class SubscriptionService(val tierPlanRateIds: Map[ProductRatePlan, String],
...
   }
 
   def getMembershipSubscriptionSummary(memberId: MemberId): Future[MembershipSummary] = {
-
-    val latestSubF = {
-      for (subscriptions <- getSubscriptions(memberId))
-      yield sortSubscriptions(subscriptions).last
-    }
+    val latestSubF = for {
+      (_, subscription) <- accountWithLatestMembershipSubscription(memberId)
```
tudorraul: There's quite a bit of redundancy happening behind the scenes here, and it's because we don't have a get subscription by account anymore. Creating one back and using that instead of the heavy `accountWithLatestMembershipSubscription` will cut back on this redundancy.

frontend/app/services/SubscriptionService.scala
```diff
@@ -185,46 +186,44 @@ class SubscriptionService(val tierPlanRateIds: Map[ProductRatePlan, String],
...
-      } yield {
-        assert(result.invoiceItems.nonEmpty, "Subscription with delayed payment returning zero invoice items in SubscriptionDetailsViaAmend call")
-        val firstPreviewInvoice = result.invoiceItems.sortBy(_.serviceStartDate).head
-
-        MembershipSummary(latestSubscription.termStartDate, firstPreviewInvoice.serviceEndDate, None,
-          subscriptionDetails.planAmount, firstPreviewInvoice.price, firstPreviewInvoice.serviceStartDate, firstPreviewInvoice.renewalDate )
-      }
+    def getSummaryViaSubscriptionAmend(memberId: MemberId) = for {
+      latestSubscription <- latestSubF
+      subscriptionDetailsF = getSubscriptionDetails(latestSubscription)
```
tudorraul: why the indirection?

tudorraul: why the indirection?

frontend/app/services/zuora/ZuoraRestService.scala
tudorraul: :+1:

frontend/app/services/SubscriptionService.scala
chrisjowen: Confusion for me comes form the fact that we are just suffixing everything service which I think is one of those dump bucket words like xxxManager etc. In this case the:

- ZuoraSoapService is acting more like a custom Soap client (as @afiore said only query and queryOne are public). 
- SubscriptionService is mainly using ZuoraSoapService to query and is really acting a lot like a repository or similar.   
- MembershipService contains much more of an orchestration/facade logic. 

These are three completely different types of responsibilities with the same suffix. Also theres a lot of stuff going on inside these files that muddle there responsibilities like mapping/transformation code and not so related objects like MemberService containing FrontendMemberRepository a salesforce client from what I can tell. 

Basically where things go and what they should be named is confusing as hell for me

tudorraul:
:+1: bar the comments

TesterSpike:
:+1:  the automated tests have mostly passed (30 out 35) - those that didn't appear to be an issue with a previous change breaking the tests.

frontend/app/services/SubscriptionService.scala
rtyley: > a bit unclear why sorting is necessary here, since only one value is then needed?

I've updated the code to try to make it clearer why this is necessary:

https://github.com/guardian/membership-
frontend/blob/fd31fe6aba51/frontend/app/services/SubscriptionService.scala#L83-L113

PR #699, opened 2015-08-24
Post QA fixes

afiore:
- Fixes an off-by-one error in the logic enforcing the free ticket allowance
- Sets `maxDiscountQuantityAvailable` also for local events

@rtyley is this setting appropriate for `LocalEventService`? We had to bump liveEventService to 3 in order to cater for the possibility of a user taking a combination of 2 discounted tickets and 1 free one.

### DIFF
--- a/frontend/app/services/EventbriteService.scala
+++ b/frontend/app/services/EventbriteService.scala
@@ -143,7 +143,7 @@ object GuardianLiveEventService extends LiveService {
 
 object LocalEventService extends LiveService {
   val apiToken = Config.eventbriteLocalApiToken
-  val maxDiscountQuantityAvailable = 3
+  val maxDiscountQuantityAvailable = 2
   val wsMetrics = new EventbriteMetrics("Local")
 
   def mkRichEvent(event: EBEvent): Future[RichEvent] =  for { gridImageOpt <- gridImageFor(event) }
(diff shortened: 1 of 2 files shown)


frontend/app/services/EventbriteService.scala
```diff
@@ -143,7 +143,7 @@ object GuardianLiveEventService extends LiveService {
 
 object LocalEventService extends LiveService {
   val apiToken = Config.eventbriteLocalApiToken
-  val maxDiscountQuantityAvailable = 2 //TODO are these discounts correct for local?
+  val maxDiscountQuantityAvailable = 3
```
davidrapson: Free tickets are only for Live events, so not sure we need this.

davidrapson:
:+1: Aside from the comment about Local tickets.

frontend/app/services/EventbriteService.scala
afiore: In order to QA free event tickets we rely on the Eventbrite dev account, which is configured to be  handled through the `LocalEventService`

rtyley: You can use your local copy of the `/etc/gu/membership-keys.conf` file to make our Dev Eventbrite account supply the events for Guardian Live, so testing with Local events is unnecessary.

The free event tickets are only for Guardian Live Events, , so this value here should stay at 2

PR #701, opened 2015-08-25
Cleanup: Remove deprecated benefits icons

davidrapson:
With the latest redesign updates we don't need them anymore.

@afiore


afiore:
:+1:

PR #702, opened 2015-08-25
Cleanup: Remove admin templates

davidrapson:
This admin work was abandoned, we don't need to keep unused code around.


afiore:
:+1:

PR #700, opened 2015-08-25
Homepage recalibration design updates

davidrapson:
Re-opening https://github.com/guardian/membership-frontend/pull/698 now that the redesign has been merged in (!). This PR updates the homepage with the design updates for the tier recalibration.

---

**TODO**

- [x] Extract out any cleanup work into a PR which can go out separately
- [x] Extract slideshow into separate PR & use new images for Space (make image gallery)
- [x] Release https://github.com/guardian/membership-frontend/pull/707
- [ ] Release https://github.com/guardian/membership-frontend/pull/715
- [ ] Final cross-browser testing and review

homepage-recalibration

@afiore @tudorraul

PR #703, opened 2015-08-25
Recalibration prep

davidrapson:
Pulls a few commits out of https://github.com/guardian/membership-frontend/pull/700 to make it easier to review.

These changes are some groundwork for the recalibration updates, abstracts some common styles a bit more (and fixes a minor markup error with the main navigation as a result). Objects in this context are distinct from components in that they are lower level styling traits (e.g., grid should also move to objects)

See: http://csswizardry.com/2015/03/more-transparent-ui-code-with-namespaces/#object-namespaces-o-

@afiore


afiore:
:+1:

PR #708, opened 2015-08-26
Fix footer help and feedback links - wrong object used for reverse-routing

rtyley:
Following yesterday's update to https://membership.theguardian.com/ 2 links were broken in the footer: 'Help' and 'feedback' - they both linked to:

https://membership.theguardian.com/Action(parser=BodyParser((no%20name)))

sfas

The fix is to use the correct reverse-router object `routes`, which returns objects of the `play.api.mvc.Call` type - calling `toString()` on that gives you a relative url (can't do the absolute, because we don't have a request header here).

Note that sadly Intellij doesn't seem to know about the routes and marks them red...
image

https://trello.com/c/mD7FyJrv/65-404-error-s-on-website

cc @afiore @davidrapson

### DIFF
--- a/frontend/app/model/Nav.scala
+++ b/frontend/app/model/Nav.scala
@@ -1,6 +1,7 @@
 package model
 
 import configuration.Links
+import controllers.routes
 
 object Nav {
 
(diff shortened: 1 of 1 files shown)


afiore:
:+1:  This was quite a tricky bug!

rtyley:
Verified: updated links (which are relative) work on https://membership.theguardian.com/ :

image

davidrapson:
Ack! Sorry about that, dangers of 1mo old branches…

PR #710, opened 2015-08-28
Display book/events feature choice when upgrading plans

afiore:
Allow book/events choice when upgrading membership trier

Also includes some basic refactoring to simplify method signatures in MemberService.


rtyley:
:+1: after a walk-through with @afiore

PR #712, opened 2015-09-01
Fix future supplier, get the fixed version from Membership common

rtyley:
See also https://github.com/guardian/membership-common/pull/88 - the `refresh()` method was not refreshing! Consequently, Zuora creds would expire after a couple of hours and then things would die.

cc @afiore @ostapneko


afiore:
:+1: The `FutureSupplier` implementation we had in membership looks identical the the one we have currently on membership-commons, so I don't think membership has ever suffered of this problem of auth creds expiring after a couple of hours. However, is good to avoid duplication and use the membership-common version :)

frontend/app/services/MemberService.scala
```diff
@@ -55,9 +51,10 @@ class FrontendMemberRepository(salesforceConfig: SalesforceConfig) extends Membe
     val stage = Config.stage
     val application = "Frontend"
 
-    val authTask = ScheduledTask("", Authentication("", ""), 0.seconds, 30.minutes)(getAuthentication)
+    override val authSupplier: FutureSupplier[Authentication] = new FutureSupplier[Authentication](getAuthentication)
 
-    def authentication: Authentication = authTask.get()
+    private val actorSystem = Akka.system
+    actorSystem.scheduler.schedule(30.minutes, 30.minutes) { authSupplier.refresh() }
```
rtyley: @vivekkr this is now where auth is periodically refreshed.

vivekkr: @rtyley Thanks Rob

Issue #717, opened 2015-09-03
Comparison table broken in Chrome 45


davidrapson:
Just raising an issue to get some visibility on this. Comparison table is currently broken on Chrome, but only on version 45 (!). Looks like it *might* be related to this crbug https://code.google.com/p/chromium/issues/detail?id=512872

**Chrome 43**
![chrome-43](https://cloud.githubusercontent.com/assets/123386/9656441/368e377c-5230-11e5-9353-4129badfbab4.png)

**Chrome 44**
![chrome-44](https://cloud.githubusercontent.com/assets/123386/9656443/3b5deda6-5230-11e5-8c1f-1e5c65758646.png)

**Chrome 45 - Broken**
![chrome-45-broken](https://cloud.githubusercontent.com/assets/123386/9656445/3e68ae14-5230-11e5-909d-636d3d100dc3.png)

**Chrome 47 - Fixed**
![chrome-47-canary-fixed](https://cloud.githubusercontent.com/assets/123386/9656600/524690bc-5231-11e5-8e98-66f74d4d4b94.png)

**Firefox**
![firefox](https://cloud.githubusercontent.com/assets/123386/9656529/df6da5ee-5230-11e5-8265-57f80eb811f4.png)

**Safari**
![safari-8](https://cloud.githubusercontent.com/assets/123386/9656532/e3539b5a-5230-11e5-8afe-22c814f692c3.png)

**IE9**
![ie9](https://cloud.githubusercontent.com/assets/123386/9656566/295d5d66-5231-11e5-9061-f4b18f4c3b02.png)

Baffling. Think the best solution is going to be to try and use some different markup to approach the new version of the comparison table and try and get that out. @rtyley @afiore

rtyley:
Utterly baffling. Fuck.

davidrapson:
Yah. Although, even thought this *does* look like a chrome bug the previous CSS was a bit gnarly and stuff like this *usually* suggests there's a simpler way. Refactored the new comparison table to use a different approach (i.e., not tabular data) which side-steps this issue https://github.com/guardian/membership-frontend/pull/723

The new version can go out on Monday as I've reduce it to just a design update and doesn't have the new boost-tier benefits in yet. Bug avoided.

davidrapson:
Also, based on our stats Chrome 45 is basically non-existent in our traffic as I *think* the windows update might lag by a version (or else auto-update takes a little while to filter through) so likely no-one apart from us will really see the broken layout.

PR #723, opened 2015-09-03
Comparison table updates

davidrapson:
This PR is a re-cut of https://github.com/guardian/membership-frontend/pull/700 with the new benefits removed and the comparison table structure refactored to avoid this bug: https://github.com/guardian/membership-frontend/issues/717. There's a few magic numbers in the CSS but it's largely no more complex than the previous approach.

screen shot 2015-09-03 at 14 52 40
screen shot 2015-09-03 at 14 52 30
screen shot 2015-09-03 at 14 52 48

@afiore


frontend/app/model/Benefits.scala
```diff
@@ -83,43 +92,36 @@ object Benefits {
...
-    """Show deep support for keeping the Guardian free, open and independent.
-    | Get invited to a small number of exclusive, behind-the-scenes functions""".stripMargin
+
+  val comparisonHiglightsList = benefitsFilter(
+    "priority_booking",
+    "discount",
+    "unique_experiences"
   )
 
   def details(tier: Tier) = tier match {
```
tudorraul: This looks to be related only to `views`, so it should not be a present in `model`

frontend/assets/stylesheets/components/_icons.scss
```diff
@@ -61,6 +61,12 @@
     width: 36px;
     height: 36px;
 }
+.icon-inline--badge {
+    width: 50px;
```
tudorraul: are `px`s automatically converted to `rem`s?

tudorraul:
:+1: bar the comments

frontend/app/model/Benefits.scala
davidrapson: Given it's so tightly related to this model, is it not preferable to live here?

frontend/assets/stylesheets/components/_icons.scss
davidrapson: Yeah, we have a task for it now https://github.com/guardian/membership-frontend/blob/master/frontend/Gruntfile.js#L71

PR #725, opened 2015-09-03
Rationalise layout utilities

davidrapson:
There is little to no conceptual difference between the `l-` prefixed layout helper classes and the layout related `u-` prefixed utility classes.

This PR rationalises these layout helpers to all be contained as a dedicated set of utilities. Easier to reason about and removes a small amount of duplication.

@chrisjowen


chrisjowen:
:+1:

PR #726, opened 2015-09-03
Add mixin for layout constraints and remove @extend

davidrapson:
I generally consider `@extend` in scss to be a bit of an anti-pattern, and [thankfully I'm not the only one](http://csswizardry.com/2014/11/when-to-use-extend-when-to-use-a-mixin/).

This PR removes the one use of `@extend` we have in the codebase by adding a mixin for layout constraints. It also has the benefit of being able to keep this logic close to the tightly coupled `side-margins-calc` logic.

@chrisjowen


chrisjowen:
:+1:

PR #727, opened 2015-09-03
Use subscription number

afiore:
Avoid using `SubscriptionId` in usage records. Use `SubscriptionNumber` instead.

**Note:** depends on #697

### DIFF
--- a/frontend/app/services/MemberService.scala
+++ b/frontend/app/services/MemberService.scala
@@ -139,10 +139,12 @@ trait MemberService extends LazyLogging with ActivityTracking {
     for {
       (account, subscription) <- tp.subscriptionService.accountWithLatestMembershipSubscription(member)
       description = s"event-id:${event.id};order-id:${order.id}"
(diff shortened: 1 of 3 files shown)


frontend/app/controllers/TierController.scala
```diff
@@ -38,8 +38,8 @@ trait DowngradeTier extends ActivityTracking {
     val futureTierName = "Friend"
     for {
       subscriptionStatus <- subscriptionService.getSubscriptionStatus(request.member)
-      currentSubscription <- subscriptionService.getSubscriptionDetails(subscriptionStatus.current)
-      futureSubscription <- subscriptionService.getSubscriptionDetails(subscriptionStatus.future.get)
+      currentSubscription <- subscriptionService.getSubscriptionDetails(subscriptionStatus.currentVersion)
+      futureSubscription <- subscriptionService.getSubscriptionDetails(subscriptionStatus.futureVersionOpt.get)
```
chrisjowen: Are we using xxxOpt (i.e. futureVersionOpt) Hungarian notation elsewhere to identify types. Why not just futureVersion?

afiore: This style choice is currently matter of debate..

PR #728, opened 2015-09-03
Remove CSP header

davidrapson:
We've been running a report-only Content-Security-Policy header on Membership for the last few weeks. Unfortunately it's pretty hard to meaningfully determine a safe-whitelist for the policy as the signal-to-noise ratio on reports is pretty low.

We could monitor this in a smarter way over time (i.e. not using Sentry) but at our current scale I don't think we get enough benefit out of CSP to be worth the extra effort. Sadly removing this for now.

@rtyley


chrisjowen:
:+1: for the steps to remove, but cannot comment on the actual decision to remove.

rtyley:
:+1: sad that CSP isn't useful yet

PR #729, opened 2015-09-04
Encode mailto links

davidrapson:
Fixes: https://trello.com/b/RmlFYzW5/membership

Mailto query parameters need to be url encoded and need to use `%20` for spaces instead of `+`

**Broken: Blank screen**
broken-mailto

**Fixed: Correct mailto link**
screen shot 2015-09-04 at 11 29 09

On the back of this I've also raised a [new trello card](https://trello.com/c/0SfTvvXt/79-consider-removing-social-links-from-events-detail-pages) to consider removing social links completely as from our stats very few people actually use them to justify them.

@chrisjowen

### DIFF
--- a/frontend/app/views/fragments/social.scala.html
+++ b/frontend/app/views/fragments/social.scala.html
@@ -6,7 +6,7 @@
     <li class="social-list__item">
         <a class="social-action social-action--email"
            target="_blank"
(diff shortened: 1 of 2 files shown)


chrisjowen:
:+1:

PR #731, opened 2015-09-04
Update FAQs

davidrapson:
This PR adds new FAQs for books & tickets. Is also refactors FAQs to remove the questions list. 

Particularly with the addition of the new FAQs the questions list became overlong. David M and I both consider the list of questions redundant so we've agreed to remove them.

screen shot 2015-09-04 at 12 44 28
screen shot 2015-09-04 at 12 44 31

@chrisjowen @afiore


TesterSpike:
Mistakes in the text above  
- "..selected 'XXX' as **you** Partner benefit..." should be **your**
- "If I choose the Tickets benefit, which **books** will I receive and when?" should have a different ending, e.g. ".., which events can I book and when?"

My impression was that if Book event was selected then they would receive a book in each quarter.  The current text gives me the impression that they will receive a book every 3 months after their join date.  Which should it be?

davidrapson:
@TesterSpike Good catch, these came straight from a copy doc so I didn't read closely enough. It's once a quarter so yeah you are right it does read differently.

davidrapson:
@TesterSpike Fixed the typo, going to follow up about the wording.

chrisjowen:
:+1:  though this is also a good candidate for https://trello.com/c/0otLHbCN/87-tech-debt-consider-splitting-out-view-model-information-from-benefits-scala.

PR #732, opened 2015-09-04
npm shrinkwrap

davidrapson:
Testing `npm shrinkwrap` to see if it improves builds.

- Locks recursive dependencies which improves stability
- Potential performance increases as we are more likely to be able to use the cache between builds due to having locked versions.

@rtyley @afiore

### DIFF
--- a/frontend/npm-shrinkwrap.json
+++ b/frontend/npm-shrinkwrap.json
@@ -1,0 +1,16628 @@
+{
+  "name": "membership",
+  "dependencies": {
+    "autoprefixer-core": {
+      "version": "5.2.1",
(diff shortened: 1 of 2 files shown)


davidrapson:
OMG no though :cry: my commit ratio.

screen shot 2015-09-04 at 12 58 55

rtyley:
:+1: 

commit ratio :bomb:

davidrapson:
Had a quick chat to Oliver Ash and the reason they use `shrinkwrap` is because they got bitten by a transitive dependency version, worth being aware of but not a problem we've had. In theory `skrinkwrap` should speed up installs due to being able to use the cache more but it's not something they've seen any practical improvements from. So closing this for now.

PR #734, opened 2015-09-04
Fix syntax error on promoTertiary action

davidrapson:
Fix syntax error on promoTertiary action. A QA id was being output as `id="'qa-learn-patron'"` when it should be `id="qa-learn-patron"`.

@TesterSpike


TesterSpike:
:+1:  I had to update the element selector because the existing css selector wasn't consistently picking up the element.

PR #735, opened 2015-09-04
Update CTA.js link for Become a Member to /choose-tier

davidrapson:
Remove some old references to `/join` which has now been removed and redirects to `/`, in this case the equivalent page should actually be `/choose-tier`.

@afiore

### DIFF
--- a/frontend/assets/javascripts/src/modules/events/cta.js
+++ b/frontend/assets/javascripts/src/modules/events/cta.js
@@ -9,7 +9,7 @@ define([
         join: {
             id: 'join',
             label: 'Become a member',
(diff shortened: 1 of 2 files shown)


chrisjowen:
Are there other places where the old /join url still exists, or any potential that users have this url bookmarked. Should we add a redirect filter to guard against this instead of redirecting to /?

davidrapson:
@chrisjowen Ah yeah, my description wasn't that clear. It is correct that `/join` should redirect to `/` (and there is a redirect rule in place for that. However, in this particular case just want them to go somewhere else.

PR #736, opened 2015-09-04
Handle tier downgrades correctly

afiore:
Once a paid member decides to downgrade, we still charge for the entire subscription term and record a subscription amendment that will make the downgrade effective only after the end of the current term. This PR ensures that partner and patron benefits are correctly recognised throughout the duration of their term. This involves a combination of REST and SOAP api calls, as the REST representation of a Zuora subscription currently provides no information about the currently valid rate plan.  

Trello: https://trello.com/c/lz2j3ZPH/78-handle-tier-downgrade-correctly
@tudorraul @chrisjowen

### DIFF
--- a/frontend/app/services/SubscriptionService.scala
+++ b/frontend/app/services/SubscriptionService.scala
@@ -5,6 +5,7 @@ import com.gu.membership.model._
 import com.gu.membership.salesforce.MemberId
 import com.gu.membership.salesforce.Tier.{Partner, Patron}
 import com.gu.membership.stripe.Stripe
+import com.gu.membership.util.Timing
 import com.typesafe.scalalogging.LazyLogging
 import forms.MemberForm.JoinForm
 import model.{FeatureChoice, MembershipSummary}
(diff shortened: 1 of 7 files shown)


afiore:
@TesterSpike

chrisjowen:
:+1: with tests to prove correct subscription version was retrived

frontend/app/services/SubscriptionService.scala
```diff
@@ -154,28 +155,51 @@ class SubscriptionService(val tierPlanRateIds: Map[ProductRatePlan, String],
...
+
+  def currentRatePlan(subscription: Rest.Subscription): Future[Option[Rest.RatePlan]] = subscription.ratePlans match {
+    case onlyRatePlan :: Nil => Future.successful(Some(onlyRatePlan))
+    case multipleRatePlans => Timing.record(zuoraSoapService.metrics, "currentRatePlan-for-multipleRatePlans") {
+      for {
+        subscriptionStatus <- getSubscriptionStatus(subscription)
+        currentRatePlan <- getRatePlan(subscriptionStatus.currentVersion)
+      } yield {
+        logger.debug(s"Current ratePlan: $currentRatePlan, subscriptionStatus: $subscriptionStatus, multipleRatePlans: $multipleRatePlans")
+        multipleRatePlans.find(_.productRatePlanId == currentRatePlan.productRatePlanId)
```
rtyley: It's worth noting that this is a bit of hack: we're matching on the ProductRatePlanId (ie the ids used in the Product Catalog) rather than the RatePlanId (the ids that relate to a specific instance of a rate plan used by a customer).

The SOAP and Rest Zuora APIs appeared to return different RatePlanIds, preventing us from matching, so we've resorted to ProductRatePlanId - which should produce the same results for our purposes.

davidrapson: "The SOAP and Rest Zuora APIs appeared to return different RatePlanIds" :scream:

rtyley: > "The SOAP and Rest Zuora APIs appeared to return different RatePlanIds"

Not all the time @davidrapson . Not all the time. Just sometimes. :scream_cat: 

There's probably a logic to it but we haven't had a chance to ask Zuora yet.

rtyley:
:+1: :shipit:

mario-galic:
:+1:

PR #737, opened 2015-09-07
Fix welcome sign out redirect

davidrapson:
A fix for https://trello.com/c/0zwp1cVG/66-bug-redirect-users-back-to-main-homepage-when-signing-out-from-welcome

When users are signed in they are redirected to `/welcome`. On clicking sign-out they  are kept on that page, it would make more sense for them to be redirected to the homepage `/`.

- Model profile links
- Extend sign-out link so that `/welcome` redirects to homepage on sign-out. Welcome is intended for signed in users

@afiore

### DIFF
--- a/frontend/app/views/welcome.scala.html
+++ b/frontend/app/views/welcome.scala.html
@@ -1,4 +1,4 @@
-@(pageImages: Seq[model.ResponsiveImageGroup])
+@(pageInfo: model.PageInfo, pageImages: Seq[model.ResponsiveImageGroup])
 
 @import configuration.Links
 @import views.support.Asset
@@ -16,7 +16,7 @@
     </div>
 }
 
-@main("Home Page") {
+@main(pageInfo.title, pageInfo=pageInfo) {
 
     <main role="main" class="l-constrained">
 
--- a/frontend/app/configuration/Links.scala
+++ b/frontend/app/configuration/Links.scala
@@ -1,5 +1,7 @@
 package configuration
 
+import controllers.routes
+
 object Links {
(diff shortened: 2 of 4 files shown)


frontend/app/views/welcome.scala.html
```diff
@@ -16,7 +16,7 @@ <h3 class="feature-panel__description">
     </div>
 }
 
-@main("Home Page") {
+@main(pageInfo.title, pageInfo=pageInfo) {
```
davidrapson: This is a bit of an oddity we have `title` in `PageInfo` but also pass it manually as a string in the template. Should probably try and remove this duplication at some point.

chrisjowen:
If I understand this correctly when a user is not signed in they should never see the /welcome page, so we should add redirect logic to the FrontPage.welcome action too.

chrisjowen:
FrontPage controller also has a bunch if view logic directly in the controller actions, and is also a good candidate for https://trello.com/c/0otLHbCN/87-tech-debt-consider-splitting-out-view-model-information-from-benefits-scala. But is beyond the scope of this PR.

davidrapson:
@chrisjowen Appreciate the point, but the main reason we don't do that is so we can keep that page fully cached and so stateless. If someone goes to `/welcome` there's nothing on the page that is sensitive/really contextual so no reason to have this page be uncached.

> FrontPage controller also has a bunch if view logic directly in the controller actions

Completely agree about this one, would like to extract this out into a Galleries collection as they don't need to be directly tier to a controller. Another PR for sure.

chrisjowen:
@davidrapson ok, makes sense. :+1:

PR #738, opened 2015-09-07
Add books and tickets benefits

davidrapson:
Adds new boost tier benefits to new comparison table.

screen shot 2015-09-07 at 11 44 43
screen shot 2015-09-07 at 11 44 53

@afiore


frontend/app/model/Benefits.scala
```diff
@@ -28,6 +28,8 @@ object Benefits {
     BenefitItem("email_updates", "Regular events email"),
     BenefitItem("offers_competitions", "Offers and competitions"),
     BenefitItem("priority_booking", "48hrs priority booking"),
+    BenefitItem("books_or_tickets", "6 tickets or 4 books", isNew = true),
```
afiore: If possible I would avoid hardcoding the number 6, and instead use `Config.zuoraFreeEventTicketsAllowance`

afiore:
:+1: bar the comment

chrisjowen:
Already approved, but maybe we should be putting this view related stuff into something other than the model namespace/folder. Creating a new story to investigate.

PR #739, opened 2015-09-07
Cleanup promo and heading helper styles

davidrapson:
This is a re-cut of https://github.com/guardian/membership-frontend/pull/724 but with a bit more thorough cleanup.

- Removes all unused promo* related fragments
- Merges the single use `h-promo` helper into promo components. (@chrisjowen: Relevant to our discussion last week, low level helpers only useful if they are used in a number of generic contexts. Here the might as well be part of the component)
- Similarly removes `h-aside` and merges into single component that uses it.

cc/ @chrisjowen


chrisjowen:
Assuming that the removed promo fragments were supposed to be removed then the removal of the utility classes and the new single component looks good. Though cant thumbs up as this branch has merge conflicts that need fixing first.

davidrapson:
@chrisjowen Yep, they were hanging around post-redesign so are all good to go. Just fixed merge conflict, so let's try again now.

PR #740, opened 2015-09-07
CSS structure cleanup

davidrapson:
This PR is a follow on from https://github.com/guardian/membership-frontend/pull/703  with an attempt to group related styles more clearly. Styles are now grouped as follows:

- **Base**: Global element level styles
- **Objects**: Low level object styles (more abstract that components), .e.g, icons, grid, links, list objects
-**Components**: The bulk of our styles, individual components
- **Scopes**: Dedicated styling contexts. Notable use is content areas (e.g., from Content API) where we can't style like a component using BEM classes and instead need to hook into element classes.
- **Utilities**: Low level helper classes, e.g., layout constraints colour helpers. Useful for any traits we want to apply on a case by case basis to element. We try and keep these to a minimum.

Notably there's no `layout/` directory any more. Most of this was just global header/footer which is a component and the rest is better contained in base styles or utilities. No `layout/` directory encourages us to not think in terms of pages and instead think in components and slices of a page.

This is broadly based on this article: http://csswizardry.com/2015/03/more-transparent-ui-code-with-namespaces/ and the ITCSS architecture.

itcsslabels

@afiore @chrisjowen


chrisjowen:
:+1:

PR #742, opened 2015-09-07
Move CSS bower_components

davidrapson:
Follows on from https://github.com/guardian/membership-frontend/pull/740

Having bower-components inside our general modules/components directory was bugging me as it causes a bit of naming/conceptual confusion as there's a merging of third-party libraries and our code component styles. Same word but different concepts (we could possibly use Modules for our
stuff to avoid this naming overlap).

This PR moves CSS bower components up into the top level and uses the default name of bower_components (the less deviation from the defaults the better imo).

Similar process to follow for JS, but CSS is lower risk so trying this first.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,9 +20,9 @@ tmp
 project/project
 project/target
 
+bower_components
 */public/javascripts
 */public/stylesheets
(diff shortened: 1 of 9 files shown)


chrisjowen:
:+1:

PR #743, opened 2015-09-07
Prefetch rate plans

tudorraul:
Rate plan ids will be inferred from Zuora at statup time and will be removed from configuration

Dependent on https://github.com/guardian/membership-common/pull/90


frontend/app/services/SubscriptionService.scala
```diff
@@ -102,79 +105,96 @@ object SubscriptionService {
...
-  def getSubscriptionStatus(memberId: MemberId): Future[SubscriptionStatus] = {
-    for {
-      subscriptions <- getSubscriptions(memberId)
 
-      if subscriptions.nonEmpty
+  import SubscriptionService._
+  val membershipProductType = "Membership"
+  val productRatePlanChargeModel = "FlatFee"
+
+  val tierPlanRateIds: ProductRatePlan => Future[String] = productRatePlan =>
```
afiore: Can we call this `tierRatePlanId`? it returns a single id, not a collection thereof. Also, is there any benefit in creating an anonymous function here instead of just a method using `def`?

tudorraul: rename under way. I started with `val` because I wanted to separate some initialisation from the body of the function

TesterSpike:
:+1:

frontend/app/services/SubscriptionService.scala
```diff
@@ -131,12 +133,29 @@ object SubscriptionService {
...
+  val membershipProductType = "Membership"
+  val productRatePlanChargeModel = "FlatFee"
+
+  def tierPlanRateId(productRatePlan: ProductRatePlan): Future[String] =
+    membershipProducts.map { products =>
+      val zuoraRatePlanId = for {
+        product <- products.find(_.`Tier__c`.contains(productRatePlan.salesforceTier))
+        zuoraRatePlan <- productRatePlan match {
+          case FriendTierPlan | StaffPlan => product.activeRatePlans.headOption
+          case paidTierPlan: PaidTierPlan =>
```
afiore: If this pattern match is exhaustive, you shouldn't need to supply a type hit there (`case paidTierPlan` should be enough).

frontend/app/services/SubscriptionService.scala
```diff
@@ -131,12 +133,29 @@ object SubscriptionService {
...
+class SubscriptionService(val zuoraSoapService: ZuoraSoapService,
                           val zuoraRestService: ZuoraRestService) extends AmendSubscription with LazyLogging {
 
   import SubscriptionService._
-  val MembershipProductType = "Membership"
+  val membershipProductType = "Membership"
+  val productRatePlanChargeModel = "FlatFee"
+
+  def tierPlanRateId(productRatePlan: ProductRatePlan): Future[String] =
+    membershipProducts.map { products =>
```
afiore: Why not extracting the core logic of this method into a pure function? This would make it much easier to test...
```scala
def tierRatePlanId(products: Seq[Rest.Product])(productRatePlan: ProductRatePlan): Option[String]
```

frontend/app/services/SubscriptionService.scala
```diff
@@ -131,12 +133,29 @@ object SubscriptionService {
   }
 }
 
-class SubscriptionService(val tierPlanRateIds: Map[ProductRatePlan, String],
-                          val zuoraSoapService: ZuoraSoapService,
+class SubscriptionService(val zuoraSoapService: ZuoraSoapService,
                           val zuoraRestService: ZuoraRestService) extends AmendSubscription with LazyLogging {
 
   import SubscriptionService._
-  val MembershipProductType = "Membership"
```
afiore: Constants should be [Capitalised](http://docs.scala-lang.org/style/naming-conventions.html)!

tudorraul: Two remarks. First, this a member of a class, not an object, hence it's not a constant it's a immutable value within the class. Second, only the first letter of constant names should be capitalised. All caps statics are so Java! ;)

frontend/app/services/SubscriptionService.scala
tudorraul: Your remark follows my line of thoughts, but for some reason the compiler identifies this value as the base class, not as a `PaidTierPlan`. It'd be a good exercise to investigate why this happens.

frontend/app/services/SubscriptionService.scala
tudorraul: Fair point

frontend/app/controllers/Healthcheck.scala
```diff
@@ -1,34 +1,70 @@
...
 
-case class Test(name: String, ok: () => Boolean)
+import scala.concurrent.Future
+
+case class TestCase(name: String, exec: () => Future[Boolean]) {
+  def failedTest() = Logger.warn(s"Test $name failed, health check will fail")
+
+  def ok() = {
+    val result = exec()
+    result.map(passed => if (!passed) failedTest())
```
afiore: aren't these two lines doing the same thing?

tudorraul: they are logging a failed test which can fail for one of two reasons: if the result of the test is false or the test has thrown an exception

frontend/app/controllers/Healthcheck.scala
```diff
@@ -1,34 +1,70 @@
...
     Cached(1) {
-      if (failedTests.nonEmpty) {
-        failedTests.foreach { test =>
-          Logger.warn(s"Test ${test.name} failed, health check will fail")
-        }
-        ServiceUnavailable("Service Unavailable")
-      } else {
-        Ok("OK")
-      }
+      updateHealthCheck()
```
afiore: Is there a good reason why we need to introduce an agent here? I would keep this code simple and side-effect free by just assuming that these health-check tests get executed whenever matching request comes in

frontend/app/controllers/Healthcheck.scala
```diff
@@ -1,34 +1,58 @@
...
-case class Test(name: String, ok: () => Boolean)
+import scala.concurrent.Await
+import scala.util.{Try, Failure, Success}
+
+case class Test(name: String, exec: () => Boolean) {
+  def failedTest() = Logger.warn(s"Test $name failed, health check will fail")
+
+  def ok() = {
+    val passed = exec()
+    if (!passed) Logger.warn(s"Test $name failed, health check will fail")
```
ostapneko: call `failedTest` instead or remove `failedTest`

frontend/app/controllers/Healthcheck.scala
```diff
@@ -1,34 +1,58 @@
...
 
 object Healthcheck extends Controller {
-  val zuoraService = TouchpointBackend.Normal.zuoraSoapService
+  import scala.concurrent.duration._
+
+  val zuoraSoapService = TouchpointBackend.Normal.zuoraSoapService
+  val subscriptionService = TouchpointBackend.Normal.subscriptionService
 
-  val tests = Seq(
+  val productRatePlanTiers = List(FriendTierPlan, StaffPlan,
```
ostapneko:
```scala
for {
  tier <- Seq(Supporter, Partner, Patron)
  annual <- Seq(true, false)
} yield PaidTierPlan(tier, annual)
```

tudorraul: Since we have more tiers that just theses 6, I think that we'll be adding to many operations that abstract away and that explicitly specifying them in a single list is more straightforward.

frontend/app/controllers/Healthcheck.scala
```diff
@@ -1,34 +1,56 @@
...
     Test("Events", () => GuardianLiveEventService.events.nonEmpty),
     Test("CloudWatch", () => CloudWatchHealth.hasPushedMetricSuccessfully),
-    Test("Zuora", () => zuoraService.lastPingTimeWithin(2.minutes))
-  )
-
-  def healthcheck() = Action { req =>
-    val failedTests = tests.filterNot(_.ok())
+    Test("ZuoraPing", () => ZuoraPing.ping)) ++
+    productRatePlanTiers.map(tier => Test(s"ZuoraCatalog: $tier", () =>
+      Try(Await.result(subscriptionService.tierRatePlanId(tier).map(_ => true), 10 millis)) match {
```
rtyley: :-1: to calling Await in the healthcheck :-(  I prefer that the healthcheck calls `.value` on Futures we need to check (which is non-blocking, just getting the current value out of the future), tho' it looks like we might need a bit of a refactor to do that here.

davidrapson:
Needs merge…

tudorraul:
@rtyley can you have a look?

davidrapson:
What's happening with this?

@rtyley @afiore @chrisjowen @vivekkr Any one else able to review this?

afiore:
We were waiting for @rtyley to have the last word on this, but I am happy to do it myself as it is long due...

davidrapson:
Yep, no need for gatekeepers unless we desperately need Roberto.

rtyley:
I am actually looking at it right now. I tried to run it this morning, couldn't get the healthcheck to pass locally.

rtyley:
...still can't get the healthcheck to pass locally. @tudorraul, could you verify that the healthcheck passes locally for you?

rtyley:
I get:

```
2015-09-15 16:12:26,452 [application-akka.actor.default-dispatcher-16] WARN  application - Test ZuoraProductRatePlans failed, health check will fail
```

rtyley:
Apart from the fact that the healthcheck isn't passing in Dev (which @tudorraul is looking at), :+1:  to this getting merged.

tudorraul:
Proof it works
1. the fact that we application can be deployed
2. joining a paid tier with a test user
screen shot 2015-09-16 at 13 04 23

PR #746, opened 2015-09-08
Added label for no free member tickets

chrisjowen:
The understanding from https://trello.com/c/Lh5JzZw7/64-add-a-no-free-members-ticket-warning-to-the-event-overview is that tickets without complimentary events should show this label


davidrapson:
:+1: But minor stylistic point, the other boolean properties are phrased as positives whereas this is negative. Could this be written as `hasFreeMembersTickets` and have the logic reversed?

chrisjowen:
Agreed, switched boolean logic.

chrisjowen:
image

PR #747, opened 2015-09-08
Renamed No Free Members Ticket to Member's ticket

chrisjowen:
The term free for member tickets has tax implications and thus needed to be removed despite it been back office code.


rtyley:
Verified updated wording on https://membership.theguardian.com/staff/event-overview :

image

The unlocked ticket classes should now also work for both:

* "Member's ticket at no extra cost" - zero-cost
* "Member's ticket at no extra cost + book" - priced at the cost of the book

PR #748, opened 2015-09-08
Implement Json endpoint for fetching the number of free event tickets

afiore:
This PR demotes `memberTierFeatures` to a private method, and refactors it so that it accepts a subscription instead of a member object. Additionally, it updates `getUsageCountWithinTerm` changing its return type from plain `Int` to `Option[Int]`. In this case, the intended semantics for `None` is that the usage is not entitled to free event tickets, while `Some` implies instead that the user has such benefit.

@rtyley

### DIFF
--- a/frontend/conf/routes
+++ b/frontend/conf/routes
@@ -47,7 +47,7 @@ GET            /staff/event-details               controllers.Staff.eventDetails
 # Subscription
 POST           /subscription/update-card          controllers.Subscription.updateCard
 OPTIONS        /subscription/update-card          controllers.Subscription.updateCardPreflight
-GET            /subscription/tickets-left         controllers.Subscription.ticketsLeft
+GET            /subscription/remaining-tickets    controllers.Subscription.remainingTickets
 
 # Events
 GET            /events                            controllers.Event.list
--- a/frontend/conf/routes
+++ b/frontend/conf/routes
@@ -47,6 +47,7 @@ GET            /staff/event-details               controllers.Staff.eventDetails
 # Subscription
 POST           /subscription/update-card          controllers.Subscription.updateCard
 OPTIONS        /subscription/update-card          controllers.Subscription.updateCardPreflight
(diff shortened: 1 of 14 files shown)


afiore:
@davidrapson I am waiting for @rtyley feedback on this, than I would hand this branch over to you for implementing the front-end part

frontend/conf/routes
```diff
@@ -47,6 +47,7 @@ GET            /staff/event-details               controllers.Staff.eventDetails
 # Subscription
 POST           /subscription/update-card          controllers.Subscription.updateCard
 OPTIONS        /subscription/update-card          controllers.Subscription.updateCardPreflight
+GET            /subscription/tickets-left         controllers.Subscription.ticketsLeft
```
davidrapson: [URL design nerd warning]

This reads weirdly, maybe `ticket-allocation` or `remaining-tickets`

afiore: Warning accepted...

davidrapson:
Added a few commits to actually show the usage on the front-end:

screen shot 2015-09-09 at 15 03 30

Not 100% ready as we would like to show a different message for first-time users.

rtyley:
LGTM, though as discussed, we'll want to keep an eye on it, make sure it's not overloading Zuora :-)

davidrapson:
Just added a couple of updates, so as long as they look OK we are ready to go:

- Updated logic to only show banner on Guardian Live events detail pages
- Updated messages so we show contextual messages depending on the state of the user
- Moved the banner markup into a fragment

**New user on listing page**

new-listing

**New user on event detail**
new-event

**Returning user on listing page**
returning-listing

**Returning user on event detail**
returning-event

@afiore @rtyley

afiore:
I had a go at running it locally and it seems fine. Ready to merge unless anyone as something against that...

davidrapson:
Go for it :shipit:

afiore:
Verified on prod

screen shot 2015-09-10 at 16 03 22

davidrapson:
:gem:

PR #749, opened 2015-09-09
Shed 30s from box startup time

tudorraul:
Zuora ping has an initial delay that also delays box startup


tudorraul:
@rtyley

rtyley:
:+1: This 30 second delay was probably came from when we didn't block for auth.

PR #751, opened 2015-09-10
Avoid calling Await in the healthcheck

rtyley:
https://github.com/guardian/membership-frontend/pull/743/files#r39142397

In general, we don't want any blocking going on in the healthcheck.

cc @tudorraul


frontend/app/services/SubscriptionService.scala
```diff
@@ -140,13 +140,22 @@ class SubscriptionService(val zuoraSoapService: ZuoraSoapService,
...
+
   def membershipProducts = zuoraRestService.productCatalogSupplier.get().map(_.productsOfType(membershipProductType))
 
-  def tierRatePlanId(productRatePlan: ProductRatePlan): Future[String] =
-    membershipProducts.map(extractRatePlanIdFromCatalog(productRatePlan))
+  val productRatePlanIdSupplier = new FutureSupplier[Map[ProductRatePlan, String]](
+    for {
+      membershipProductCatalog <- membershipProducts
+    } yield (for {
+      productPlan <- productRatePlanTiers
```
tudorraul: I think using the `for` comprehension here makes code less readable and more verbose than using `map`

tudorraul:
:+1: bar the comment

PR #753, opened 2015-09-10
Design Consistency: Control links

davidrapson:
Improves design consistency of control links in header footer; both now use the same markup pattern and match behaviour of those on theguardian.com

header-controls mov

@tudorraul @afiore


tudorraul:
:+1:

PR #754, opened 2015-09-10
Don't call /subscription/remaining-tickets if < Partner

davidrapson:
Minor tweak to avoid calling `/subscription/remaining-tickets` if our tier is < Partner. We weren't showing the banner before, but this just ensures we don't even try to make the call in the first place.

@afiore


afiore:
:+1:

PR #755, opened 2015-09-10
Design Consistency: Global navigation

davidrapson:
- Extract primary and secondary navigation into separate fragments
- Move secondary navigation outside header (hidden, but broken in redesign)
- Improve BEM/naming-conventions for primary navigation markup
- Small bit of design refinement on small screens

**Before**
screen shot 2015-09-10 at 17 23 52

**After**
screen shot 2015-09-10 at 17 23 41

@tudorraul @afiore


afiore:
:+1:

PR #756, opened 2015-09-10
Design Consistency: Layout / global fixes

davidrapson:
- z-index bug on welcome slideshow overlaying mobile navigation
- Remove default colour from link, already scoped to text styles in `.copy` class and only needed there by default. Resolves a number of design consistency issues.

**Nav Before**
screen shot 2015-09-10 at 17 45 50

**Nav After**
screen shot 2015-09-10 at 17 44 52


afiore:
:+1:

PR #757, opened 2015-09-11
Design Consistency: Listings

davidrapson:
This PR does a bit of consistency work to ensure all listings have similar key-lines on the titles.  I've also used the clean up as an opportunity to tidy up the offers and competitions listing and merge it into the Info controller.

screen shot 2015-09-11 at 12 36 44
screen shot 2015-09-11 at 12 36 40
screen shot 2015-09-11 at 12 36 52

@afiore


frontend/app/controllers/Info.scala
```diff
@@ -152,6 +152,14 @@ trait Info extends Controller {
     Ok(views.html.info.patron(pageInfo, pageImages))
   }
 
+  def offersAndCompetitions = CachedAction { implicit request =>
+    val results = GuardianContentService.offersAndCompetitionsContent.map(ContentItemOffer)
+      .filter(_.content.fields.map(_("membershipAccess")).isEmpty)
+      .filter(!_.content.webTitle.startsWith("EXPIRED"))
+      .filter(_.imgOpt.nonEmpty)
```
afiore: Suboptimal code! By calling `filter` three times you are actually traversing the sequence three times. Replace with:

```scala

    val results = 
      GuardianContentService.offersAndCompetitionsContent.map(ContentItemOffer)
         .filter(offer => offer.content.fields.map(_("membershipAccess")).isEmpty &&
                          offer.content.webTitle.startsWith("EXPIRED") &&
                          offer.imgOpt.nonEmpty)

Also, why not move this into a method of `GuardianContentService` and keep the controller a bit more dry?
```

davidrapson: Yeah, totally agree. Wasn't my intended scope (just moving the template), but no reason I can't leave the campsite a little better than I found it. Happy to clean this up as part of this PR.

afiore:
:+1: bar the comment..

rtyley:
@davidrapson LGTM, anything blocking deploy?

PR #761, opened 2015-09-14
Analytics: Add custom dimensions for recording users state

davidrapson:
Add custom dimensions for recording users state

- Record if user is signed in
- Record if user is a Member

The plan is to monitor this data and if these dimensions are recorded correctly we'll look at also recording the actual tier of Membership (Supporter/Partner etc.)

These dimensions should give us some insight as to how existing members and signed in users are using the site.

See: https://support.google.com/analytics/answer/2709828

@dominickendrick @tudorraul

### DIFF
--- a/frontend/assets/javascripts/src/modules/analytics/ga.js
+++ b/frontend/assets/javascripts/src/modules/analytics/ga.js
@@ -1,6 +1,11 @@
 /* global ga: true */
-define(function() {
+define(['src/utils/user'], function(user) {
     'use strict';
+
+    var dimensions = {
+        signedIn: 'dimension1',
+        member: 'dimension2'
+    };
 
     function init() {
         /*eslint-disable */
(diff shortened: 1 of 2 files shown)


frontend/assets/javascripts/src/modules/analytics/ga.js
```diff
@@ -24,7 +31,15 @@ define(function() {
          */
         ga('require', 'linkid', 'linkid.js');
 
-        ga('send', 'pageview');
+        if(isLoggedIn) {
+            ga('set', dimensions.signedIn, !!isLoggedIn);
```
chrisjowen: No reason to coherse to a boolean if already one.

chrisjowen:
var dimensions = {
        signedIn: 'dimension1',
        member: 'dimension2'
    };

Did we actually setup GA with dimensions named dimension1 and dimension2? These seem very non-descriptive. Are these correct?

davidrapson:
@chrisjowen Yep… you get 20 keys to use `dimension1… dimension20`. We can give them descriptive names in the dashboard itself, but to log to them we have to use the explicit key name of `dimension1`.

PR #762, opened 2015-09-14
Past events prep: Event template refactor

davidrapson:
Bit of prep for the past events listing. Tidies up the naming conventions for our events views.

- guardianLive -> eventsList (not just Live events)
- masterclass -> masterclassesList (more explicit name)
- page -> eventDetail (remove ambiguity)

@chrisjowen @afiore


frontend/app/controllers/Event.scala
```diff
@@ -111,74 +110,72 @@ trait Event extends Controller with ActivityTracking {
...
+    Ok(views.html.event.eventDetail(pageInfo, event))
   }
 
   private def chronologicalSort(events: Seq[model.RichEvent.RichEvent]) = {
     events.sortWith(_.event.start < _.event.start)
   }
-
-  def list = CachedAction { implicit request =>
+  
+  def eventsList = CachedAction { implicit request =>
```
afiore: personally I think that this method was better named before... `Event.list` reads better than `Events.eventList`

davidrapson: Yeah agreed, maybe I only need to rename the template not the controller method.

afiore:
:+1:

PR #765, opened 2015-09-14
Improve Identity API client error handling

tudorraul:
HTTP transport and deserialisation issues are masked, this fixes that.


davidrapson:
:+1:

PR #766, opened 2015-09-15
Upgrade zxcvbn to v3.5

davidrapson:
- Updates to explicit version number now that `zxcvbn` has stabilised a bit with version 3.5. No longer using git hash(!)
- Allows us to use zxcvbn as an AMD module rather than having to load as a global.
- Add a few robustness checks discovered in related form modules whilst testing password strength as standalone pattern.

password mov

@rtyley @tudorraul


rtyley:
:+1:

PR #767, opened 2015-09-15
Add enhanced facebook events for phd campaign

davidrapson:
Adds extra Facebook events for new campaign starting on Wednesday.

@tudorraul @chrisjowen

### DIFF
--- a/frontend/assets/javascripts/src/modules/analytics/facebook.js
+++ b/frontend/assets/javascripts/src/modules/analytics/facebook.js
@@ -1,48 +1,81 @@
-define(function() {
+/* global fbq, _fbq */
+/* eslint-disable no-unused-vars, no-underscore-dangle */
+define(['lodash/collection/forEach'], function(forEach) {
     'use strict';
 
-    function usSupporterLandingPage() {
+    var FACEBOOK_PROPERTY_ID = '869473446445427';
+
+    var URL_MAPPINGS = {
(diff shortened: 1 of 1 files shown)


chrisjowen:
Don't understand whats happening here, if no one else can approve then happy to go through it with you so you can explain this to me. Does look random though with various magic numbers and minified code.

davidrapson:
Yep happy to chat through it.

davidrapson:
Happy to chat through it.

frontend/assets/javascripts/src/modules/analytics/facebook.js
```diff
@@ -1,48 +1,81 @@
...
+        '/join/friend/thankyou': 'CompleteRegistration',
+        '/join/supporter/thankyou': 'CompleteRegistration',
+        '/join/partner/thankyou': 'CompleteRegistration',
+        '/join/patron/thankyou': 'CompleteRegistration'
+    };
+
+    function callByPathname(mappings, callback) {
+        var pathname = window.location.pathname;
+        forEach(mappings, function(val, url) {
+            if(url === pathname && callback) {
```
tudorraul: Perhaps the callback check can be done one, outside the loop. This may lead to a shorter code that uses `indexOf` on `EVENT_MAPPINGS` to determine if the current path is eligible.

tudorraul:
code LGTM :+1: bar the comment

PR #768, opened 2015-09-16
Warn on staff event-overview if ticket sales end before event start

rtyley:
After discussion Sophie Kinsella (Assistant events manager on Membership), we agreed that it would be helpful to make sure events haven't accidentally been created with ticket sales ending substantially before the event start.

https://membership.theguardian.com/staff/event-overview

image

I've set the warning label to show if ticket sales end earlier than 2 hours before the event. I also had to exclude 'Guestlists' as these seem to often end before the event starts.

https://trello.com/c/9SatPfOJ/110-warn-when-ticket-sales-end-long-before-event-start

cc @chrisjowen

### DIFF
--- a/frontend/app/model/Eventbrite.scala
+++ b/frontend/app/model/Eventbrite.scala
@@ -5,14 +5,15 @@ import com.gu.membership.salesforce.Tier
 import com.netaporter.uri.Uri
 import com.netaporter.uri.dsl._
 import configuration.Config
-import org.joda.time.Instant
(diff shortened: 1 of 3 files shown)


rtyley:
Tweaked to give slightly clearer periods ("Ticket sales end 10 days before event start: ..."):

image

davidrapson:
:+1: 

(there's probably a bit of work do be done at a later date to make the layout a bit more space efficient)

rtyley:
Verified, looks good on PROD:

image

PR #770, opened 2015-09-17
Filter Error: duplicate define: jquery

davidrapson:
We get quite a few errors on Membership for `Error: duplicate define: jquery` (see https://app.getsentry.com/the-guardian/membership/group/31343105/). We don't use jQuery on the site but I *think* this is a side-effect of jQuery being AMD compatible and so if it's loaded by a user/browser-extension you might get this warning. 

This PR is a test of using Raven's `ignoreErrors` to filter this out.

- Extract raven config into module
- Add filter for Error: duplicate define: jquery

@tudorraul

### DIFF
--- a/frontend/assets/javascripts/src/main.js
+++ b/frontend/assets/javascripts/src/main.js
@@ -1,5 +1,6 @@
 require([
     'ajax',
+    'src/modules/raven',
     'src/modules/analytics/setup',
     'src/modules/welcome',
     'src/modules/slideshow',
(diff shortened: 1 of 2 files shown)


frontend/assets/javascripts/src/main.js
```diff
@@ -20,11 +21,10 @@ require([
     'src/modules/identityPopup',
     'src/modules/identityPopupDetails',
     'src/modules/metrics',
-    'src/modules/patterns',
-    // Add new dependencies ABOVE this
-    'raven'
```
tudorraul: what was this pointing to initially and do we need to remove that?

davidrapson: This was pointing to Raven which is the sentry client, it's an implied global variable, hence the awkward comment. It's not moved to `frontend/assets/javascripts/src/modules/raven.js` below.

tudorraul:
:+1:

PR #772, opened 2015-09-18
Adjust signed in reporting

davidrapson:
After testing this, we need a couple of tweaks to the signed-in member reporting to make this useful.

- GA dimensions take a string, without `toString()` the state just shows as `1`.
- We need to log the false state too, so this rejigs things to log signed-out / non-member too.

Currently looks like this:

screen shot 2015-09-18 at 11 12 09

But we want it to look like this (from Subscriptions):

screen shot 2015-09-18 at 11 15 57

@tudorraul @chrisjowen


davidrapson:
@tudorraul Mind having a quick look at this today?

tudorraul:
:+1:

PR #774, opened 2015-09-18
What's On: Event location filters

davidrapson:
This PR adds location filters to the events listing page, and also does some work to improve the information density of the listing page.

- Adds location filter
- Removes prices from listing
- Slim card style on small screens

filters mov

events-fitlers-large

Biggest difference is on small screens

**Before**
screen shot 2015-09-21 at 16 39 23

**After**
screen shot 2015-09-21 at 16 38 55

5 events + filters vs. 1.5 events in a screenful

@tudorraul @chrisjowen


dominickendrick:
:+1:

PR #775, opened 2015-09-21
What's on prep: Remove tagged events list

davidrapson:
This is some prep for https://github.com/guardian/membership-frontend/pull/774.

It turns out we have a tagged list page for `/events` (in addition to Masterclasses). However it only matches on title, so it's weird…

See this amazing page with just events with "Burt" in the title:

screen shot 2015-09-21 at 11 31 30

We don't link to this anywhere and as we're revisiting how we do filtering on events pages now is probably the time to clean it out, as I don't think it's an expected feature.

### DIFF
--- a/frontend/app/services/EventbriteService.scala
+++ b/frontend/app/services/EventbriteService.scala
@@ -54,7 +54,7 @@ trait EventbriteService extends WebServiceHelper[EBObject, EBError] {
 
   def mkRichEvent(event: EBEvent): Future[RichEvent]
   def getFeaturedEvents: Seq[RichEvent]
-  def getTaggedEvents(tag: String): Seq[RichEvent]
+  def getTaggedEvents(tag: String): Seq[RichEvent] = Seq.empty
   def getEventsArchive: Option[Seq[RichEvent]] = Some(eventsArchive)
   def getPartnerEvents: Seq[RichEvent] = events.filter(_.providerOpt.isDefined)
 
@@ -131,7 +131,6 @@ object GuardianLiveEventService extends LiveService {
     yield GuLiveEvent(event, gridImageOpt, contentApiService.content(event.id))
 
   override def getFeaturedEvents: Seq[RichEvent] = EventbriteServiceHelpers.getFeaturedEvents(eventsOrderingTask.get(), events)
-  override def getTaggedEvents(tag: String): Seq[RichEvent] = events.filter(_.name.text.toLowerCase.contains(tag))
   override def start() {
     super.start()
     eventsOrderingTask.start()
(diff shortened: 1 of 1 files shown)


frontend/app/services/EventbriteService.scala
```diff
@@ -103,6 +103,14 @@ abstract class LiveService extends EventbriteService {
   val gridService = GridService(Config.gridConfig.url)
   val contentApiService = GuardianContentService
 
+  def mkRichEvent(event: EBEvent): Future[RichEvent] = for {
```
davidrapson: It was bugging me that we had the same `mkRichEvent` code (apart from the type) duplicated in the Live and Local services, so this is an attempt to move it up into the shared service. I'd appreciate some :eyes: on this as I just guessed my way to this and I have no idea if it's idiomatic or not. The pattern matching is non-exhaustive too so idk, it compiled though…

@ostapneko @chrisjowen

frontend/app/services/EventbriteService.scala
```diff
@@ -54,7 +54,7 @@ trait EventbriteService extends WebServiceHelper[EBObject, EBError] {
   def mkRichEvent(event: EBEvent): Future[RichEvent]
   def getFeaturedEvents: Seq[RichEvent]
   def getEvents: Seq[RichEvent]
-  def getTaggedEvents(tag: String): Seq[RichEvent]
+  def getTaggedEvents(tag: String): Seq[RichEvent] = Seq.empty
```
davidrapson: I'd *like* to be able to define this just on MasterclassesService, but I couldn't work out how to make that actually work…

frontend/app/services/EventbriteService.scala
```diff
@@ -103,6 +103,14 @@ abstract class LiveService extends EventbriteService {
   val gridService = GridService(Config.gridConfig.url)
   val contentApiService = GuardianContentService
 
+  def mkRichEvent(event: EBEvent): Future[RichEvent] = for {
+    gridImageOpt <- gridImageFor(event)
+    props = (event, gridImageOpt, contentApiService.content(event.id))
+  } yield this match {
+    case GuardianLiveEventService => GuLiveEvent.tupled(props)
+    case LocalEventService => LocalEvent.tupled(props)
```
rtyley: Probably a bad idea, sadly... you're trying to work out what subclass you're in, better to just let them live in the subclasses.

davidrapson:
@rtyley Look OK aside from the comment?

rtyley:
I built this as a quick and dirty way of demonstrating that filtering was a possible thing! Fine to see it go... :+1:

PR #776, opened 2015-09-21
Remove unused event item fragment and styles

davidrapson:
Remove unused event item fragment and styles, deprecated as part of the recent past events listing work but wasn't removed at the time.

@tudorraul


tudorraul:
:scissors: :+1:

PR #777, opened 2015-09-21
Slide, rather than fade status bar

davidrapson:
Tweak to the free tickets status bar. Previously we faded it in, but now we slide it in:

slide mov

This avoids the current issue with having to *reserve* the space for it when fading in to avoid a jump.

**Before**
screen shot 2015-09-21 at 17 12 18

**After**
screen shot 2015-09-21 at 17 12 22

@tudorraul


tudorraul:
:+1:

PR #779, opened 2015-09-23
Event description, truncate to first sentence if possible

davidrapson:
**Before**
screen shot 2015-09-23 at 12 10 13

**After**

screen shot 2015-09-23 at 12 45 26

@joelochlann

### DIFF
--- a/frontend/app/model/Eventbrite.scala
+++ b/frontend/app/model/Eventbrite.scala
@@ -44,7 +44,7 @@ object Eventbrite {
       mcPattern.replaceAllIn(clean, "")
     }
 
-    lazy val blurb = truncateToWordBoundary(text, 120)
+    lazy val blurb = firstSentenceOrTruncate(text, 120)
   }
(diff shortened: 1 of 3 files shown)


rtyley:
Seems fine, though if you wanted to be pernickety, it's worth noting that a one-sentence description will end up having a period at the end of it, while a multi-sentence description will end up _not_ having a period at the end of it.

davidrapson:
Meh… this makes it worse in practice. Going to close this for now.

screen shot 2015-09-25 at 13 49 02

rtyley:
I've just learned about [`BreakIterator.html.getSentenceInstance()`](http://docs.oracle.com/javase/7/docs/api/java/text/BreakIterator.html#getSentenceInstance()) which might actually be useful to get the effect you were after... :bread: 

http://docs.oracle.com/javase/7/docs/api/java/text/BreakIterator.html#sentence

PR #780, opened 2015-09-24
Fix errors on patterns page

davidrapson:
Wrap all event patterns with `headOption`. This should stop the errors caused due to the events service being async, so `head` might fail.

screen shot 2015-09-24 at 11 27 46

Thanks to @joelochlann for prompting me to sort this.

### DIFF
--- a/frontend/app/views/patterns/patterns.scala.html
+++ b/frontend/app/views/patterns/patterns.scala.html
@@ -277,7 +277,9 @@
     }
 
     @pattern("Series Detail") {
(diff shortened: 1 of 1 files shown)


davidrapson:
@joelochlann Can you give this a once-over when you've got a second?

It's worth first reading our [pull-request recommendations doc](https://github.com/guardian/recommendations/blob/master/pull-requests.md) to get an idea of how we work.

Once you've done that give this a :+1: and we can run through the deployment process if you've not been through it already.

joelochlann:
:+1:

PR #784, opened 2015-09-28
Centralise all event overview templates in a single directory

davidrapson:
This does a bit of cleanup work to centralise all event overview fragments into a single place rather than spreading them around the views directory.


dominickendrick:
:+1:

PR #786, opened 2015-09-29
Try to reduce some noise in raven errors

davidrapson:
We generally get quite a bit of noise in our JS error tracking (e.g., https://app.getsentry.com/the-guardian/membership/group/65170122/) trying to reduce this a bit. Tested with omniture validator tool as it touches that code.

@joelochlann @rtyley


joelochlann:
:+1:

PR #789, opened 2015-10-01
No special treatment for programming partner events in listing

davidrapson:
As discussed in sprint planning, we're moving partner events into the main listing as there's no business need to keep them separate and is a simpler approach.

**Before (separate container)**
screen shot 2015-10-01 at 11 30 39

**After (programming partner events in main list)**
screen shot 2015-10-01 at 11 30 30

@joelochlann


joelochlann:
:+1:

PR #792, opened 2015-10-05
Try and get grid image for masterclasses event

davidrapson:
Currently some Masterclasses events don't have images, this is because there is one article linking to multiple eventbrite links. We use eventbrite as the source of truth whereas on theguardian.com the article is the source of truth so there's a mismatch of approaches here which causes conflicts.

As the Masterclasses team already use the comment builder for tagging we might as well avoid the need to have eventbrite ids in CAPI (blergh) and just allow them to provide a link to a grid image in the same way we do with live/local.

This PR attempts to use a grid image url or otherwise fallback to the CAPI lookup.

**Current**
screen shot 2015-10-05 at 17 42 23

**Dev events using grid image comment**
screen shot 2015-10-05 at 17 36 25

**Default masterclasses**
screen shot 2015-10-05 at 17 45 15

@joelochlann @rtyley

### DIFF
--- a/frontend/app/model/RichEvent.scala
+++ b/frontend/app/model/RichEvent.scala
@@ -118,12 +118,13 @@ object RichEvent {
 
   case class MasterclassEvent(
     event: EBEvent,
-    data: Option[MasterclassData]
+    data: Option[MasterclassData],
+    image: Option[GridImage]
   ) extends RichEvent {
     val detailsUrl = routes.Event.details(event.slug).url
     val hasLargeImage = false
     val canHavePriorityBooking = false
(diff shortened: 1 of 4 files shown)


joelochlann:
:+1:

rtyley:
One comment I would make is that Composer already has the facility to add EventBrite Ids to articles. If the Masterclasses team were to tag their articles with EBIds, associattion could be made that way.

davidrapson:
@rtyley Probably worth us having a chat about this. It does, but the problem we are seeing is that relationship is not quite what we'd expect.

davidrapson:
As discussed going to close this for now.

PR #795, opened 2015-10-07
Log the user's identity ID to raven / sentry

tomverran:
Also update the eslint no-console rule to only emit warnings for local development


rtyley:
:+1:

davidrapson:
:+1: let's ship this

tomverran:
Confirmed on PROD

PR #796, opened 2015-10-07
Capture stripe token creation validation errors with Raven

tomverran:
During our time on this @davidrapson and I noted that we duplicate the checks Stripe does prior to allowing the form to be submitted so the chances of this code being fired seem remote but there would be benefit to knowing if this is happening.

### DIFF
--- a/frontend/assets/javascripts/src/modules/raven.js
+++ b/frontend/assets/javascripts/src/modules/raven.js
@@ -15,15 +15,20 @@ define(['src/utils/user', 'raven'], function (user) {
          * Set up Raven, which speaks to Sentry to track errors
          */
         Raven.config(dsn, {
-            whitelistUrls: [ /membership\.theguardian\.com/ ],
+            whitelistUrls: [ /membership\.theguardian\.com/, /mem\.thegulocal\.com/, /localhost/ ],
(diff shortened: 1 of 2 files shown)


rtyley:
:+1:

tomverran:
looks okay to me on prod insofar as raven reports correctly

frontend/assets/javascripts/src/modules/raven.js
```diff
@@ -15,15 +15,20 @@ define(['src/utils/user', 'raven'], function (user) {
...
          */
         Raven.config(dsn, {
-            whitelistUrls: [ /membership\.theguardian\.com/ ],
+            whitelistUrls: [ /membership\.theguardian\.com/, /mem\.thegulocal\.com/, /localhost/ ],
             tags: tags,
             ignoreErrors: [ /duplicate define: jquery/ ],
-            ignoreUrls: [ /platform\.twitter\.com/ ]
+            ignoreUrls: [ /platform\.twitter\.com/ ],
+            shouldSendCallback: function(data) {
+                if(window.guardian.isDev) {
```
craigjbass: What.
What
What
What
What

PR #797, opened 2015-10-08
Fix links in client-side-principles

davidrapson:
Github auto-links URLs in text, plus this Markdown syntax doesn't work anyway… 

@joelochlann


joelochlann:
:+1:

PR #799, opened 2015-10-12
Re add optimizely

davidrapson:
Planning to run another A/B test, so re-adding optimizely for a bit. @joelochlann


frontend/app/views/fragments/event/eventCollection.scala.html
```diff
@@ -1,9 +0,0 @@
-@(events: Seq[model.RichEvent.RichEvent], limit: Int = 4)
```
dominickendrick: I assume this was removed cos it is no longer used ?

dominickendrick:
:+1:

frontend/app/views/fragments/event/eventCollection.scala.html
davidrapson: Yep, just being lazy and piggybacking a bit.

PR #801, opened 2015-10-12
Check that the members API is consistent with tier info in the app

ostapneko:
Increment a Cloudwatch metric every time `/user/me` is accessed in order to monitor the API consistency with the data already accessible from membership frontend.

In case of failure, it won't produce an error, but will log the mismatch, and also increment the appropriate Cloudwatch counter.

@rtyley @chrisjowen @afiore


rtyley:
:+1:

ostapneko:
Errorring at the moment on Staff members, I need to investigate on what to d with those ...

ostapneko:
image

PR #802, opened 2015-10-13
Fix bug getting the latest invoice items

tomverran:
It is possible to make amendments through the Zuora web UI which change the ID of the subscription, resulting in a subscription ID with no associated invoices.

As such this code alters the query to fetch the payment summary for a given member to instead query on subscription name (number) and then filters the items to return only those associated with the most recent invoice.

### DIFF
--- a/frontend/app/services/SubscriptionService.scala
+++ b/frontend/app/services/SubscriptionService.scala
@@ -53,16 +53,15 @@ object SubscriptionService {
   }
 
   /**
-   * Given an array of subscription invoice items
-   * return only items corresponding to the last invoice
-   * @param items The incoming list of items with many invoices
+   * Given an array of subscription invoice items, return only items corresponding to the last invoice
+   * @param items The incoming list of items with many invoices, potentially associated to multiple subscription versions
    */
   def latestInvoiceItems(items: Seq[InvoiceItem]): Seq[InvoiceItem] = {
     if(items.isEmpty)
       items
     else {
-      val temp = items.sortBy(_.serviceStartDate)
-      temp.filter(_.id == temp.last.id)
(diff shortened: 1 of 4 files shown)


frontend/app/services/SubscriptionService.scala
```diff
@@ -53,6 +53,20 @@ object SubscriptionService {
...
   /**
+   * Given an array of subscription invoice items
+   * return only items corresponding to the last invoice
+   * @param items The incoming list of items with many invoices
+   */
+  def latestInvoiceItems(items: Seq[InvoiceItem]): Seq[InvoiceItem] = {
+    if(items.isEmpty)
+      items
+    else {
+      val temp = items.sortBy(_.serviceStartDate)
```
dominickendrick: This may read slightly better ?
```
      val recentId = items.sortBy(_.serviceStartDate).last.id
      items.filter(_.id == recentId)
```

frontend/app/services/SubscriptionService.scala
```diff
@@ -53,6 +53,20 @@ object SubscriptionService {
...
+      temp.filter(_.id == temp.last.id)
+    }
```
afiore: this should be `_.subscriptionId == temp.last.subscriptionId`

chrisjowen:
Build is broken, needs pushing with correct membership-common version. Happy to plus one when this is done

afiore:
It should be ready to go now

chrisjowen:
:+1:

afiore:
Verified on prod using an account for which we were previously encountering the null-pointer exceptions and profile page is displaying as expected

screen shot 2015-10-15 at 12 07 59

PR #804, opened 2015-10-15
Add patron column

joelochlann:
This PR adds a patron column to the comparison table on the homepage and to the package promos above it. Along with this, it makes wide-reaching changes to the layout of the package promo component which affect every page where that component appears. Broadly these are:

- Mobile view will change to include full details, everywhere this component appears in the site
- Desktop view will be stacked vertically (as per design) on the homepage only. Elsewhere we will leave the column view in place.

I've also fixed the an issue with the 3-column promo layout. On tablet, previously these tried to stay as columns despite the fact that there wasn't enough space for their contents. Now they change as follows:

- Mobile = stacked (previously without details, now with details)
- Tablet = stacked with details
- Desktop = three columns

Table before:
picture 10

Table after:
picture 12

Package promos before:
picture 9

Package promos after:
picture 11

Mobile before:
picture 13

Mobile after:
picture 14
picture 15

Broken tablet view before:
picture 17

Fixed tablet view:
picture 18

@davidrapson


davidrapson:
> I've also fixed the an issue with the 3-column promo layout. On tablet, previously these tried to stay as columns despite the fact that there wasn't enough space for their contents. Now they change as follows:

Is it maybe worth pulling out that fix into a separate PR?

frontend/app/views/fragments/tier/comparisonTable.scala.html
```diff
@@ -30,30 +33,28 @@
...
         <div class="comparison-table__checkmark__inner">
         @fragments.common.checkmark(checkmark)
         </div>
     </div>
 }
 
-@comparisonPricing(tier: Tier) = {
-    <div class="comparison-table__trail">
+@comparisonPricing(tier: Tier, isActive: Boolean = false) = {
+    <div class="comparison-table__trail comparison-table--clickable@if(isActive){ is-active}" data-tier="@tier.name.toLowerCase">
```
davidrapson: Bit of BEM weirdness here, you've got a parent modifier `comparison-table--clickable` on a child `comparison-table__trail` is there another way of phrasing this selector?

frontend/assets/images/inline-svgs/raw/cross.svg
```diff
@@ -1 +1,8 @@
...
\ No newline at end of file
+<?xml version="1.0" encoding="utf-8"?>
```
davidrapson: Event though these are minified buy the build step it's probably still worth running these through https://jakearchibald.github.io/svgomg/ just to clean out the cruft and maybe remove `fill="#BDBDBD"` just to be on the safe side.

frontend/assets/images/inline-svgs/raw/tick.svg
```diff
@@ -1 +1,7 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="36" height="28.719" viewBox="0 0 36 28.719"><path fill="#fff" d="M36 1.77l-1.771-1.77-24.86 21.883-7.599-7.599-1.77 1.771 9.124 12.664 3.54-3.541-.005-.004z"/></svg>
\ No newline at end of file
+<?xml version="1.0" encoding="utf-8"?>
```
davidrapson: Ditto above

frontend/assets/javascripts/src/modules/comparisonTable.js
```diff
@@ -0,0 +1,65 @@
...
+    var ACTIVE_CLASS = 'is-active';
+    var COMPARISON_TABLE = document.querySelector('.comparison-table');
+
+    function init() {
+        addHoverListeners();
+        addClickListeners();
+    }
+
+    function addClassToTier(clazz, tier) {
+        var tierElems = COMPARISON_TABLE.querySelectorAll('.comparison-table--clickable[data-tier="'+tier+'"]');
```
davidrapson: It's probably OK to use `$` here which is a wrapper for qwery and [bonzo](https://github.com/ded/bonzo) so you can just do a one-liner $('.selector').addClass('my-class') and it will add a class to a set of elements.

davidrapson: Also, use a [`js-` prefixed class](https://github.com/guardian/membership-frontend/blob/master/docs/client-side-principles.md#javascript) here please

frontend/assets/javascripts/src/modules/comparisonTable.js
```diff
@@ -0,0 +1,65 @@
...
+            });
+        });
+    }
+
+    function addClickListeners() {
+        bean.on(COMPARISON_TABLE, 'click', '.comparison-table--clickable', function(e) {
+            var tier = $(e.currentTarget).data('tier');
+            addClassToTier(ACTIVE_CLASS, tier);
+
+            var otherElems = COMPARISON_TABLE.querySelectorAll('.comparison-table--clickable:not([data-tier="'+tier+'"])');
```
davidrapson: General comment throughout the file to use a [`js-` prefixed class](https://github.com/guardian/membership-frontend/blob/master/docs/client-side-principles.md#javascript) for all selectors.

frontend/assets/javascripts/src/modules/comparisonTable.js
```diff
@@ -0,0 +1,65 @@
...
+            for (var i = 0; i < otherElems.length; i++) {
+                otherElems[i].classList.remove(ACTIVE_CLASS);
```
davidrapson: Again, can probably use `$` if you are needing to loop through elements.

frontend/assets/javascripts/src/modules/comparisonTable.js
```diff
@@ -0,0 +1,65 @@
...
+
+    function addClassToTier(clazz, tier) {
+        var tierElems = COMPARISON_TABLE.querySelectorAll('.comparison-table--clickable[data-tier="'+tier+'"]');
+        for (var i = 0; i < tierElems.length; i++) {
+            tierElems[i].classList.add(clazz);
+        }
+    }
+
+    function addHoverListeners() {
+        $('.comparison-table--clickable').each(function(elem) {
```
davidrapson: I might be misreading what this does, but can we not do this with a `:hover` CSS selector and no JS?

frontend/assets/stylesheets/components/_action.scss
```diff
@@ -201,6 +206,30 @@
     }
 }
 
+.action--snap-tablet {
```
davidrapson: Doesn't look like this is used anymore?

frontend/assets/stylesheets/components/_comparison-table.scss
```diff
@@ -12,6 +19,47 @@ $_table-gutter: $gs-gutter / 2;
...
+    &:not(.is-active) {
+        .comparison-table__checkmark__inner,
+        .comparison-table__tier__name,
+        .comparison-table__tier__pricing,
+        .comparison-table__trail__inner {
+            opacity: 0.9;
+        }
+    }
+
+    &.is-hover {
```
davidrapson: Can we not do this just with `:hover`?

frontend/assets/stylesheets/components/_packages.scss
```diff
@@ -7,149 +7,178 @@
...
+
+@mixin package-promo-base() {
     border-top: 5px solid;
     background-color: $c-neutral7;
-}
+    padding: rem($gs-gutter / 4);
+
     .package-promo__header {
-        padding: rem($gs-gutter / 4);
+        margin-bottom: rem($gs-gutter / 2);
```
davidrapson: You can remove the `rem()` function call here as they are no longer needed (we still have a few files with them in if you fancy doing some cleanup).

frontend/assets/stylesheets/objects/_grid.scss
```diff
@@ -113,10 +124,55 @@
     }
 }
 
-/* 3up Stacked (3 > 1)
+/* 3up Stacked (Cols per row: 3 on larger viewport > 1 on smaller)
    ========================================================================== */
 
+// 3 on desktop > 1 on tablet > 1 on mobile
 .grid--3up-stacked {
+    &.grid--equal-heights {
```
davidrapson: Does this need to be a chained class? Can is just be an extra modifier we add as `grid--equal-heights`. If it *does* need to be chained can we reduce some duplication either by combining the classes or using a mixin, whichever makes more sense for you.

frontend/assets/stylesheets/objects/_grid.scss
```diff
@@ -113,10 +124,55 @@
...
+        @include mq(desktop) {
+            display: flex;
+        }
+        .grid__item {
+            @include mq(desktop) {
+                display: flex;
```
davidrapson: Couple of things about the flexbox use:

1. Mind posting a couple of screenshots from non-flexbox browsers
2. Probably worth a check in Firefox and Safari as flexbox is occasionally buggy using this shorthand version.
3. It might be worth wrapping this code in a check so it's excluded from our old-IE stylesheet:

```
@if ($browser-supports-flexbox) {
            // flex styles
        } @else {
            // non-flex styles
        }
```

frontend/assets/stylesheets/objects/_grid.scss
```diff
@@ -113,10 +124,55 @@
...
+            margin-bottom: rem($gs-baseline);
+            width: 33.3% - ($gutter-width-fluid / 2);
+            &:nth-of-type(3n+1) { clear: left; border-left: none; }
+        }
+    }
+}
+
+
+// 3 on desktop > 3 on tablet > 1 on mobile
+// TODO: use mixin and just pass different breakpoint as a variable, rather than duplicating
```
davidrapson: TODO?

davidrapson:
:+1: bar the comments.

Main questions are just a) how the flexbox usage degrades and b) whether we can streamline the JS a bit.

frontend/assets/stylesheets/components/_packages.scss
joelochlann: I'll do a separate PR for removing `rem()` universally

frontend/assets/stylesheets/objects/_grid.scss
joelochlann: Yes it would be much better as a standalone modifier. But I did it like this because the breakpoints vary depending on which type of grid class you are using. So a generic one would assume the wrong breakpoint if used in combination with certain grid classes, e.g. if `grid--equal-heights` assumes column view for desktop only, but is used in combination with `grid--3up-stacked-mobile`

I guess a mixin where we could pass in the breakpoint as a variable would solve this. Does that make sense to you or is there some other solution I've overlooked?

davidrapson: Gotcha, a mixin sounds good to me.

davidrapson:
Need a merge

joelochlann:
**Tested flexbox in Safari:**
picture 21

joelochlann:
**Tested flexbox in Firefox:**

picture 22

PR #806, opened 2015-10-15
Missing message on home page

chrisjowen:
Additional text was required on homepage that was not clear in the story


afiore:
:+1:

PR #808, opened 2015-10-15
Removed 3 months free and savings information

chrisjowen:
Several changes based upon this trello story:
https://trello.com/c/vmaUsXmn/186-price-rise-messaging-ahead-of-the-actual-price-rise-on-monday-2nd-november

The story is almost unreadable but these final set of changes are simply around changing the background image and also anywhere that mentioned the savings for patron and partner now says "3 months free" or nothing.


afiore:
Can we avoid making the logic in the `Benefit` case class more complex and maybe add some pattern matching directly in the view? The pricing is about to become dynamic anyway and as a result the Benefits class will change substantially. So I would be more keen on handling this copy overrides directly in the view

chrisjowen:
@afiore This is debatable, the view model should only provide the data that is needed for the views it supports. For this case we would have to match on Tier which in order to determine which ones to remove the savings messages from. 

Theres two reasons why I don't want to do this;

1) there is already existing matching on Tier in the benefits model to change what we add to the view.  So if we go your approach then we would either have to move all the matching over to the view or duplicate efforts and do the same thing in two different places

2) The view doesn't care about the tier, it simply presents the pricing information from the view model and is agnostic to whether this is about a tier or some other pricing benefit.

afiore:
All right, well, considering that this is supposed to go out rather soon, we can defer the discussion to later. :+1:

PR #809, opened 2015-10-16
Roundel covering menu now resolved

chrisjowen:
Changed the z-index of the top menu allowing it to overlap the roundal on the homepage


rtyley:
:+1:

PR #810, opened 2015-10-16
Add myself to alumni

davidrapson:
And with it I give you some boring facts about my tenure:

- I'm responsible for 54% of PRs on Membership [1]. Technically makes me majority stakeholder of the project…
- 2886 lines away from my goal of reaching a negative commit ratio on Membership. This is my greatest failure. *(Frantically tries to find 3k lines of code to delete)*
- No #2 committer, but still miles away from @feedmypixel's record…

screen shot 2015-10-16 at 12 22 20

Finally some exiting[2] commit ratio stats. Turns out I write a lot of CSS, and particularly I love deleting CSS. More scarily I wrote enough Scala to be considered dangerous, surprisingly almost as much Scala as JS.
```
Changes for pattern ".scala$"
+4156, -3040, 1116 total (1.37:1 +/- ratio)

Changes for pattern ".scala.html$"
+11366, -10227, 1139 total (1.11:1 +/- ratio)

Changes for pattern ".scss$"
+19215, -20711, -1496 total (0.93:1 +/- ratio)

Changes for pattern ".js$"
+4693, -4355, 338 total (1.08:1 +/- ratio)
```

*[1] Technically cheating as this is only since we went public*
*[2] May not actually be exiting*

Over and out.

### DIFF
--- a/frontend/public/humans.txt
+++ b/frontend/public/humans.txt
@@ -5,11 +5,6 @@ Twitter: @rtyley
 From: Blyth, England
 Location: London, England
 
-Client-side Developer: David Rapson
-Twitter: @davidrapson
-From: Birmingham, England
-Location: London, England
-
 /* ALUMNI */
 
 Senior Developer: Chris Finch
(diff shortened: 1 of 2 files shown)


feedmypixel:
@davidrapson its the taking part that counts, we are after all winners :) and looking at the graphs what a beautiful hand over of the baton :+1:

PR #812, opened 2015-10-16
Recompile all assets on watched change

joelochlann:
Currently we have separate watch tasks for compileCss and compileJs, with the aim of making the watch recompile a bit quicker (e.g. it will just recompile CSS if you've only modified Sass files). But this means the `public` directory is not completely cleaned out before recompiling, meaning you can end up in a halfway state with both dev & production assets in `public`. So I've changed it to do a complete compile when watch detects changes.

@davidrapson


davidrapson:
:100: :+1:

PR #817, opened 2015-10-19
Improve cta visibility on comparison table

joelochlann:
- On desktop, the 'Join Now' button appears on hover
- On mobile & table the arrow button always appears

**Old mobile:**
picture 24

**New mobile:**
picture 23


joelochlann:
I've reverted "Show 'Join Now' button on hover" because it seems to be mired in controversy, but am pushing out the other fix (show all buttons on mobile & tablet) without a :+1: as this is a clear defect fix which should go out ASAP

PR #818, opened 2015-10-20
Redirect to the identity registration with campaign code

chrisjowen:
Identity need to track better there referers so we now need to pass through a INTCMP query string parameter that will show where they came from and which tier they chose i.e. MEM_{PAGE}_{TIER}. 

See this card for all the permutations:
- https://trello.com/c/ZUTlcWIT/174-enable-identity-to-track-referrals-from-membership

### DIFF
--- a/frontend/app/actions/RegistrationUri.scala
+++ b/frontend/app/actions/RegistrationUri.scala
@@ -1,0 +1,40 @@
+package actions
+
+import com.netaporter.uri.Uri
+import configuration.Config
+import play.api.mvc.RequestHeader
+
+object RegistrationUri {
+  val REFERRER_HEADER: String = "referer"
+  val CAMPAIGN_SOURCE: String = "MEM"
+
+  def parse(request: RequestHeader) = {
+    val redirectUrl: String = Config.idWebAppRegisterUrl(request.uri)
+    extractCampaignCode(request.headers(REFERRER_HEADER), request.path) match {
+      case Some(campaignCode) => redirectUrl.concat(s"&INTCMP=${campaignCode.get}")
+      case _ => redirectUrl
+    }
+  }
+
+  private def extractCampaignCode(referer: String, path: String) = {
+    val refererUrl = Uri.parse(referer)
+
+    val campaignReferer = refererUrl.path match {
+      case "/supporter" => "SUPUK"
+      case "/us/supporter" => "SUPUS"
+      case "/offers-competitions" => "COMP"
+      case "/events" => "EVT"
+      case _ => "HOME"
+    }
+    val campaignTier = path match {
+      case "/join/supporter/enter-details" => Some("SUP")
+      case "/join/partner/enter-details" => Some("PAR")
+      case "/join/patron/enter-details" => Some("PAT")
+      case "/join/friend/enter-details" => Some("FRI")
+      case _ => None
+    }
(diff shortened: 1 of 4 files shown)


frontend/app/actions/RegistrationUri.scala
```diff
@@ -0,0 +1,40 @@
...
+    extractCampaignCode(request.headers(REFERRER_HEADER), request.path) match {
+      case Some(campaignCode) => redirectUrl.concat(s"&INTCMP=${campaignCode.get}")
+      case _ => redirectUrl
+    }
+  }
+
+  private def extractCampaignCode(referer: String, path: String) = {
+    val refererUrl = Uri.parse(referer)
+
+    val campaignReferer = refererUrl.path match {
```
joelochlann: Is it possible to get these strings programatically using Play's routing mechanism? That would make things easier if URLs change in future - hopefully these raw paths would only be stored in the `routes` file

frontend/test/actions/RegistrationUriTest.scala
```diff
@@ -0,0 +1,55 @@
...
+      assertCodeFor("MEM_SUPUK_FRI", "/join/friend/enter-details", "https://membership.theguardian.com/supporter")
+    }
+
+  }
+
+
+  private def assertCodeFor(code: String, path: String, referrer: String) =  {
+    val request = new RequestBuilder().path(path).header("referer", referrer).build()._underlyingHeader()
+    val regUri: String = RegistrationUri.parse(request)
+    println(regUri)
```
joelochlann: Should this still be here? If it's to provide debug information when running the tests should it give a bit more info? e.g. `println(s"Registration URI: ${regUri}")` or something like that

joelochlann:
:+1:

frontend/app/actions/RegistrationUri.scala
```diff
@@ -0,0 +1,51 @@
...
+import configuration.Config
+import play.api.mvc.RequestHeader
+
+object RegistrationUri {
+  val REFERRER_HEADER: String = "referer"
+  val CAMPAIGN_SOURCE: String = "MEM"
+
+  def parse(request: RequestHeader) = {
+    val redirectUrl: String = Config.idWebAppRegisterUrl(request.uri)
+    extractCampaignCode(request.headers(REFERRER_HEADER), request.path) match {
```
rtyley: This has caused a lot of errors on the site, see:

https://app.getsentry.com/the-guardian/membership/group/90393824
https://app.getsentry.com/the-guardian/membership/group/90393829

frontend/app/actions/RegistrationUri.scala
```diff
@@ -0,0 +1,51 @@
...
+      case Some(campaignCode) => redirectUrl.concat(s"&INTCMP=${campaignCode.get}")
+      case _ => redirectUrl
+    }
+  }
+
+  private def extractCampaignCode(referer: String, path: String) = {
```
rtyley: IntelliJ tells me that the return type of this method is `Option[Some[String]]` - that's indicative of poor Scala code.

In general, you don't want to nest `Option`s - you want to flatten them into a single `Option` with flatMap, or a for-comprehension. Also, `Some` being in the actual type signature indicates that it's redundant - the return type should just be `Option[String]`

PR #819, opened 2015-10-20
Exclude staff from analytics

afiore:
Urgent. This is needed in order to fix an ongoing deployment to prod

### DIFF
--- a/frontend/app/configuration/Config.scala
+++ b/frontend/app/configuration/Config.scala
@@ -74,12 +74,12 @@ object Config {
   val googleAnalyticsTrackingId = config.getString("google.analytics.tracking.id")
 
   val facebookJoinerConversionTrackingId =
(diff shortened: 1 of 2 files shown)


tomverran:
:+1:

frontend/app/configuration/Config.scala
```diff
@@ -74,12 +74,12 @@ object Config {
   val googleAnalyticsTrackingId = config.getString("google.analytics.tracking.id")
 
   val facebookJoinerConversionTrackingId =
-    Tier.all.map { tier => tier -> config.getString(s"facebook.joiner.conversion.${tier.slug}") }.toMap
+    Tier.allPublic.map { tier => tier -> config.getString(s"facebook.joiner.conversion.${tier.slug}") }.toMap
```
tomverran: I think we should deprecate generating configuration keys programatically in the future
being able to reliably find usages of configuration keys is convenient

PR #820, opened 2015-10-20
Re-instate .aws/config setting in README

joelochlann:
Removed in PR https://github.com/guardian/membership-frontend/pull/678 but it appears it is used by `awscli`. And sometimes `awscli` complains about the `region` setting being inside `credentials` so we shouldn't be putting it in there.

@rtyley


rtyley:
Looks to match what we see in http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html#cli-config-files :+1:

PR #821, opened 2015-10-20
Fix regression in Staff discounted tickets

afiore:
Since the recent update to membership-commons `0.92`, Staff members are not anymore assigned the `Partner` tier, but a more specific `Staff` tier. This change fixes two recent regression which 1) prevent staff members from accessing events discounts, 2) accidentally include a staff tier in the comparison table.


tomverran:
:+1:

afiore:
Verified both regressions on prod and the fix has been effective

PR #822, opened 2015-10-21
Update js to display event price to staff members

afiore:
This should make sure that discounted event prices are also shown to Staff members


tomverran:
:+1:  as a quick fix
However we should pull a list of tiers eligible for discounts from the backend, otherwise this will happen again in the future

afiore:
@tomverran I have added a new property to the `/user/me` endpoint called `discountTicketTiers` which maps to the relevant constant in the code. While I feel that this specific addition is a bit adhoc, it should at least help keeping things in sync between the frontend and the backend.

tomverran:
much better! :+1:

PR #823, opened 2015-10-21
Avoid hardcoding tier whitelist

afiore:
Avoid listing existing tiers in the `user/me` endpoint. At the same time, prevent potential out-of-sync issues by driving the front-end logic through boolean values produced by the backend.

@rtyley @tomverran

### DIFF
--- a/frontend/app/controllers/User.scala
+++ b/frontend/app/controllers/User.scala
@@ -80,9 +80,10 @@ trait User extends Controller {
     "tier" -> member.tier.name,
     "isPaidTier" -> member.tier.isPaid,
     "joinDate" -> member.joinDate,
-    "benefits" -> Json.obj {
-      "discountedEventTickets" -> Benefits.DiscountTicketTiers.contains(member.tier)
-    }
+    "benefits" -> Json.obj(
+      "discountedEventTickets" -> Benefits.DiscountTicketTiers.contains(member.tier),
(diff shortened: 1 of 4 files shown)


rtyley:
:+1: :+1: 

This is much better than putting static information into the `user/me` endpoint!

Not a blocker for merge, but I would say that it's a shame that we're encoding this information in two ways in the code, ie:

https://github.com/guardian/membership-frontend/blob/090c41b/frontend/app/model/Benefits.scala#L9-L10

...duplicates (swapping axis) some of the data in:

https://github.com/guardian/membership-frontend/blob/090c41b/frontend/app/model/Benefits.scala#L92
https://github.com/guardian/membership-frontend/blob/090c41b/frontend/app/model/Benefits.scala#L104

It would be cool, at some point in the future, to be able to get rid of the `DiscountTicketTiers` & `ComplimenataryTicketTiers` vals by driving that information from benefits listed against each tier.

afiore:
Cool, I think I will go ahead with merging this now. I like the idea of evolving the Benefits class into some sort of access control facility. So far this class has been just as a bag of copy strings that we display in our views based on the tier. I have created an new ticket on [Trello](https://trello.com/c/6BXNfbld/210-refactor-access-control-and-permissions-logic-in-membership-frontend) to do this, hopefully we can get it done sooner rather than later.

PR #826, opened 2015-10-22
Rewire controllers/views to use dynamic pricing data

afiore:
- Extend Play actions to embed pricing data into the request object.
- Simplify some of the pricing view code, moving it into implicit class `views.support.Prices.Copy`.
- Thread dynamic pricing information through views and view fragments.

### DIFF
--- a/frontend/app/model/Benefits.scala
+++ b/frontend/app/model/Benefits.scala
@@ -15,30 +15,6 @@ case class Benefits(tier: Tier, pricing: Option[Pricing]) {
     case Staff => staffBenefitsList
   }
 
-  val cta = s"Become a ${tier.name.toLowerCase}"
-
-  val leadin = tier match {
-    case Partner => "Supporter benefits, plus…"
-    case Patron => " Partner benefits, plus…"
-    case _ => "Benefits"
-  }
-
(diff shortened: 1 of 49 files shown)


frontend/app/model/Benefits.scala
```diff
@@ -1,75 +1,56 @@
...
 import com.gu.membership.salesforce.Tier
-import com.gu.membership.salesforce.Tier.{Partner, Patron, Staff}
+import com.gu.membership.salesforce.Tier._
 import configuration.Config.zuoraFreeEventTicketsAllowance
 
-object Benefits {
-
-  val DiscountTicketTiers = Set[Tier](Staff, Partner, Patron)
-  val ComplimenataryTicketTiers = Set[Tier](Partner, Patron)
+case class Benefits(tier: Tier, pricing: Option[Pricing]) {
```
rtyley: We should probably separate pricing away from Benefits, because are benefits shouldn't vary whether the user is a test-user or not.

rtyley:
:+1:

afiore:
Verified on prod and seems to work as expected
screen shot 2015-10-23 at 15 43 39

PR #827, opened 2015-10-23
Quick hack to hide events with bad EventBrite settings

rtyley:
...that can't be fixed without creating new EventBrite events.

https://trello.com/c/Tj6mJ5Hh/212-national-theatre-event-hack


afiore:
:+1: all right, maybe it would be good to know when we can remove it!

PR #829, opened 2015-10-26
Event fees message

chrisjowen:
Now that we are passing on event fees to non members we need to display the correct prices from eventbrite. The non member price should include the fee and show how much the fee is and the partner/patron price should include the deduced fees in there discount price. 

image

image

### DIFF
--- a/frontend/app/model/Eventbrite.scala
+++ b/frontend/app/model/Eventbrite.scala
@@ -95,6 +95,11 @@ object Eventbrite {
 
   case class EBPricing(value: Int) extends EBObject {
     lazy val formattedPrice = formatPriceWithCurrency(value)
+    def +(other: Option[EBPricing]) : Option[EBPricing] = {
+      other.map( cost =>
+	EBPricing(value+cost.value)
+      )
+    }
   }
 
   case class EventTimes(created: Instant, start: DateTime)
@@ -109,6 +114,7 @@ object Eventbrite {
                            quantity_sold: Int,
                            on_sale_status: Option[String], // Currently undocumented, so treating as optional. "SOLD_OUT", "AVAILABLE", "UNAVAILABLE"
                            cost: Option[EBPricing],
+			   fee: Option[EBPricing],
                            sales_end: Instant,
                            sales_start: Option[Instant],
                            hidden: Option[Boolean]) extends EBObject {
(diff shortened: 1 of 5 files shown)


frontend/app/model/Eventbrite.scala
```diff
@@ -86,6 +86,11 @@ object Eventbrite {
 
   case class EBPricing(value: Int) extends EBObject {
     lazy val formattedPrice = formatPriceWithCurrency(value)
+    def +(other: Option[EBPricing]) : Option[EBPricing] = {
```
dominickendrick: Could this be called `addFee` rather than an using an operator overload.

frontend/app/model/Eventbrite.scala
```diff
@@ -86,6 +86,11 @@ object Eventbrite {
 
   case class EBPricing(value: Int) extends EBObject {
     lazy val formattedPrice = formatPriceWithCurrency(value)
+    def addFee(other: Option[EBPricing]) : Option[EBPricing] = {
```
afiore: Why on earth would I want to add an "optional" fee? Wouldn't be better something like?

`addFee(that: EBPricing): EBPricing`

frontend/app/model/Eventbrite.scala
```diff
@@ -179,7 +190,7 @@ object Eventbrite {
   }
 
   case class DiscountBenefitTicketing(generalRelease: EBTicketClass, member: EBTicketClass) {
-    val saving = generalRelease.priceInPence - member.priceInPence
+    val saving = (generalRelease.priceInPence + generalRelease.feeInPence) - member.priceInPence
```
afiore: No need for rounded braces... `+` has higher priority of `-`

afiore:
:+1: overall fine, except for comments...

frontend/app/model/Eventbrite.scala
chrisjowen: I'd prefer to leave it for clarity

joelochlann: I agree with you there @chrisjowen, relying on operator precedence makes things harder to read in my opinion

chrisjowen: Ok 2v1 Ill change it

PR #830, opened 2015-10-26
Refactor eventbrite

chrisjowen:
Separated the eventbrite service responsibilities into the client and the caching of events and renamed appropriatly


chrisjowen:
@rtyley did you ever take a look at this, if you disagree with spliting file by responsibility then can I ask why?

PR #835, opened 2015-10-28
Added guard code for potentially

chrisjowen:
Referer URL missing if the user directly hits a page that requires registration directly from the url bar. Guard code added, also crazy nested option removed.

### DIFF
--- a/frontend/app/actions/RegistrationUri.scala
+++ b/frontend/app/actions/RegistrationUri.scala
@@ -9,49 +9,38 @@ object RegistrationUri {
   val REFERRER_HEADER: String = "referer"
   val CAMPAIGN_SOURCE: String = "MEM"
 
+  val referingPageCodes = Map(
+    controllers.routes.Info.supporter().toString() -> "SUPUK",
+    controllers.routes.Info.supporterUSA().toString() -> "SUPUS",
+    controllers.routes.Info.offersAndCompetitions().toString()-> "COMP",
+    controllers.routes.WhatsOn.list().toString() -> "EVT",
+    "/" -> "HOME"
+  )
+
+  val campaignTierCodes = Map[String, String](
+    controllers.routes.Joiner.joinPaid(Tier.Supporter).toString() -> "SUP",
+    controllers.routes.Joiner.joinPaid(Tier.Partner).toString() -> "PAR",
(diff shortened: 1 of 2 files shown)


frontend/app/actions/RegistrationUri.scala
```diff
@@ -9,43 +9,38 @@ object RegistrationUri {
...
-      case `supporterTier` => Some("SUP")
-      case `partnerTier` => Some("PAR")
-      case `patronTier` => Some("PAT")
-      case `friendTier` => Some("FRI")
-      case _ => None
-    }
-    campaignTier.map {
-      tier => Some(Seq(CAMPAIGN_SOURCE, campaignReferer, tier).mkString("_"))
-    }
+    referingPageCodes.get(refererUrl.path)
```
ostapneko: While we are at it, I would guard the `Uri.parse` as well with like `Try(Uri.parse).toOption`

frontend/app/actions/RegistrationUri.scala
```diff
@@ -9,43 +9,38 @@ object RegistrationUri {
...
+    redirectUrl.concat(s"&INTCMP=$campaignCode")
   }
 
-  private def extractCampaignCode(referer: String, path: String) = {
+  private def extractCampaignCode(refererOpt: Option[String], path: String) : String = {
+
+    val referingCode = refererOpt.flatMap(getReferingPageCode)
+    val campaignTier: Option[String] = campaignTierCodes.get(path)
+
+    (Seq(CAMPAIGN_SOURCE) ++ referingCode ++ campaignTier).mkString("_")
```
ostapneko: Does it rely on implicit conversion from `Option` to `Seq`?

If so, then we can easily have an inconsistent code without a referingCode.

What about a simple `Option`-based for-comprehension instead?

chrisjowen: Regarding the fail case, yes thats exactly what we want to happen at least they want to know the source is Membership. Nope no Option to Seq implicit conversion going on here, internal magic of IndexedSeqLike pimipery. What advantage would I get form for comprehension here?

ostapneko: Well if we are happy to send an inconsistent code, clearly there is no need for for-comprehension

ostapneko:
:+1:

chrisjowen:
Tested on production, redirects witout error without referer

PR #839, opened 2015-10-29
Javacript disabled causing error logging noise

chrisjowen:
When javascript is disabled on submitting the join as a friend form the user is given a 500 error. The fix is to stop the submitting of the form and to display a link to the "enable javascript" help page the same as the header banner shows. 

An example of the error logging is : https://app.getsentry.com/the-guardian/membership/group/90708377/events/3831377884/

A visual of the fix is:

image

### DIFF
--- a/frontend/app/views/fragments/form/submit.scala.html
+++ b/frontend/app/views/fragments/form/submit.scala.html
@@ -11,7 +11,6 @@
 
 <fieldset class="fieldset fieldset--simple">
     <div class="fieldset__heading"></div>
-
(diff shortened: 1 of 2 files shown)


rtyley:
:+1: that's great

rtyley:
Verified on PROD, button looks good with javascript disabled:

Friend (I could still join as a friend, if I had JS enabled):
image

Partner:
image

...odd black dot after the button, but a slgihtly off appearance is not a problem, given that we're already in JS-off land.

PR #840, opened 2015-10-29
Better stripe exception handling

tomverran:


ostapneko:
Wouldn't the `warn` level be more appropriate? Also, it's probably worth squashing, this is a lot of commits for a tiny change :)

ostapneko:
:+1:

tomverran:
There's now fewer logs on sentry but its still logging the failed API call once
so not entirely fixed, but a step in the right direction

PR #841, opened 2015-10-29
Update mem-common version to 0.94

ostapneko:
Add more logging whenever a Zuora call is made. Note that 0.94 does *not* include changes made in 0.93 but is an update of 0.92

@rtyley @chrisjowen @tomverran @joelochlann


chrisjowen:
:+1 when the build passes

PR #843, opened 2015-11-02
Update website copy to reflect new pricing

afiore:
@chrisjowen mind having a look? also, should we keep the CSS code for the roundel even if the roundel itself is gone?

- [Trello card](https://trello.com/c/8dHeLK3a/39-annual-price-increases-for-supporter-partner-and-patron-tiers-target-mid-october-to-plan-in-the-testing)

### DIFF
--- a/frontend/app/views/index.scala.html
+++ b/frontend/app/views/index.scala.html
@@ -9,15 +9,6 @@
 @import views.support.Asset
 
 @main("") {
-
-
-    <div class="roundel l-constrained">
-        <div class="roundel__container">
-            <h1 class="roundel__title">Annual Partners and Patrons get 3 months free</h1>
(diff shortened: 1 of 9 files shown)


frontend/app/views/index.scala.html
```diff
@@ -211,7 +202,6 @@ <h3 class="h-intro">
             </div>
             <div class="listing__content">
                 <h4 class="h-intro h-intro--brand u-margin-bottom">Member levels</h4>
-                <p>Join as an annual Partner or Patron Member by <br/> 31 October 2015 and get 3 months free</p>
```
chrisjowen: Is there a reason why we are not mentioning the price changes on the homepage anymore?

chrisjowen:
:+1: but I think we should have the price change info on the homepage too (see comment)

PR #844, opened 2015-11-02
Use membership-common 0.95

tomverran:
This enables the passing back of stripe error codes to JS


joelochlann:
:+1:

tomverran:
confirmed on PROD
Signed up as a friend, signed up as a partner with a bad then good card

PR #846, opened 2015-11-03
Use membership-common 0.96

tomverran:


afiore:
:+1:

tomverran:
I became a friend on prod, which is what broke last time

image

PR #849, opened 2015-11-05
Better example image url for the Event Comment Builder

rtyley:
Better example url in the [event comment builder](https://membership.theguardian.com/staff/event-overview/details):

##### Before
image
##### After
image

Alice in the Guardian Live Events team encountered a problem when she created this event:

https://www.eventbrite.co.uk/edit?eid=19419554410

and entered this incorrect url in the metadata:

The image url is incorrect because it doesn't include the 'crop' section: it should be like this:

...the Grid image service only works with crops, not the raw uploaded image.

cc @joelochlann 

https://trello.com/c/WbMGcAwD/231-give-better-indication-to-event-admins-when-event-image-url-is-entered-incorrectly


rtyley:
Verified on PROD:

image

PR #850, opened 2015-11-06
Add registration bookmarklet to test user page

joelochlann:
I got tired of filling out all the fields on Identity every time I create a new test user, so I created a bookmarklet to prompt for the Test User key then fill out the fields and submit the form. It seemed to me that the best way to share this with the team was to put it as a link on the Test Users page. That way it can be quickly bookmarked by anyone going through the test user creation process.


joelochlann:
@rtyley Does this seem like a good idea?

rtyley:
Looks good. An enhancement would be for the bookmarklet to actually grab/generate the Test User key itself, but this is still cool.

Don't forget that the identity team are going to be heavily tweaking the registration page at some point in the future, so your script might break and people would need to update their bookmarklet.

I think @theefer's Teleporter bookmarklet deals with this problem nicely by keeping the code in the bookmarklet to a minimum - the bookmarklet loads a separately hosted script each time it runs:

https://github.com/guardian/teleporter/blob/bd042aba/src/main/webapp/index.html#L18-L21

joelochlann:
@rtyley Thanks for the suggestions. Have finally found time to implement one of them! I've made the bookmarklet fetch its main script dynamically, as in your example.

I held off merging this until I'd made this improvement because once someone adds the bookmarklet we have to manually ask them to update it. So before everyone starts using this, I wanted to have the capability to dynamically update it.

I've had to tweak our asset pipeline a bit. The bookmarklets are served up in `assets/bookmarklets` and will be copied over into the correct location as part of our `grunt compile` task. The path has to remain the same over time for the bookmarklet to work though so they're treated separately from the rest of our JavaScript. Specifically, we don't hash the asset path in prod, and we keep the same path when doing a prod or dev compile.

One other caveat: if you create your bookmarklet from your dev instance, the script URL will point to your dev instance so won't work when it's not running. Of course you can always use the prod one, even against non-prod instances of identity. But we will still occasionally need the dev bookmarklet when testing modifications to the script before deploying. So in dev, the bookmarklet is named "Dev Register" to distinguish it from the permanently available one hosted on prod.

joelochlann:
cc @mario-galic since I know he loves bookmarklets :bookmark::bookmark::bookmark::bookmark:

rtyley:
> Specifically, we don't hash the asset path in prod, and we keep the same path when doing a prod or dev compile.

Do you know what the cache time will be on that asset path?

rtyley:
:+1:

joelochlann:
> Do you know what the cache time will be on that asset path?

Good point. Following on from our discussion, it turns [they're actually cached for a year](https://github.com/guardian/membership-frontend/blob/master/frontend/conf/application.conf#L112-L113). I think I should find a way to override this just for the bookmarklets

frontend/assets/bookmarklets/identity-signup.js
```diff
@@ -0,0 +1,8 @@
+(function() {
+    var username = prompt('Test User key (first name, last name, username and password will be set to this value):');
+    if (!username) return;
+    ['firstName', 'secondName', 'publicFields_username', 'password', 'primaryEmailAddress'].forEach(function(field) {
+        document.getElementById('user_'+field).value = field === 'primaryEmailAddress' ? username + '@@gu.com' : username;
```
joelochlann: I need to get rid of this `@@` which was to escape Play's templating - incorrect now that this is in its own js file

PR #853, opened 2015-11-10
[WIP] Tell the client when we get a Zuora card decline

joelochlann:
This is linked to [this PR in membership-common](https://github.com/guardian/membership-common/pull/112). We want to pass back some info to the browser when Zuora gets a card decline from its payment gateway.

@afiore


afiore:
As discussed, it would good to push some of the error handling logic into ` MemberService.createMember`. As we want our UI to explicitly handle a number of possible error states (e.g. payment declined), it is time for this method to return a more informative type than `Future[MemberId]`. Consider refactoring it so that it can return something like an `Either[MemberError, MemberId]` or a `MemberResult` type encompassing possible error values ...

joelochlann:
Superseded by #1019

PR #860, opened 2015-11-16
Revert back to #859

afiore:
PR #859 was reverted as I encountered an error while manually
testing the deployed release. After investigating the logs I have
realised that the error was a red haring, and that is in fact part
of the expected program behaviour.

Specifically, I was trying to change plan for a user which already has
a pending amendment, thus incurring into a

    RuntimeError("Cannot amend subscription, amendments are already pending")

Sentry page:
  https://app.getsentry.com/the-guardian/membership/group/84359597/


ostapneko:
:+1:

PR #863, opened 2015-11-24
remove temporary fix forcing a clean sbt update

afiore:
@ostapneko this is not necessary anymore, so I am rolling it back to how it was before


chrisjowen:
:+1:

PR #864, opened 2015-11-25
Update event discount copy to explain booking fees

tomverran:
#### discounted tier

was
image

now
image

#### without discounted tier

was
image

now
image

### DIFF
--- a/frontend/app/model/Eventbrite.scala
+++ b/frontend/app/model/Eventbrite.scala
@@ -198,10 +198,12 @@ object Eventbrite {
 
   case class DiscountBenefitTicketing(generalRelease: EBTicketClass, member: EBTicketClass) {
     val saving = (generalRelease.priceInPence + generalRelease.feeInPence) - member.priceInPence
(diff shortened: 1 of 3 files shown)


chrisjowen:
One really small thing one slightly bigger issue. Firstly can you indent priceInfoEvent.scala.html its really hard to see that the discountTicketing came from the result of the fold. 

Secondly there have been discussions in general about only passing to the view what the view need, i.e. some kind of view model. I appreciate that the prefer that the RichEvent that we pass down is a mixture of a ViewModel and for lack of better word a business model. But passing the Config object to the view too seems odd. I know that Tom and Andrea did some work on separating these things in other places, I wondered if this would be a difficult thing to do here too.

tomverran:
So I do agree passing domain objects into the view is pretty bad, I've removed the dependency on the config at least but splitting out the view logic from the Eventbrite response seems like a big job

chrisjowen:
I appreciate that the size of the change probably exceeds this PR, so :+1

tomverran:
image

image

PR #865, opened 2015-11-25
Added UET tracking for bin advertisment tracking

chrisjowen:
Bing has its own tracking js client that we needed to ensure was called on every page request see: https://trello.com/c/ZVoyTeMT/218-add-the-universal-event-tracking-uet-tag-to-the-site


tomverran:
looks good to me :+1:

PR #867, opened 2015-11-27
Multi currency

ostapneko:
This is only changing the checkout and upgrade form, not the international supporter pages.

The form behaves differently upon joining and upgrading.
  - When you join, the default, preselected country is based on a `country-group` query string. The currency is derived from the selected country for the billing address if it is set, otherwise the delivery address. Whenever the selected country is changed, the currency is changed accordingly
  - When you upgrade, we default to the country set in identity; you are still free to change it. However, we respect only your Zuora account currency (which cannot be change in any way)
  - We constraint the possible upgrade path only to those which make sense for your currencies


frontend/app/actions/CommonActions.scala
```diff
@@ -96,6 +96,29 @@ trait CommonActions {
...
+    override protected def refine[A](request: AnyMemberTierRequest[A]): Future[Either[Result, SubscriptionRequest[A]]] = {
+      val subService = request.touchpointBackend.subscriptionService
+
+      subService.membershipCatalog.get().zip(
+        subService.currentSubscription(request.member)
+      ).map { case (catalog, sub) =>
+        val currentTier = request.member.tier
+        val targetCurrencies = catalog.paidTierDetails(targetTier).currencies
+
+        if (!sub.isInTrialPeriod && targetCurrencies.contains(sub.accountCurrency) && targetTier > currentTier) {
```
ostapneko: We should check that a downgrade is not pending

ostapneko: Also, should be moved to the subscription service

frontend/app/controllers/Joiner.scala
```diff
@@ -82,17 +86,54 @@ trait Joiner extends Controller with ActivityTracking with LazyLogging {
...
+         countriesWithCurrencies = CountryWithCurrency.whitelisted(paidDetails.currencies, GBP),
+         details = paidDetails,
+         userFields = setCountry(privateFields, cg),
+         marketingChoices = marketingChoices,
+         passwordExists = passwordExists,
+         pageInfo = pageInfo))
+    }
+  }
+
+  private def countryGroup(implicit req: RequestHeader): CountryGroup = (for {
```
ostapneko: Move this to the routes

rtyley: https://www.playframework.com/documentation/2.2.x/ScalaRouting#Optional-parameters

frontend/app/actions/CommonActions.scala
```diff
@@ -96,6 +96,29 @@ trait CommonActions {
...
+          Right(SubscriptionRequest(sub, request))
+        } else {
+          Left(Ok(views.html.tier.upgrade.unavailable(currentTier, targetTier)))
```
ostapneko: Log to Cloudwatch whenever we land on this page

frontend/assets/javascripts/src/modules/form/options.js
```diff
@@ -0,0 +1,199 @@
...
+        renderCardDetailsNotes();
+        renderSubmitButton();
+    }
+
+    var hasParent = function (elem) {
+        return elem.parentNode;
+    }
+
+    function hideBillingAddress() {
+        if (!hasParent(BILLING_ADDRESS_EL) && checkoutForm.showBillingAddress) {
```
joelochlann: what is `hasParent` actually checking for here? whether it's actually in the DOM or not? if so maybe this could be made clearer

frontend/app/views/support/PageInfo.scala
```diff
@@ -0,0 +1,38 @@
+package views.support
+
+import com.gu.i18n.{Country, CountryGroup, Currency}
+import com.gu.membership.model.{BillingPeriod, Year}
+import configuration.{Config, CopyConfig}
+import model.EventSchema
+import play.api.libs.json._
+import views.support.PageInfo.FormI18n
+
+case class PageInfo(title: String = CopyConfig.copyTitleDefault,
```
ostapneko: extract it in another PR

frontend/assets/javascripts/src/modules/form/options.js
joelochlann: in fact is there a strong reason why you need to add & remove from the DOM here rather than just hiding/showing the billing address? if not I think hiding/showing would be simpler

frontend/app/views/support/PageInfo.scala
```diff
@@ -0,0 +1,38 @@
...
+                    description: Option[String] = Some(CopyConfig.copyDescriptionDefault),
+                    image: Option[String] = Some(Config.membershipUrl + Asset.at("images/common/mem-promo.jpg")),
+                    schemaOpt: Option[EventSchema] = None,
+                    customSignInUrl: Option[String] = None,
+                    stripePublicKey: Option[String] = None,
+                    formI18n: FormI18n = FormI18n(CountryGroup.UK.defaultCountry, CountryGroup.UK.currency, Year)) {
+  }
+
+object PageInfo {
+  case class FormI18n(defaultCountry: Option[Country],
```
ostapneko: Rename it

frontend/app/views/support/PageInfo.scala
```diff
@@ -0,0 +1,38 @@
...
+import views.support.PageInfo.FormI18n
+
+case class PageInfo(title: String = CopyConfig.copyTitleDefault,
+                    url: String = "/",
+                    description: Option[String] = Some(CopyConfig.copyDescriptionDefault),
+                    image: Option[String] = Some(Config.membershipUrl + Asset.at("images/common/mem-promo.jpg")),
+                    schemaOpt: Option[EventSchema] = None,
+                    customSignInUrl: Option[String] = None,
+                    stripePublicKey: Option[String] = None,
+                    formI18n: FormI18n = FormI18n(CountryGroup.UK.defaultCountry, CountryGroup.UK.currency, Year)) {
```
ostapneko: Rename as well

frontend/assets/javascripts/src/modules/form/options.js
```diff
@@ -0,0 +1,199 @@
...
+        var attr = 'data-' + period;
+        var paymentTakenEl = $(dataCurrencySel(checkoutForm.currency, CARD_NOTE_PRICING_CHARGE), CARD_DETAILS_NOTE_EL)[0];
+        var periodEl = $(dataCurrencySel(checkoutForm.currency, CARD_NOTE_PERIOD), CARD_DETAILS_NOTE_EL)[0];
+
+        paymentTakenEl.innerHTML = paymentTakenEl.getAttribute(attr);
+        periodEl.innerHTML = periodEl.getAttribute(attr);
+    }
+
+    function renderSubmitButton() {
+        toArray($(DATA_CURRENCY_SEL, SUBMIT_BTN_EL)).forEach(function (el) {
```
joelochlann: you can just do `$(DATA_CURRENCY_SEL, SUBMIT_BTN_EL).each(function (el) {`

frontend/assets/javascripts/src/modules/form/options.js
```diff
@@ -0,0 +1,199 @@
...
+    function selectCountry(selectEl, country) {
+        toArray($('option', selectEl)).forEach(function (el) {
+            if ($(el).val() == country) {
+                $(el)[0].selected = true;
+            }
+        });
+    }
+
+    function renderPaymentOptions() {
+        toArray($(BILLING_PERIOD_CONTAINER_EL, PAYMENT_OPTIONS_CONTAINER_EL)).forEach(function (el) {
```
joelochlann: here and similar, the bonzo wrapped set has an `.each()` method so you can call that

frontend/assets/javascripts/src/modules/form/options.js
```diff
@@ -0,0 +1,199 @@
...
+            toggleCurrency(el);
+            setBillingPeriod(el);
+        });
+    }
+
+    function renderCardDetailsNotes(period) {
+        toArray(CARD_NOTE_CURRENCIES_ELS).forEach(function (el) {
```
joelochlann: again, wrapping this in a `$()` and using `.each()` would be better

frontend/assets/javascripts/src/modules/form/options.js
```diff
@@ -0,0 +1,199 @@
...
+                checkoutForm.billingCountry = selectedCountry;
+            }
+
+            renderPrices();
+        });
+
+        bean.on(BILLING_COUNTRY_EL, 'change', function(e) {
+            var input = e.target;
+            var selectedCountry = $(input).val();
+            var selectedEl = toArray($('option', input)).filter(function (el) {
```
joelochlann: You should be able to do `$('option:selected', input')` rather than all this

joelochlann: Actually I think that's just a [jQuery thing](https://api.jquery.com/selected-selector/)... of course it can't be that simple

frontend/app/actions/package.scala
```diff
@@ -14,6 +15,8 @@ package object actions {
 
   type GoogleAuthRequest[A] = AuthenticatedRequest[A, googleauth.UserIdentity]
 
+  val countryGroupKey = "country-group"
```
tomverran: So this is referenced in views and such but doesn't seem to be read here, am I missing something?
ideally it'd be in the same place that reads it

frontend/app/controllers/FrontPage.scala
```diff
@@ -1,20 +1,20 @@
...
 
 trait FrontPage extends Controller {
-
   val liveEvents: EventbriteService
   val localEvents: EventbriteService
   val masterclassEvents: EventbriteService
 
   def index = CachedAction.async { implicit request =>
-    implicit val currency: Currency = GBP
+    implicit val currency = GBP
```
tomverran: So this is showing up in a few places, I remember a discussion about making the default currency more global

frontend/app/controllers/Staff.scala
```diff
@@ -27,8 +27,8 @@ trait Staff extends Controller {
   }
 
   def catalogDiagnostics = GoogleAuthenticatedStaffAction.async { implicit request =>
-    val testCat = TouchpointBackend.TestUser.subscriptionService.getMembershipCatalog()
-    val normalCat = TouchpointBackend.Normal.subscriptionService.getMembershipCatalog()
+    val testCat = TouchpointBackend.TestUser.subscriptionService.getMembershipCatalog
```
tomverran: :cat:

frontend/app/services/SubscriptionService.scala
```diff
@@ -148,174 +161,219 @@ class SubscriptionService(val zuoraSoapClient: soap.ClientWithFeatureSupplier,
...
-  def getUsageCountWithinTerm(subscription: rest.Subscription, unitOfMeasure: String): Future[Option[Int]] = {
-    val features = memberTierFeatures(subscription)
-    val startDate = DateTimeHelpers.formatDateTime(subscription.termStartDate)
-
-    val usageCountF = zuoraSoapClient.query[Usage](AndFilter(("StartDateTime", startDate),
-							                                               ("SubscriptionNumber", subscription.subscriptionNumber),
-							                                               ("UOM", unitOfMeasure))).map(_.size)
+  def getUsageCountWithinTerm(subscription: model.Subscription, unitOfMeasure: String): Future[Option[Int]] = {
+    val features = subscription.features
+    //TODO: review date formats here
```
tomverran: I get nervous when I see todos

tomverran:
if you're happy then I'm happy
this is quite a big PR!

:+1:

PR #870, opened 2015-12-03
Remove /user/me/details endpoint

ostapneko:
Deprecated in favor of the members data API

@tomverran


tomverran:
looks good to me when ci passes :+1:

PR #871, opened 2015-12-03
Include info about Guardian Local

AnnaChesters:
cc @rtyley


AnnaChesters:
Dale has now approved this change.

PR #873, opened 2015-12-03
Description change on event homepage

AnnaChesters:


AnnaChesters:
Dale has approved this change too.

rtyley:
Hey @AnnaChesters - I think you fixed the text on this page:

https://membership.theguardian.com/events/calendar

... but not this page:

https://membership.theguardian.com/events

It looks like that bit of text is duplicated in more than one place, sorry about that :(

You probably want to fix this file:

https://github.com/AnnaChesters/membership-frontend/blob/master/frontend/app/views/event/eventsList.scala.html#L29

AnnaChesters:
You're right! I just tried to change that one too but not sure if it worked....?

PR #877, opened 2015-12-07
Add membership signup bookmarklet

joelochlann:
Added a basic bookmarklet for filling out the membership registration page. It just fills in placeholder data like "Address Line 1", "Address Line 2" etc and valid test card details, and then submits the form. It assumes the first name & last name will be autopopulated because the user is signed in. Obviously it will only work for test users.

See this PR for more history of how we're handling bookmarklets: #850 

@chrisjowen


chrisjowen:
:+1:

PR #878, opened 2015-12-07
Sanity check to avoid polluting logs with bookmarklet errors

joelochlann:
Previously if you fire a bookmarklet on the wrong page, you get JS errors because expected elements are missing. This doesn't do any real harm, but if you're logging clientside errors with Sentry it pollutes the log with these harmless errors. So I've put in a sanity check to bail out early if we think we're on the wrong page.


tomverran:
looks good :+1:

PR #881, opened 2015-12-08
Post multi currency refactoring

ostapneko:
- Remove campaign codes
- Move Stripe customer lookup out of joiner controller
- Group id user properties into a presenter
- User service providers in controllers
- Make join paid a tiny bit clearer

@tomverran @afiore


tomverran:
:+1: 

Trait based dependency management is a lot better than no dependency management at all

PR #885, opened 2015-12-08
Fix ticket sale CTA for Staff tier

rtyley:
The old code [didn't address Staff members](https://github.com/guardian/membership-frontend/blob/d81f23dde/frontend/assets/javascripts/src/modules/events/cta.js#L36-L39) (because they used to be considered Partners) and it also had a fair bit of date-comparison logic occurring on the client-side. I've moved that logic into `views.support.TicketSalesCTA`, and deleted a good chunk of Javascript as a result.

'Become a Member' is the right CTA for a logged-out user, or a logged in one who isn't a member:

image

https://trello.com/c/XfvVAeno/246-bug-advanced-ticket-sales-for-partners-and-patrons-isn-t-working-https-membership-theguardian-com-event-guardian-local-threecour

cc @joelochlann @AnnaChesters


frontend/assets/javascripts/tests/spec/cta.spec.js
```diff
@@ -1,68 +0,0 @@
-/**
```
joelochlann: Would it make sense to add a serverside test to replace this?

joelochlann:
:+1: bar one question about the deleted test

frontend/assets/javascripts/tests/spec/cta.spec.js
rtyley: Good call, I've added this in `TicketSaleCTATest`

rtyley:
Looks good for these cases:

image

image

I'll have to check with @AnnaChesters if we can get an advance ticketing event set up.

PR #886, opened 2015-12-08
Prevent live-events feed from failing on bad 'main-image' annotations

afiore:
Bugfix for https://goo.gl/oPr4K7

Prevents the live events feed to completely fail when an event 'main-image' annotation points to an image with no associated crops/exports.

### Failing example ###

Take this [sample file](https://media.gutools.co.uk/images/7560968d99ed2fcb382a46c19f29a79f1bf42d4d
) uploaded on `media.gutools.co.uk`. Its corresponding JSON representation returned by `https://api.media.gutools.co.uk/images/7560968d99ed2fcb382a46c19f29a79f1bf42d4d` has the following structure:

```json

  "data": {
    "id": "7560968d99ed2fcb382a46c19f29a79f1bf42d4d",
    "uploadTime": "2015-12-08T15:38:36Z",
    "uploadedBy": "andrea.fiore",
    "lastModified": "2015-12-08T15:40:39Z",
    "exports": [],
    "source": {
      "file": "http://media-service-example.com"
   }
``` 

Notice that, as no crops have been created, the `data.exports` attribute is empty. This results in `GridService.getRequestedCrop`  happily returning a blank `Some(GridImage(Nil)...)`, and thus causing a NoSuchElement exception [later down the line](https://github.com/guardian/membership-frontend/blob/43f56ba0fdce3c5b1d59d49e288d429aa6c76c5a/frontend/app/model/ResponsiveImage.scala#L51).

This fix adds additional checks to prevent `GridService` from returning a blank `GridImage` as well as initialising a `ResponsiveImageGroup` from a blank `GridImage`.

### New flag for event-preview dashboard ###

I have also added a new flag for image with no crops (see screenshot below)

screen shot 2015-12-09 at 11 07 25

### DIFF
--- a/frontend/app/model/Eventbrite.scala
+++ b/frontend/app/model/Eventbrite.scala
@@ -12,6 +12,7 @@ import play.api.Logger
 import play.api.data.validation.ValidationError
 import play.api.libs.functional.syntax._
 import play.api.libs.json._
+import services.GridService
 import utils.StringUtils._
 import views.support.Asset
 import views.support.Dates.YearMonthDayHours
@@ -287,6 +288,9 @@ object Eventbrite {
           None
       }
     } yield uri
+
+    val mainImageHasNoCrop: Boolean =
+      mainImageUrl.fold(false)(uri => uri.query.param(GridService.CropQueryParam).isEmpty)
 
     val slug = slugify(name.text) + "-" + id
 
--- a/frontend/app/model/ResponsiveImage.scala
+++ b/frontend/app/model/ResponsiveImage.scala
@@ -13,7 +13,7 @@ object ResponsiveImageGenerator {
 }
 
 object ResponsiveImageGroup {
-  def apply(content: Content): Option[ResponsiveImageGroup] = for {
+  def fromContent(content: Content): Option[ResponsiveImageGroup] = for {
     elements <- content.elements
     element <- elements.find(_.relation == "thumbnail")
   } yield ResponsiveImageGroup(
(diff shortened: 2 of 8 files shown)


frontend/app/model/Eventbrite.scala
```diff
@@ -289,6 +290,9 @@ object Eventbrite {
       }
     } yield uri
 
+    val mainImageWithNoCrop: Boolean =
```
tomverran: Given that this is a boolean it'd be nice to name it something that is more obviously a predicate
like mainImageHasNoCrop or something, I'd expect mainImageWithNoCrop to be an uncropped main image

afiore: Agreed, let's go for `mainImageHasNoCrop`

frontend/app/model/ResponsiveImage.scala
```diff
@@ -13,7 +13,7 @@ object ResponsiveImageGenerator {
 }
 
 object ResponsiveImageGroup {
```
afiore: I feel like overriding `apply` here wasn't great, as we are actually changing the normal signature of apply by returning an `Option[T]` instead of a `T` directly

afiore:
@rtyley @tomverran

frontend/app/model/ResponsiveImage.scala
tomverran: yeah definitely

tomverran:
:+1:

frontend/app/services/GridService.scala
```diff
@@ -16,7 +16,11 @@ import play.api.libs.ws.WSRequest
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
 
+object GridService {
+  val cropQueryParam = "crop"
```
rtyley: Small style nit: should be capitalised http://docs.scala-lang.org/style/naming-conventions.html#constants-values-variable-and-methods

rtyley:
:+1: badge looks good

mario-galic:
Ignore the failed post-deployment test. This is the issue with `skipConfirmation=true` parameter sometimes not being recognised in Saucelabs. I am working on solving this.

afiore:
cool, I was about to ask you... thanks for the heads up!

PR #887, opened 2015-12-09
Ensure staff can get advanced tickets, like they should

rtyley:
In conversation with Nick Miller and @AnnaChesters, confirmed that Staff should be getting advanced Event ticket sales too.

Currently, because Staff _aren't_ getting advanced tickets (they get a  'Upgrade Membership' CTA):

image

...but they _are_ already getting 20% discount tickets, because the list of who gets what was defined separately, in `DiscountTicketTiers`. Now `PriorityBookingTiers = DiscountTicketTiers`, and that drives who gets advanced tickets.

cc @tomverran @JuliaBellis


tomverran:
:+1:

PR #891, opened 2015-12-14
In tablet view, fix content width at 740px and expand margins

joelochlann:
This fixes a longstanding discrepancy in the way we adjust the max content width at various breakpoints. We should have fixed width content and variable margins from 740px (tablet) onwards, snapping the content to larger fixed widths at 980px and 1140px. This is how [theguardian.com](http://theguardian.com) works. But for historical reasons we have always had a variable width tablet view.

For example, at 960px we are still in tablet mode but nearing the 980px desktop breakpoint. We should still have the content at 740px but expanded margins, like so:

picture 59

Before it was like this:

picture 60

@tomverran


tomverran:
:+1: looks fine to me, I didn't realise sass supported maps!

PR #892, opened 2015-12-14
Split MemberService and Subscription service into multiple services

ostapneko:
This commit just moves methods around and do some wiring.
The behaviour is left untouched whenever possible with as
little refactoring as possible

The intent is the following:

- The membership-specific logic goes into MemberService and CatalogService
- MemberService depends on ZuoraService, SalesforceService,
  StripeService, CatalogService and PaymentService


frontend/app/services/MemberService.scala
```diff
@@ -1,136 +1,114 @@
...
-    val apiPassword = salesforceConfig.apiPassword
-    val apiToken = salesforceConfig.apiToken
+  import EventbriteService._
 
-    val stage = Config.stage
-    val application = "Frontend"
+  private val logger = Logger(getClass)
 
-    override val authSupplier: FutureSupplier[Authentication] = new FutureSupplier[Authentication](getAuthentication)
+  override def currentSubscription(contactId: ContactId): Future[model.Subscription] =
```
tomverran: it would be nice to reconcile this class with `SubscriptionService` in membership-common,
this currentSubscription method could probably call `findByProductFamily` in there

Given that this PR is dedicated to moving things rather than changing implementations though this comment is just a note for the future

ostapneko: Precisely; the idea is to first split up thing in frontend, with the intent of moving some stuff into common once we have a clearer picture (as well as bringing back some code from common as well, e.g. `Tier`)

frontend/test/services/MemberServiceTest.scala
```diff
@@ -0,0 +1,49 @@
...
+
+    def plan(t: PaidTier) = PaidTierPlan.yearly(t, Current)
+
+    "return both books and events for patrons" in {
+      await(service.featureIdsForTier(plan(Patron), Set(Books, FreeEventTickets))) mustEqual List(feature1, feature2).map(_.id)
+      await(service.featureIdsForTier(plan(Patron), Set())) mustEqual List(feature1, feature2).map(_.id)
+    }
+
+    "return only one book or event for partner" in {
+      await(service.featureIdsForTier(plan(Partner), Set(Books))) mustEqual List(feature1.id)
```
tomverran: hmm is there some way mark tests as async to avoid all the calls to `await`?  Like the must eventually equal thing which some other test frameworks have

ostapneko: This doesn't come from vanilla Scala, but from specs2 pimped with Play: http://etorreborre.github.io/specs2/guide/SPECS2-3.0/org.specs2.guide.Matchers.html#derive-matchers

tomverran:
:+1:

PR #893, opened 2015-12-15
Remove obsolete update card endpoint

ostapneko:
This is apparently only used by the MMA page, which is now using the members data API.

@tomverran @afiore


tomverran:
There's an update card method in TouchpointBackend I'd quite like to go away if nothing else is using it too

tomverran:
:+1:

PR #894, opened 2015-12-15
Correct fixed widths for event detail page

joelochlann:
This fixes some alignment issues on the event detail page caused by PR #891. In #891 I modified the general mechanism which is used (almost) everywhere for handling the variable width margins and snapping of fixed widths at various breakpoints. Unfortunately, the event detail CSS duplicates this mechanism for some reason and it wasn't changed.

For a future PR the event CSS & HTML is in need of cleanup. As well as various visual defects, it seems to duplicate in a custom way lots of general-purpose layout that we use throughout the site. This is an issue for maintainability.

@tomverran


tomverran:
:+1: 

I also agree with the desire to remove duplicate code

PR #897, opened 2015-12-15
Add missing padding in tablet view of event detail page

joelochlann:
Fixes a longstanding issue in the tablet view of the event detail page (we had no left or right padding on the main body of the event).

#### Before
picture 62

#### After
picture 63


tomverran:
:+1:

PR #899, opened 2015-12-15
Show the default 'Book tickets' message to non-members on Event page

rtyley:
Show non-members a ticket-purchasing button with **'Book tickets'**, rather than **'Become a member'**. @JuliaBellis asked for this change, on the basis that new users on the event page _want to buy a ticket_, and they need a call-to-action that relates to what they want to do.

_NB This is just a text change - the user will still be taken through the Membership sign-up process before being able to purchase the ticket._

**Before:**
image

**After:**
image

cc @joelochlann @davidmcdowell

### DIFF
--- a/frontend/assets/javascripts/src/modules/events/cta.js
+++ b/frontend/assets/javascripts/src/modules/events/cta.js
@@ -9,7 +9,6 @@ define([
         buy: {},
         join: {
             id: 'join',
-            label: 'Become a member',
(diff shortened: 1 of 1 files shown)


afiore:
:+1:

rtyley:
Confirmed on PROD:

image

https://membership.theguardian.com/event/guardian-live-the-digital-mirror-19223927284

PR #901, opened 2015-12-16
Extract explanatory method to explain why we use thumbnails

rtyley:
I was personally puzzled by this until I dug out https://github.com/guardian/membership-frontend/pull/628 - it's just because we want consistent aspect-ratios on our images, so they'll line up nicer.

cc @joelochlann


joelochlann:
:+1: Thanks @rtyley. Would've been nice if someone had put that comment in originally! :stuck_out_tongue_winking_eye:

PR #903, opened 2015-12-16
Ensure 'Book Tickets' button on event is wide enough

joelochlann:
Our px-to-rem plugin allows us to put raw pixel widths in our Sass files - they will later be converted to rem (use of functions like `rem()` elsewhere has been redundant for a while but remain for historical reasons).

The magic number here may come back to bite anyone who stuffs more text into the button later, but it will do for now (no time right now to think of a variable-width solution).

#### Before
picture 68

#### After
picture 67

@tomverran


tomverran:
:+1:

PR #904, opened 2015-12-16
Update mem-common to 0.118

ostapneko:


frontend/app/services/MemberService.scala
```diff
@@ -210,7 +225,9 @@ class MemberService(identityService: => IdentityService,
 
     for {
       newRatePlanId <- catalogService.findProductRatePlanId(newTierPlan)
-      featureIds <- featureIdsForTier(newTierPlan, Set.empty)
+      featureIds <- zuoraService.getFeatures.map { fs =>
```
afiore: It would be better here to just set `featureIds = Set.empty` as we are just generating a preview

afiore:
:+1:

PR #910, opened 2015-12-21
Homepage restyling

joelochlann:
Various styling changes as documented in [this ticket](https://trello.com/c/ks2bfbWQ/226-home-page-design-optimisations) and in screenshots below. Also includes a JS change to select a tier in the table on hover rather than on click.

#### Before
picture 81

#### After
picture 80

#### Before
picture 79

#### After
picture 78

#### Before
picture 83

#### After
picture 84

#### Before
picture 85

#### After
picture 86


afiore:
:+1: for the js part

PR #912, opened 2015-12-22
Filter benefits by country group

joelochlann:
Benefits related to events are not supposed to be promoted outside the UK, so we need to filter these by country group when displaying package feature components on the site.

#### US Supporter Page, Before
picture 87

#### US Supporter Page, After
(price difference is just due to touchpoint env differences)
picture 88

@rtyley

### DIFF
--- a/frontend/app/controllers/Info.scala
+++ b/frontend/app/controllers/Info.scala
@@ -1,6 +1,6 @@
 package controllers
 
-import com.gu.i18n.{GBP, USD}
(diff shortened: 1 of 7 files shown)


rtyley:
:+1:

mario-galic:
Ignore the post-deployment testing fail. It is due to Sauce Labs failing to start a browser: 

```
Cause: org.openqa.selenium.WebDriverException: The Sauce VMs failed to start the browser or device
```

PR #913, opened 2015-12-23
Filter highlights by country group

joelochlann:
Following on from PR #912, we also need to filter the highlights on the left-hand side of the package promo.

#### Before
picture 92

#### After
picture 91

@ostapneko

### DIFF
--- a/frontend/app/views/support/DisplayText.scala
+++ b/frontend/app/views/support/DisplayText.scala
@@ -1,11 +1,11 @@
 package views.support
 
 import com.gu.i18n.CountryGroup
+import com.gu.i18n.CountryGroup._
 import com.gu.salesforce.Tier
 import com.gu.salesforce.Tier._
 import configuration.Config.zuoraFreeEventTicketsAllowance
-import model.Benefits.marketedOnlyToUK
-import model.{Benefits, PaidTierDetails}
+import model.{Highlights, Benefits, PaidTierDetails}
 import views.support.Pricing._
 
 object DisplayText {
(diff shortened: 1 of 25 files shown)


frontend/app/views/support/DisplayText.scala
```diff
@@ -34,23 +34,8 @@ object DisplayText {
...
-        Highlight(s"Includes $zuoraFreeEventTicketsAllowance tickets to Guardian Live events (or 4 Guardian-published books) per year", isNew = true),
-        Highlight("Get priority booking and 20% discount on Guardian Live, Guardian Local and most Guardian Masterclasses")
-      )
-
-      case Patron => List(Highlight("Show deep support for keeping the Guardian free, open and independent."),
-        Highlight("Get invited to a small number of exclusive, behind-the-scenes functions")
-      )
-      case _ => Nil
+    def highlights(countryGroup: CountryGroup = UK) = Highlights.forTier(tier).filterNot {
+      highlight => Highlights.marketedOnlyToUK(highlight) && countryGroup != UK
```
ostapneko: I'd rather make `Benefit.forTier` take another `countryGroup` argument as well, otherwise it feels a bit to ad-hoc I think.

joelochlann: You mean pass the `countryGroup` to `Benefits.forTier` and `Highlights.forTier`?
and do the `.filterNot()` inside that function rather than in `DisplayText`?

ostapneko: s/benefit/highlight

ostapneko:
:+1:

PR #915, opened 2016-01-05
Use OkHttp from mem-common instead of play-git-hub

mario-galic:
@rtyley 

As per [comment](https://github.com/guardian/membership-frontend/commit/1e96508f5687513211b7f8c896fc5eaa0b984f5c#commitcomment-15263470).


rtyley:
:+1:

PR #918, opened 2016-01-06
Update to mem-commons 0.123

afiore:
- Read ratePlanIds configuration from membership-common instead of local `conf` folder


ostapneko:
:+1:

PR #921, opened 2016-01-07
Stop Identity cookie expiration being reported to Sentry

rtyley:
https://app.getsentry.com/the-guardian/membership/issues/94806719/

See also https://github.com/guardian/identity-play-auth/pull/6


rtyley:
Looks good on PROD- the rate of traffic on this issue in membership-frontend was much lower than it was on the Members Data API, in any case. marking resolved in Sentry

PR #925, opened 2016-01-08
Set friend currencies based on delivery address

afiore:
- Upgrade to membership-common v0.126
- Tweak to the catalog diagnostic page
- display an error in the catalog diagnostics page when the friend rate plan does not contain all possible currencies defined in the catalog (this is required in order to allow downgrade to friend)  

screen shot 2016-01-07 at 17 46 48


afiore:
@ostapneko @tomverran

ostapneko:
:+1:

PR #928, opened 2016-01-11
Fix disappearing draft-events 

afiore:
Avoid constructing invalid interval instances. This prevented the entire draft live event feed from displaying. 

screen shot 2016-01-11 at 11 17 13


ostapneko:
:+1:

PR #929, opened 2016-01-11
Use bean library to fire events programmatically

afiore:
This fixes a bug which prevented the join form to working properly for IE8 users


tomverran:
image

PR #930, opened 2016-01-11
Restyle US supporter landing page

joelochlann:
A reskin of the US supporter page according to designs in these tickets:

https://trello.com/c/9pSkQIiM/272-updates-to-us-supporter-landing-page
https://trello.com/c/3gH0adzO/161-international-supporter-row-eu-supporter-landing-pages

with the exception of the following which will be done in separate PRs:
* Package promo restyling, sitewide
* Quote box restyling, sitewide
* Remove the following nav links: events, masterclasses, competitions, patrons
* price should be $49/year, $4.99/month (take out US)

Preview of changes (full-width version only, though there are lots more changes at mobile/tablet/desktop widths):

#### Before
picture 104

#### After
picture 105

#### Before
picture 108

#### After
picture 109

#### Before
picture 110

#### After
picture 111


joelochlann:
@AWare @tomverran

frontend/app/views/info/supporterUSA.scala.html
```diff
@@ -1,66 +1,107 @@
...
+@main("Supporters", pageInfo=pageInfo) {
+    @fragments.page.heroBanner("hero-banner--us-supporter") {
+        Become a Guardian&nbsp;Supporter
+    } {
+        Supporters keep our journalism fearless and free to tell the stories other news organizations won’t.
+        <span class="text-pale">Plus, join today and $1 will be donated to the Guardian Charity Appeal.</span>
+    }()
+
+    @section(hasBorder = false) {
+        @fragments.tier.packagePromoUSASupporter(supporterPlans)
```
tomverran: I may be misreading this but is hasBorder used anywhere

tomverran: ah I am indeed misreading this, ignore me!

tomverran:
:+1:

frontend/app/views/fragments/tier/packagePromoUSASupporter.scala.html
```diff
@@ -1,50 +1,46 @@
...
+                <li>Welcome pack, card and gift</li>
+                <li>Regular Members updates</li>
+            </ul>
+        </div>
         <div class="package-promo__actions">
-            <a class="action action--snap u-no-bottom-margin qa-package-@Supporter.toString.toLowerCase" @actionAttrs>
+            <div class="package-promo__actions__pricing">
+                @fragments.pricing.paidPriceInfo(supporterPlans, US.currency)
+            </div>
+            <a class="action action--no-icon u-no-bottom-margin qa-package-@Supporter.toString.toLowerCase" @actionAttrs>
```
tomverran: oh actually is Supporter.toString explicitly defined or is this relying on scala's default implementation of toString?
There might be a more reliable name field you can use somewhere that doesn't depend on internal details as much as toString does

joelochlann: Ha, yes, well spotted. The existing code actually manages to work by sheer fluke - `Supporter` is now a case class so its default `.toString` gives "Supporter"! But it should be using the instance of the case class, `supporter`.

@ostapneko Why did these get changed from case objects to case classes? It's a bit confusing and I can't see why you would create another instance of a tier case class

ostapneko: @joelochlann We need to have both a `Supporter` trait and a `Supporter` value. In theory you could do

```scala
trait Supporter
case object Supporter extends Supporter
```

but then Scala gets confused sometimes between the trait and the value. Otherwise you can disambiguate with `Supporter.type` but it's arguably not very readable.

By using a nullary case class, you can get both a value and a class in a compact way. But it looks like we didn't manage to catch ALL the regressions, particularly in views.

PR #931, opened 2016-01-12
Removed unnecessary AWS details.

AWare:
Removed unnecessary AWS details.
Fixed aws cp command to use correct path and run as sudo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,15 +72,12 @@ Follow the instructions in [`/nginx/README.md`](./nginx/README.md) in this proje
 
 ### Setup AWS credentials
 
-Setup AWS credentials. Ask your teammate to create an account for you and securely send you the access key. For security, you must enable [MFA](http://aws.amazon.com/iam/details/mfa/).
-
-In `~/.aws/credentials` add the following:
-
+Install the awscli:
 ```
-[membership]
-aws_access_key_id = [YOUR_AWS_ACCESS_KEY]
-aws_secret_access_key = [YOUR_AWS_SECRET_ACCESS_KEY]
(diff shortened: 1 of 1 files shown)


AWare:
@adamnfish re: aws details

README.md
```diff
@@ -94,7 +91,7 @@ region = eu-west-1
 Download our private keys from the `membership-private` S3 bucket. If you have the AWS CLI set up you can run:
 
 ```
-aws s3 cp s3://membership-private/DEV/membership-keys.conf /etc/gu --profile membership
```
adamnfish: I'd recommend using `~/.gu` instead of `/etc/gu`. This has quite a few benefits but the main one that's relevant here is that you won't need to run this command as root.

README.md
```diff
@@ -72,16 +72,13 @@ Follow the instructions in [`/nginx/README.md`](./nginx/README.md) in this proje
...
 ```
-[membership]
-aws_access_key_id = [YOUR_AWS_ACCESS_KEY]
-aws_secret_access_key = [YOUR_AWS_SECRET_ACCESS_KEY]
+brew install awscli
 ```
 
+Fetch the developer AWS credentials, discuss how we do this with a team member.
+
 Ensure you have the following in `~/.aws/config` (you may already have this from setting up another AWS application):
```
adamnfish: While you're updating the AWS details in this document anyway, this bit doesn't do anything and should probably also come out.

(the preference in the `default` profile won't apply to your `membership` profile)

tomverran:
:+1:

PR #932, opened 2016-01-12
Avoid erroring when a pending amend already exists

afiore:
As part of the recent effort to clean up Sentry and to reduce the number of noisy exceptions, this handles the pending amends and other `MemberService` error cases as (throwable) values. When a `PendingAmendError ` is returned, we render a specific error page pointing the user to the profile page.

screen shot 2016-01-12 at 17 13 13


afiore:
@tomverran @ostapneko

tomverran:
:+1:

PR #934, opened 2016-01-12
Dollars without country code

joelochlann:
On the US supporter page there's no ambiguity about what $ means, so we don't need to display US$.

#### Before
picture 115

#### After (different prices due to dev env)
picture 114

@afiore


tomverran:
looks okay to me :+1:

dominickendrick:
Oh Prout-bot, you're so fancy

PR #935, opened 2016-01-13
Pull card details from Zuora on the paid to paid upgrade page

tomverran:
So essentially to break our dependency on Salesforce for stripe info we need to:

- stop reading any card information from there
- Use a subscription rather than a contact when checking payment status

This PR handles the former, I am tackling the latter on https://github.com/guardian/membership-frontend/tree/subscription-action-refiner

### DIFF
--- a/frontend/app/controllers/TierController.scala
+++ b/frontend/app/controllers/TierController.scala
@@ -1,10 +1,10 @@
 package controllers
 
-import com.google.gdata.data.docs.Year
+import _root_.services.{IdentityApi, IdentityService}
 import com.gu.i18n.CountryGroup
 import com.gu.i18n.CountryGroup._
 import com.gu.identity.play.PrivateFields
-import com.gu.memsub.{ProductFamily, Membership, Month, BillingPeriod}
+import com.gu.memsub._
 import com.gu.memsub.BillingPeriod._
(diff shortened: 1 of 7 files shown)


frontend/app/controllers/TierController.scala
```diff
@@ -87,7 +87,7 @@ trait UpgradeTier extends StripeServiceProvider with CatalogProvider {
       PageInfo(initialCheckoutForm = formI18n, stripePublicKey = stripeKey)
     }
 
-    def fromFree(subscription: FreeSubscription, contact: Contact[FreeTierMember, NoPayment]): Future[Result] =
+    def fromFree(subscription: FreeSubscription, contact: Contact[_, _]): Future[Result] =
```
tomverran: I'm using underscores here to force compile errors if you try to do anything with the type parameters as I want to remove them soon

ostapneko:
:+1:

tomverran:
Yep I'm going to remove `PaymentStatus` from Contact in a branch on membership-common when I go through replacing `PaidMemberAction` with `PaidSubscriptionAction` or similar

ostapneko:
:+1: Provided the Contact loses its parametrisation on PaymentStatus right after

PR #936, opened 2016-01-13
Soft-deploy session-cam as a switchable feature

afiore:
Inclusion of the Sessioncam beacon is triggered by the presence of the `sessioncam-on` cookie. This can be set by staff members by visiting '/staff/sessioncam/on' and unset by visiting '/staff/sessioncam/off'

We are not adding it by default as it does add an extra **~65KB** to our JS build!

**Without Sessioncam**
screen shot 2016-01-13 at 13 12 14

**With Sessioncam**
screen shot 2016-01-13 at 13 21 39

### DIFF
--- a/frontend/app/controllers/Staff.scala
+++ b/frontend/app/controllers/Staff.scala
@@ -4,9 +4,10 @@ import com.gu.memsub.Membership
 import com.gu.memsub.services.CatalogService
 import play.api.libs.concurrent.Akka
 import play.api.libs.concurrent.Execution.Implicits.defaultContext
-import play.api.mvc.Controller
+import play.api.mvc.{DiscardingCookie, Cookie, Controller}
 import services._
 import views.support.Catalog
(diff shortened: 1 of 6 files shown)


afiore:
@rtyley @tomverran

rtyley:
Lgtm, but one small tweak:should not need to authenticate to access the urls

frontend/app/controllers/Staff.scala
```diff
@@ -30,6 +31,16 @@ trait Staff extends Controller {
     Ok(views.html.eventOverview.details(request.path))
   }
 
+  // Temporary solution to evaluate SessionCam without actually activating it across the entire site
+  def dropSessionCamCookie = GoogleAuthenticatedStaffAction { implicit request =>
```
rtyley: Remove `GoogleAuthenticatedStaffAction`  - shouldn't need to authenticate to try this out!

PR #938, opened 2016-01-13
Zuora cards attempt 2

tomverran:
membership-common 129 fixes the behaviour when you don't have a payment method


tomverran:
@ostapneko would you mind re-reviewing? Just 6559f1b that is new really

tomverran:
I will merge this tomorrow morning

PR #939, opened 2016-01-13
Fix imagemin by pegging vinyl-fs to 2.2.1

tomverran:


afiore:
:+1: We should ban `^` from our `package.js` files

PR #942, opened 2016-01-18
Improve country/currency detection for new users

afiore:
- Handle the country string coming from identity as either a country code (alpha2) or a country name, improving the likelihood of pre-selecting the user country correctly in the payment form.
- Make the country/currency pre-selection business rules a bit more explicit, by factoring them in the `CheckoutForm.forIdentityUser` method.


afiore:
@tomverran

tomverran:
:+1: when membership-common is done, I approve of the tests

PR #944, opened 2016-01-18
Create EU supporter page

joelochlann:
Added a European supporter landing page at `/eu/supporter`. It isn't currently linked to from anywhere so in that sense it's not "live" yet. When we add a country switcher at the top and start sending promotional emails linking to this page, it will be considered "live". Before that point, we will need to:
* Add rate plans in Euros (currently it falls back to GBP because this rate plan is missing)
* Get a final confirmation of copy & images from Commercial

I've factored out some common components into their own templates, but have opted for a separate template for each regional supporter page because of the amount of potential variation between regions (pretty much all text, the video, most images).

picture 116
picture 117

@tomverran


tomverran:
looks good to me, I agree that the promo pages have a strong chance of diverging over time :+1:

PR #947, opened 2016-01-20
Remove 'free of charge' from supporter highlight text

joelochlann:
This affects package promos around the site, e.g.

#### Before
picture 120

#### After
picture 119

#### Before
picture 122

#### After
picture 121

@tomverran


tomverran:
:+1:

PR #948, opened 2016-01-20
Create rest of world supporter page

joelochlann:
At `/intl/supporter`

picture 123

@tomverran


frontend/app/views/info/supporterInternational.scala.html
```diff
@@ -0,0 +1,72 @@
+@import com.gu.i18n.CountryGroup
+@import com.gu.memsub.Current
+@import com.gu.salesforce.Tier.Supporter
+@import com.gu.membership.PaidMembershipPlans
+@import views.support.PageInfo
+@(supporterPlans: PaidMembershipPlans[Current, Supporter],
+    pageInfo: PageInfo,
+    pageImages: Seq[model.ResponsiveImageGroup])(implicit token: play.filters.csrf.CSRF.Token, countryGroup: CountryGroup)
```
ostapneko: Why do we need to pass a `CountryGroup`, if this page is specific to `RestOfTheWorld` ? It might as well be hardcoded

ostapneko:
:+1:

frontend/conf/routes
```diff
@@ -91,6 +91,7 @@ GET            /patrons                               controllers.Info.patron
 GET            /supporter                             controllers.Info.supporter
 GET            /us/supporter                          controllers.Info.supporterUSA
 GET            /eu/supporter                          controllers.Info.supporterEurope
+GET            /intl/supporter                        controllers.Info.supporterInternational
```
ostapneko: Shouldn't this be `/intl/supporter` like in theguardian.co.uk

joelochlann: ????

joelochlann: it is `/intl/supporter`

ostapneko: sorry `int/supporter` :)

frontend/app/views/info/supporterInternational.scala.html
joelochlann: Yes, fair point, but I'm not sure whether we might start supporting more countries and sharing templates between them and it seemed like this might make refactoring a tiny bit easier. Plus we already had this implicit parameter in existing controllers

frontend/conf/routes
joelochlann: which other link are you comparing it to? `theguardian.com/int/supporter` is a 404, theguardian.co.uk redirects to `http://www.theguardian.com/uk `... where is the `/int/supporter`?

ostapneko: http://www.theguardian.com/preference/edition/int

As a general rule, we tried to map the country groups to what we had on the main website, that is us, uk, au and int, that's why the CountryGroup ids are exactly those.

But I guess that's not the end of the world, it's just a route. It's just a bit confusing, as the associated countryGroup query parameter is `int` instead of `intl`

joelochlann: OK, `int` makes more sense then. I used `intl` because it was mentioned in the Trello ticket: https://trello.com/c/3gH0adzO/161-international-supporter-row-eu-supporter-landing-pages

joelochlann: Created PR #949 to fix

PR #954, opened 2016-01-21
Fixing analytics in Firefox

paulbrown1982:
Fixing bug where analytics did not work on Firefox, because navigator.doNotTrack === "unspecified" so prefixing ! returned true rather than false. Chrome returns 'null' for that value. Also fixed that IE and Safari use window.doNotTrack rather than navigator.

/cc @rtyley 

Also see https://github.com/guardian/subscriptions-frontend/pull/329


rtyley:
:+1: we should just see this work for subscribe.theguardian.com before we push this out on membership.

PR #955, opened 2016-01-22
Increase Java heap memory to 1024

mario-galic:
@afiore @rtyley 

Resolves [Sentry event](https://app.getsentry.com/the-guardian/membership/issues/96432625/)

Increased allocated Java heap memory from 512MB to 1024MB to prevent future out of memory crashes.

Note that AWS m3.medium instance on which mem-frontend runs has 3.75 GB RAM available.


afiore:
:+1:

PR #956, opened 2016-01-22
Increase Content API query period to 30 min

mario-galic:
@afiore @rtyley

Increase Content API query period from 2 minutes to 30 minutes.

[Sentry issue](https://app.getsentry.com/the-guardian/membership/issues/68558521/activity/)


afiore:
:+1:

PR #958, opened 2016-01-24
Add country switcher

joelochlann:
Added a dropdown menu to switch between localised variants of the supporter page. This only appears on supporter landing pages (`/supporter`, `/us/supporter`, `/eu/supporter`, `/int/supporter`).

picture 129

@paulbrown1982  @tomverran


tomverran:
looks ok to me :+1:

PR #960, opened 2016-01-25
Geocode

afiore:
redirect users heading to the `/supporter` path to a country-specific supporter page based on the `X-Fastly-Country-Code` value


afiore:
@joelochlann

joelochlann:
:+1:

PR #962, opened 2016-01-26
Revert "Geocode"

afiore:
Reverts guardian/membership-frontend#960


afiore:
This brings back master into a workable state, as membership commons 0.144 is failing

PR #963, opened 2016-01-26
Change default country and currency when joining

ostapneko:
Give priority to the requested country group as opposed to the identity user one.

@afiore @joelochlann


afiore:
:+1:

PR #966, opened 2016-01-26
Updated twitter hashtag

JuliaBellis:
We were tweeting with a #GuardianMembership hashtag which is not in line with the 'new' branding.

@rtyley


rtyley:
:+1:

JuliaBellis:
Merging PR - Paul think's it's ok!

PR #968, opened 2016-01-27
Remove duplicated ScheduledTask

mario-galic:
@tomverran @rtyley 

Use ScheduledTask from membership-common.


afiore:
As the akka actorSystem is global, perhaps we could define its implicit as part of  the main [config object](https://github.com/guardian/membership-frontend/blob/master/frontend/app/configuration/Config.scala).

afiore:
:+1: overall

frontend/app/configuration/Config.scala
```diff
@@ -183,4 +184,8 @@ object Config {
 
   def digipackRatePlanIds(env: String) = DigitalPackRatePlanIds.fromConfig(
     config.getConfig(s"touchpoint.backend.environments.$env.zuora.ratePlanIds.digitalpack"))
+
+  object Implicits {
+    implicit val akkaSystem = Akka.system
```
rtyley: Surprised this is necessary, but ok!

rtyley:
LGTM

PR #969, opened 2016-01-27
Fix handling of events benefit

afiore:
Fixes a bug whereby the free event ticket benefit was not handled correctly by the app


afiore:
@tomverran

tomverran:
looks good to me :+1:

PR #976, opened 2016-02-04
Bump to latest version of mem-common, and associated versions

rtyley:
Latest version is from guardian/membership-common#175

cc @tomverran


tomverran:
:+1:

PR #977, opened 2016-02-05
Turned on sessioncam for all users.

paulbrown1982:
Turned on sessioncam for all users.
Removed the old feature toggling cookie and endpoints to set the cookie.

/cc @rtyley 
/cc @tomverran


rtyley:
:+1:

PR #978, opened 2016-02-05
Fix the eventbrite banner

tomverran:
So two problems fixed here

- We were only fetching tickets you managed to buy at the exact same time you signed up
-  By using the _subscription_ start date rather than the _term_ start date we would not have reset allocations each year, had we been picking them up at all

image

### DIFF
--- a/frontend/app/services/MemberService.scala
+++ b/frontend/app/services/MemberService.scala
@@ -23,7 +23,7 @@ import forms.MemberForm._
 import model.Eventbrite.{EBCode, EBOrder, EBTicketClass}
 import model.RichEvent.RichEvent
 import model.{PaidSubscription, _}
-import org.joda.time.DateTime
+import org.joda.time.{DateTimeZone, DateTime}
 import play.api.Logger
 import play.api.libs.concurrent.Execution.Implicits.defaultContext
 import tracking._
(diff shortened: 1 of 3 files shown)


AWare:
⭐️⭐️⭐️⭐️⭐️
:+1:

frontend/app/services/MemberService.scala
```diff
@@ -305,7 +305,7 @@ class MemberService(identityService: IdentityService,
 
   override def getUsageCountWithinTerm(subscription: Subscription, unitOfMeasure: String): Future[Option[Int]] = {
     val features = subscription.features
-    val startDate = subscription.startDate.toDateTimeAtCurrentTime()
+    val startDate = subscription.startDate.toDateTimeAtStartOfDay(DateTimeZone.forID("America/Los_Angeles"))
```
rtyley: Should `subscription.startDate` have changed to `subscription.termStartDate` with this pull request? The PR description makes it sound like it should:

```
By using the subscription start date rather than the term start date we would not have reset allocations
each year, had we been picking them up at all
```

tomverran: gosh, yes, not my finest hour 😨  I think I might've had rebase troubles or something.

rtyley: Ha! Thanks Tom! I was wondering if there was some weird business logic reason why we couldn't use `termStartDate` after all - like maybe it doesn't work well with 6-months-free membership offer, which thankfully we're no longer doing...

tomverran: I spent my train journey home digging and it looks like startDate used to refer to the term start date. Finding this confusing I rather dangerously changed it shortly before I left: https://github.com/guardian/membership-frontend/commit/7216c6c35c9d96fe615e75276bf1d6a902679abd#diff-abb6c51f77945af0f3a1eca1252ea19bL298

So hopefully we at least got 8 months of working tickets between Feb and Oct, and term start date is definitely what it should be

PR #979, opened 2016-02-05
Use memsub-common-play-auth, remove google-auth config from Membership

rtyley:
The releveant code is now in:

https://github.com/guardian/memsub-common-play-auth

This is working towards having common code for Membership and Subscriptions
with Google-Group based restrictions:

https://trello.com/c/boSbvE9P/356-user-authentication-google-groups-for-new-subscription-checker-tool-that-works-with-zuora-as-well-as-legacy-cas

cc @AWare

### DIFF
--- a/frontend/app/actions/ActionRefiners.scala
+++ b/frontend/app/actions/ActionRefiners.scala
@@ -1,16 +1,16 @@
 package actions
 
-import services._
 import actions.Fallbacks._
-import com.gu.googleauth.{GoogleGroupChecker, UserIdentity}
+import com.gu.googleauth.UserIdentity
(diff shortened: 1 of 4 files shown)


AWare:
:+1:

rtyley:
> Post-deployment testing failed!

This was due to the integration test hitting the new https://github.com/guardian/identity-frontend - the integration test still expects the old-style pages, and thus dies. The brilliant @AWare has volunteered to add support for dropping the correct cookie to ensure we don't go to the new version (yet).

As for this PR, google login is still correctly enforced on going to https://membership.theguardian.com/test-users

PR #984, opened 2016-02-12
Tweak acceptance tests to point at new identity frontend (and point dev at it too)

AWare:
Identity are a/b testing their new frontend, and will be turning their old one off soon. 

Get the new one from https://github.com/guardian/identity-frontend

Bookmarklets are currently broken, and will follow in a new pr. The new identity frontend sets more restrictive Content-Security-Policy headers, so we can't load bookmarklets so easily.

### DIFF
--- a/frontend/app/views/testing/testUsers.scala.html
+++ b/frontend/app/views/testing/testUsers.scala.html
@@ -37,6 +37,24 @@
                         <li>
                             <a href="javascript:
                             (function() {
+                                if (!document.getElementById('register_field_username')) {
+                                    alert('Element with id register_field_username not found. Are you on the right page?');
+                                    return;
+                                }
+                                var username = prompt('Test User key (first name, last name, username and password will be set to this value):');
+                                if (!username) return;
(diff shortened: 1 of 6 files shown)


paulbrown1982:
:+1:

AWare:
Putting this on hold until identity-frontend supports a logout route. 
:-1:

frontend/app/views/testing/testUsers.scala.html
```diff
@@ -37,6 +37,24 @@
                         <li>
                             <a href="javascript:
                             (function() {
+                                if (!document.getElementById('register_field_username')) {
```
joelochlann: This should go in a separate script like the existing bookmarklet - this is so you can update it without users needing to manually delete & recreate their bookmarklets

AWare:
Spoken with @guardian/identity signout route not in new frontend. This only affects dev, and we'll just clear our cookies. 
a cookie

:+1:

AWare:
^Good spot :+1:

PR #989, opened 2016-02-16
Updating membership-frontend to the latest membership-common

paulbrown1982:
I haven't wired the roundelHtml from the model into the view yet - it's still using the homepage's template. We can generalise the roundel next sprint when we work on price-reducing promotions.

/cc @AWare @tomverran


tomverran:
:+1: 

I do feel that we should aim to remove the very view specific fields like roundelHtml in promo code as early as possible though

paulbrown1982:
Agreed, we'll discuss it today with @rtyley in time for sprint planning tomorrow.

tomverran:
Merging this as I need it for Zuora timezone work

PR #990, opened 2016-02-16
English heritage landing page continued

AWare:
screen shot 2016-02-16 at 12 06 20
screen shot 2016-02-16 at 12 06 43

Need to deal with responsive lead_in and hero banner image position. These will need new pull requests.


paulbrown1982:
:+1:

PR #1000, opened 2016-02-22
fix omniture codes getting lost when geolocating

tomverran:
Omniture puts the INTCMP tracking code into a cookie in JavaScript so it won't work unless the code is present on a page where JS executes. The supporter geolocating page chops off the tracking code currently so there's no tracking :crying_cat_face: 

And PR 1000...

...For a bug.

Yay? :neutral_face:


AWare:
💯*10.

:+1:

PR #1001, opened 2016-02-22
Update cloud-formation script user data section

paulbrown1982:
* Updated cloud-init script to download app via the AWS CLI rather than wget, as it's faster.
* Also added option for use on a t2.medium AMI, which we now use.
* Removed no-longer-used "apt-get -y update"

/cc @rtyley @mario-galic


rtyley:
:+1:

PR #1003, opened 2016-02-23
Approved images for EH promocode

AWare:
Final english heritage promo code copy and images.
 
screen shot 2016-02-23 at 16 54 42
screen shot 2016-02-23 at 16 59 03


joelochlann:
:+1:

PR #1004, opened 2016-02-24
Use membership-common 161

tomverran:
Membership common 161 makes Subscribe take a NonEmptyList of RatePlans so this PR just brings membership-frontend up to date with that


AWare:
:+1:

PR #1008, opened 2016-02-25
Fix guava-jdk5 dependency introduced by Google Directory API client

rtyley:
We don't currently use DirectDebit/Go-Cardless on Membership, but the Google Groups library was still pulling in the awful `guava-jdk5` dependency, leading to hidden classpath collisions.

https://github.com/guardian/play-googleauth/pull/37
https://github.com/guardian/subscriptions-frontend/pull/363#issuecomment-186190081
https://github.com/guardian/subscriptions-frontend/pull/366

cc @paulbrown1982


paulbrown1982:
:+1:

rtyley:
Was able to successfully sign up as a supporter with a  test-user on PROD:

image

PR #1009, opened 2016-02-25
Forward the GU-IdentityToken or the SC_GU_U cookie to identity

DiegoVazquezNanini:
Makes Membership frontend use the SC_GU_U cookie or the GU-IdentityToken header containing the identity app specific.
It passes all authentication mechanics to the id api and upgrade identity-play-auth with a fix for a null pointer exception.

### DIFF
--- a/frontend/app/controllers/IdentityRequest.scala
+++ b/frontend/app/controllers/IdentityRequest.scala
@@ -7,16 +7,16 @@ case class IdentityRequest(headers: List[(String, String)], trackingParameters: 
 
 object IdentityRequest extends RemoteAddress {
   def apply(request: Request[_]): IdentityRequest = {
-    val cookie = request.cookies.get("SC_GU_U")
     val ipAddress = clientIp(request)
 
-    val headers = List(("X-GU-ID-Client-Access-Token" -> s"Bearer ${Config.idApiClientToken}")) ++
-      cookie.map("X-GU-ID-FOWARDED-SC-GU-U" -> _.value)
+    val headers = List("X-GU-ID-Client-Access-Token" -> s"Bearer ${Config.idApiClientToken}") ++
+      request.cookies.get("SC_GU_U").map(_.value).map("X-GU-ID-FOWARDED-SC-GU-U" -> _) ++
+      request.headers.get("GU-IdentityToken").map("Authorization" -> _)
 
     val trackingParameters = List() ++
-      request.headers.get("Referer").map(("trackingReferer" -> _)) ++
-      request.headers.get("User-Agent").map(("trackingUserAgent" -> _)) ++
-      ipAddress.map(("trackingIpAddress" -> _))
+      request.headers.get("Referer").map("trackingReferer" -> _) ++
+      request.headers.get("User-Agent").map("trackingUserAgent" -> _) ++
+      ipAddress.map("trackingIpAddress" -> _)
 
     IdentityRequest(headers, trackingParameters)
 
--- a/frontend/app/services/AuthenticationService.scala
+++ b/frontend/app/services/AuthenticationService.scala
@@ -9,9 +9,9 @@ object AuthenticationService extends com.gu.identity.play.AuthenticationService 
 
   val identityKeys = Config.idKeys
 
-  def authenticatedIdUserProvider(targetClientId: String): (RequestHeader) => Option[AuthenticatedIdUser] = AuthenticatedIdUser.provider(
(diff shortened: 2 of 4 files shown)


frontend/app/services/AuthenticationService.scala
```diff
@@ -1,9 +1,17 @@
...
+import play.api.mvc.RequestHeader
 
 object AuthenticationService extends com.gu.identity.play.AuthenticationService {
   def idWebAppSigninUrl(returnUrl: String) = Config.idWebAppSigninUrl(returnUrl)
 
   val identityKeys = Config.idKeys
+
+  override lazy val authenticatedIdUserProvider: (RequestHeader) => Option[AuthenticatedIdUser] = AuthenticatedIdUser.provider(
+    AccessCredentials.Cookies.authProvider(identityKeys),
+    AccessCredentials.Token.authProvider(identityKeys, "members-data-api")
```
rtyley: Gah, we do want the Membership-Frontend and the Members-Data-API (on members-data-api.theguardian.com) to have the same target-client-id.

I know we spoke about this on Friday, but I now feeling guilty about deciding on `members-data-api` as the common target-client-id, when it should probably be `membership`. No one is actually using it yet (although the mobile team are just getting started with using it!) so maybe we still are in time to change it before all the mobile clients go out! Would you be amenable to that @DiegoVazquezNanini ? If so, this would be the only other code we need to change:

https://github.com/guardian/membership-attribute-service/blob/34406d9a3/membership-attribute-service/app/services/IdentityAuthService.scala#L15

frontend/app/controllers/IdentityRequest.scala
```diff
@@ -7,16 +7,16 @@ case class IdentityRequest(headers: List[(String, String)], trackingParameters:
...
 object IdentityRequest extends RemoteAddress {
   def apply(request: Request[_]): IdentityRequest = {
-    val cookie = request.cookies.get("SC_GU_U")
     val ipAddress = clientIp(request)
 
-    val headers = List(("X-GU-ID-Client-Access-Token" -> s"Bearer ${Config.idApiClientToken}")) ++
-      cookie.map("X-GU-ID-FOWARDED-SC-GU-U" -> _.value)
+    val headers = List("X-GU-ID-Client-Access-Token" -> s"Bearer ${Config.idApiClientToken}") ++
+      request.cookies.get("SC_GU_U").map(_.value).map("X-GU-ID-FOWARDED-SC-GU-U" -> _) ++
+      request.headers.get("GU-IdentityToken").map("Authorization" -> _)
```
rtyley: So there will be a change in Identity to make the client-specific `GU-IdentityToken` allowed to perform the required actions? Add a link here when you create the PR, I'd like to see it!

frontend/app/controllers/IdentityRequest.scala
```diff
@@ -7,16 +7,16 @@ case class IdentityRequest(headers: List[(String, String)], trackingParameters:
...
 
     val trackingParameters = List() ++
-      request.headers.get("Referer").map(("trackingReferer" -> _)) ++
-      request.headers.get("User-Agent").map(("trackingUserAgent" -> _)) ++
-      ipAddress.map(("trackingIpAddress" -> _))
+      request.headers.get("Referer").map("trackingReferer" -> _) ++
+      request.headers.get("User-Agent").map("trackingUserAgent" -> _) ++
+      ipAddress.map("trackingIpAddress" -> _)
```
rtyley: At some point we can probably abstract this at little using [`AccessCredentials`](https://github.com/guardian/identity-play-auth/blob/6a05d186a7/src/main/scala/com/gu/identity/play/AuthenticatedIdUser.scala#L30) to generate suitable forwarding headers (so, whatever the user authenticated with, we'll pass that on) - this is all fine for now tho'.

rtyley:
LGTM, except for the question in https://github.com/guardian/membership-frontend/pull/1009/files#r54554397 about whether we can be bothered to change our target-client-id to `membership`

frontend/app/services/AuthenticationService.scala
DiegoVazquezNanini: OK. I will change the targetClientId to "membership"

frontend/app/controllers/IdentityRequest.scala
DiegoVazquezNanini: Yes. Identity will make a change to support this.

frontend/app/controllers/IdentityRequest.scala
DiegoVazquezNanini: Next time I touch this code ... I will do it. ;)

rtyley:
:+1:

PR #1010, opened 2016-02-26
New identity branding query param

AWare:
Give the identity links a clientId=members query param to load the membership branding. Leaving comment activity until we update lodash and identity implement that endpoint in their new frontend

### DIFF
--- a/frontend/app/configuration/Config.scala
+++ b/frontend/app/configuration/Config.scala
@@ -42,14 +42,18 @@ object Config {
 
   val idWebAppUrl = config.getString("identity.webapp.url")
 
+  val idMember = "clientId" -> "members"
+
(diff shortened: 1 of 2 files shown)


frontend/app/configuration/Config.scala
```diff
@@ -42,14 +42,18 @@ object Config {
 
   val idWebAppUrl = config.getString("identity.webapp.url")
 
+  val idMember = "clientId" -> "members"
+
+  private val idSkipConfirmation: (String, String) = "skipConfirmation" -> "true"
```
tomverran: could make this non private / private for consistency

tomverran: ah I hear it is used elsewhere

Carry on

tomverran:
:+1:

PR #1012, opened 2016-02-26
let identity devs use test-users

AWare:
Adding the identity devs to the test-users google groups list.


rtyley:
:+1: ...at the very least, accessing https://membership.theguardian.com/test-users will activate the [`ANALYTICS_OFF_KEY`](https://github.com/guardian/membership-frontend/search?utf8=%E2%9C%93&q=AnalyticsCookieName) so the hits won't sway stats.

PR #1014, opened 2016-03-01
Remove PATH modification from pre-commit hook

mario-galic:
@tomverran @rtyley 

Having ```/usr/local/bin``` as the first entry in `PATH` confuses NVM.


tomverran:
:+1:

PR #1016, opened 2016-03-01
Update to membership-common 0.164

mario-galic:
@tomverran @paulbrown1982


tomverran:
:+1: looks good to me, I'd be inclined to wait for @paulbrown1982 to return from his meeting before merging

PR #1017, opened 2016-03-01
Display price correctly on event thank you page

joelochlann:
Think this went awry during Andrea & Tom's big refactoring for international prices.

#### Before:

screen shot 2016-03-01 at 11 32 26

#### After:
picture 134

@rtyley

### DIFF
--- a/frontend/app/views/fragments/event/itemSnapshotOrder.scala.html
+++ b/frontend/app/views/fragments/event/itemSnapshotOrder.scala.html
@@ -1,5 +1,7 @@
 @import configuration.Config
 @import views.support.Asset
+@import com.gu.i18n.GBP
+@import com.gu.memsub.Price
 
 @(event: model.RichEvent.RichEvent, order: model.Eventbrite.EBOrder)
 
@@ -27,7 +29,7 @@
                         <div class="order-summary__id">Order no. @order.id</div>
                         <div class="order-summary__ticket">
                             @order.ticketCount ticket@if(order.ticketCount > 1){s}
-                            @if(order.totalCost > 0) { Pricing.pretty(@order.totalCost, GBP) } else { free }
+                            @if(order.totalCost > 0) { @Price(order.totalCost, GBP).pretty } else { free }
                         </div>
                         <div class="order-summary__reference copy">The order has been saved in <a href="@(Config.eventbriteUrl + "/gettickets")">MyTickets</a> on Eventbrite. We've also sent confirmation to your registered EventBrite email address.</div>
                         <img class="u-margin-top" src="@Asset.at("images/logos/eventbrite.svg")">


rtyley:
Looks good - incidentally, might be worth checking to see if this error occurs anywhere else in the codebase?

rtyley:
Yep, this now working on PROD:

https://membership.theguardian.com/events.json

```
{
   "events":[
      {
         "id":"20940795483",
         "url":"https://membership.theguardian.com/event/guardian-live-a-nations-theatre-conversation-farnham-i-liked-it-but-i-couldnt-book-it-20940795483",
         "mainImage":{
            "id":"ecf6ea115a9c1286865fc792208f514b05e4026b",
            "crop":"0_0_3500_2101"
         }
      },
      ...
   ]
}
```

joelochlann:
Yeah good point. Had a quick grep for `Pricing.pretty` and it doesn't appear anywhere. Anywhere else that `Price` appears looks fine - basically this is the only place where the price isn't wrapped in specific abstractions for either membership tiers or events.

This does make me wonder whether I should use a similar abstraction for `EBOrder` as we do for `EBEvent` and take a bit of logic out of the template

joelochlann:
Now we've got `totalCostText` in `EBOrder`
```scala
  case class EBOrder(id: String, first_name: String, email: String, costs: EBCosts, attendees: Seq[EBAttendee]) extends EBObject {
    val ticketCount = attendees.length
    val totalCost = costs.gross.value / 100f
    val totalCostText = if(totalCost > 0) Price(totalCost, GBP).pretty else "free"
  }
```
to parallel `priceText` and `feeText` in `EBTicketClass`
```scala
  case class EBTicketClass(id: String,
                           name: String,
                           free: Boolean,
                           quantity_total: Int,
                           quantity_sold: Int,
                           on_sale_status: Option[String], // Currently undocumented, so treating as optional. "SOLD_OUT", "AVAILABLE", "UNAVAILABLE"
                           cost: Option[EBPricing],
                           fee: Option[EBPricing],
                           sales_end: Instant,
                           sales_start: Option[Instant],
                           hidden: Option[Boolean]) extends EBObject {
    val isHidden = hidden.contains(true)

    val isGuestList = isHidden && name.toLowerCase.contains("guestlist")

    val isMemberBenefit = isHidden && name.toLowerCase.startsWith("guardian member")
    val isComplimentary = isHidden && name.toLowerCase.startsWith("member's ticket at no extra cost")

    val isSoldOut = on_sale_status.contains("SOLD_OUT") || quantity_sold >= quantity_total

    val priceInPence = cost.map(_.value).getOrElse(0)
    val feeInPence = fee.map(_.value).getOrElse(0)
    val priceValue = formatPrice(priceInPence)
    val priceText = cost.map(_.formattedPrice).getOrElse("Free")
    val feeText = fee.filter(_.value>0).map(_.formattedPrice)
    val totalCost = cost.map(c => c add fee.getOrElse(EBPricing(0)))
    val currencyCode = GBP.toString
  }
```

Seems a tiny bit better. Next would be to get them both using the same code for price formatting...

joelochlann:
Confirmed on prod:

picture 135

PR #1021, opened 2016-03-02
Support custom ticket price descriptions on Events

rtyley:
Tina asked if it would be possible to add more explanatory text next to the ticket price (because the price on [one of their new events](https://membership.theguardian.com/preview-event/22493299063) is quite high due to it including a book). Eventbrite does allow a 'description' field on tickets:

https://www.eventbrite.co.uk/developer/v3/endpoints/events/#ebapi-post-events-id-ticket-classes

So long as this description field isn't being used for any other purpose, I think we should be fine to add it as extra descriptive text. Note that the info must be added to the 'general release' ticket (ie the
first non-hidden ticket). It'll look like this (the descriptive text is _'Price includes signed book'_):

image

It's configured like this in the EventBrite event editor: image

Request from Tina on the Guardian Live Events team:

```
---------- Forwarded message ----------
From: Tina Moynihan
Date: 2 March 2016 at 17:08
Subject: Adding text?

This event price of £23 and includes a signed book. Can we say this where
I have positioned the yellow arrow? Otherwise the high price point might
put people off from clicking through to the event.

Thanks,
Tina

https://membership.theguardian.com/preview-event/22493299063
```

cc @JuliaBellis @joelochlann

### DIFF
--- a/frontend/app/model/Eventbrite.scala
+++ b/frontend/app/model/Eventbrite.scala
@@ -111,6 +111,7 @@ object Eventbrite {
    */
   case class EBTicketClass(id: String,
                            name: String,
+                           description: Option[String],
                            free: Boolean,
                            quantity_total: Int,
                            quantity_sold: Int,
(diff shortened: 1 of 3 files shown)


rtyley:
cc @tomverran

tomverran:
:+1:

PR #1024, opened 2016-03-04
Use serverside redirect not JavaScript for ticket join flow

joelochlann:
This fixes a bug in the join/upgrade flow when purchasing tickets. If you entered the join/upgrade flow by clicking on the "Book Tickets" button, you should see a link to Eventbrite on the thank you page, as below:

picture 143

The serverside logic was (mostly) in place to do this, but because the link href was being changed by JavaScript it was never going via the redirect code path so the `preJoinReturnUrl` was not being set. Also the cookie was never being dropped on the upgrade path, only the join path, so I added that. And I also had to add a condition to prevent this backdoor to priority tickets:

1. non-member, try to book tickets during partner/patron priority period
2. get directed to join flow, choose supporter tier
3. on thank you page, you get the eventbrite link (which should only be available to partners & patrons)

@tomverran @paulbrown1982


tomverran:
:+1:

PR #1025, opened 2016-03-07
Add ability to modify the button in the packagePromo view fragment

tomverran:
So the brief is/was to modify the EH landing page to change the partner highlights text and also to change "Become a partner" to "Find out more" and make it point to the homepage.

I think we should move towards modelling the package promo container contents with objects rather than lots and lots of parameters, so to advance us towards that goal I've modelled the CTA button, the eventual aim being to have one big PackagePromo object we can tweak for any variants

### DIFF
--- a/frontend/app/model/PackagePromo.scala
+++ b/frontend/app/model/PackagePromo.scala
@@ -1,0 +1,8 @@
+package model
+import com.netaporter.uri.Uri
+
+object PackagePromo {
+
+  // the call-to-action button on the package promo view element
+  case class CtaButton(text: String, to: Uri, attributes: Map[String, String])
+}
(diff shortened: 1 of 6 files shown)


tomverran:
@joelochlann would you mind taking a look? thanks!

frontend/app/model/PackagePromo.scala
```diff
@@ -0,0 +1,8 @@
+package model
+import com.netaporter.uri.Uri
+
+object PackagePromo {
+
+  // the call-to-action button on the package promo view element
+  case class CtaButton(text: String, to: Uri, attributes: Map[String, String])
```
joelochlann: maybe this case class doesn't need to be wrapped in an object? not sure what extra benefit we get that isn't already provided by Scala's package system

tomverran: Ah so remembering back I originally had a PackagePromo case class which this was the companion object of, then I scaled back my refactoring since this has a deadline.

I think this would be the place for this class to go if PackagePromo was a case class but since it isn't yet I'm not so sure.

If you're happy for this to be merged I'll spend some time later this week on a PackagePromo case class

joelochlann:
:+1: Looks good! And a better abstraction for the package promo components sounds like a good idea

PR #1030, opened 2016-03-08
Round edges on input fields

tomverran:
image


AWare:
:+1: (Also Ben had a look and liked it)

PR #1031, opened 2016-03-08
Propagate CMP tracking code on redirect

joelochlann:
We already pass through INTCMP but according to Chris Mulholland we should also pass through CMP.

@tomverran


tomverran:
:+1:

PR #1032, opened 2016-03-09
update membership-commmon tand memsub-common-play-auth

AWare:
Updating these together to avoid the transitive dependency issue.


tomverran:
:+1:

PR #1033, opened 2016-03-09
Add event brand text

paulbrown1982:
Adding the event's branding below the location text on a grid item. This will allow us to remove the branding from the event's name in EventBrite.

https://trello.com/c/oVKYarCL/144-don-t-display-events-brand-in-event-title-on-list-page-display-brand-in-design-instead

cc @rtyley


tomverran:
:+1:

PR #1037, opened 2016-03-10
Moving the event brand text above the description

paulbrown1982:
Because it doesn't quite look right on the big event image:

picture 40


AWare:
:+1:

PR #1038, opened 2016-03-10
Update membership common, and make a few changes to accomodate

AWare:
@tomverran is refactoring a bunch in membership-common, and here's a change to make this compatible with either typed subscriptions. This method is up for refactoring when the new change to common drops.


joelochlann:
:+1:

PR #1042, opened 2016-03-11
Fixing missing Masterclass images

paulbrown1982:
* Reinstated the legacy scraping method of extracting eventbrite IDs from the set of masterclass content API articles. This was switched off if the article already had an eventbrite association due to the assumption that there was a 1:1 mapping between a content article and a masterclass eventbrite event.
* Still in disbelief that the fix was this straightforward! Thanks @rtyley for advising about the legacy scraping logic.

picture 42

cc @AWare @tomverran


AWare:
:+1:

PR #1043, opened 2016-03-14
Put HTTP cache in front of EventBrite API

paulbrown1982:
EventBrite API responses are occasionally taking longer than 10 seconds to return, due to the amount of data we are requesting of them, and the frequency of our calls.

We currently refresh the data in each app for 2 minutes, however the boxes often don't go healthy because they fail to get the live data in the 10 second window when they boot up. To solve this for the short term, we will put an HTTP reverse proxy cache in the way (CloudFront for now), and call it every 60 seconds to keep it warm. The TTL of the entries will be set to around 60+ seconds or so, so all-in-all, we get the same worst case e2e time for updates to appear.

* Use CloudFront URL to facade EventBrite API, with a ttl of 61 seconds.
* Request data from CloudFront more quickly so as to keep the CloudFront cache warm.
* Spread the "live","draft" and "ended" request groups out over the minute so that they're note all scheduled at once.
* Should fix that archive page doesn't contain live events too - which was due to the timeout exception breaking the pagination required to get all the old events.

https://trello.com/c/JiBUBnlo/388-the-events-archive-is-only-displaying-guardian-local-events-https-membership-theguardian-com-events-archive-it-should-show-guard

cc @tomverran @rtyley

### DIFF
--- a/frontend/app/services/EventbriteService.scala
+++ b/frontend/app/services/EventbriteService.scala
@@ -41,8 +41,8 @@ trait EventbriteService extends WebServiceHelper[EBObject, EBError] {
     req.newBuilder().url(url)
   }
 
-  def eventsTaskFor(status: String, refreshTime : FiniteDuration): ScheduledTask[Seq[RichEvent]] =
-    ScheduledTask[Seq[RichEvent]](s"Eventbrite $status events", Nil, 1.second, refreshTime) {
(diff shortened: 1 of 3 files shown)


tomverran:
assuming the cache works this looks good to me! :+1:

paulbrown1982:
We will probably want to swap it out for a Varnish node, aggregation app, or Redis-type-thing sometime soon - something to save the need for the web apps to paginate. Lots of ways to solve this problem :)

PR #1048, opened 2016-03-18
Use a BillingSchedule when showing the paid to paid upgrade summary

tomverran:
An assumption was made that the preview invoices we get back from Zuora when upgrading will consist of the following, in order:

 - refund for current plan
 - payment for new plan

As such the code took the first two invoices and added them together.
With discounts however the second invoice item is often a discount, and so we ended up adding up the refund and a discount to give a minus figure for the amount paid today:

image

Now this problem was fixed elsewhere by using a higher level `BillingSchedule` object which groups up invoice items by service start date. As such we can fix this problem by creating one from the PreviewInvoiceItems we were using and it all works perfectly:

image


AWare:
LGTM

PR #1051, opened 2016-03-18
Remove replicated touchpoint conf

rtyley:
This config is already in membership-common:

https://github.com/guardian/membership-common/tree/a109a90/conf

...and which version is going to be loaded by the config library is likely to be dependent on classpath order (ie, basically random)

sbt-assembly shows this output:

```
[error] deduplicate: different file contents found in the following:
[error] touchpoint.DEV.conf
[error] /home/rtyley/.ivy2/cache/com.gu/membership-common_2.11/jars/membership-common_2.11-0.174.jar:touchpoint.DEV.conf
[error] deduplicate: different file contents found in the following:
[error] touchpoint.PROD.conf
[error] /home/rtyley/.ivy2/cache/com.gu/membership-common_2.11/jars/membership-common_2.11-0.174.jar:touchpoint.PROD.conf
[error] deduplicate: different file contents found in the following:
[error] touchpoint.UAT.conf
[error] /home/rtyley/.ivy2/cache/com.gu/membership-common_2.11/jars/membership-common_2.11-0.174.jar:touchpoint.UAT.conf
```

cc @tomverran


tomverran:
:+1:

I take it you checked the contents of membership-common look the same!
They definitely should be

rtyley:
> I take it you checked the contents of membership-common look the same!

The ones in `membership-common` have _more_ config in, but do appear to be the same for the matching keys...

PR #1052, opened 2016-03-18
Remove all use of play-ws

rtyley:
https://github.com/mttkay/signpost/pull/14
https://github.com/playframework/playframework/issues/3365#issuecomment-198399016

cc @tomverran

### DIFF
--- a/frontend/app/filters/AddEC2InstanceHeader.scala
+++ b/frontend/app/filters/AddEC2InstanceHeader.scala
@@ -1,8 +1,9 @@
 package filters
 
-import play.api.Play.current
+import com.gu.lib.okhttpscala._
+import com.squareup.okhttp
+import com.squareup.okhttp.OkHttpClient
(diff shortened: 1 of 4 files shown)


wsargent:
:-/

rtyley:
> :-/

I really appreciate the great work that goes into Play-WS (eg ['Fixing the Most Dangerous Code in the World'](https://tersesystems.com/2014/01/13/fixing-the-most-dangerous-code-in-the-world/) to be secure by default, awesome work) but we've already moved a lot of our code to OKHttp, so standardising on that choice of HTTP library is also kind of good for us.

wsargent:
We've moved the SSL code out to https://github.com/typesafehub/ssl-config so Akka and other libraries can take advantage of it.  It returns an `SSLContext` so you can use it with whatever Java library you like.

AWare:
:+1: (We're using OKHttp in a bunch of places we don't use play.)

PR #1053, opened 2016-03-21
Apply discounts to upgrades

tomverran:
Wire up membership frontend to pass upgrades & subscribe calls through PromoService


tomverran:
@AWare

AWare:
:+1:

PR #1055, opened 2016-03-22
Fix pending amendments not being checked before previewing upgrades

tomverran:
- we didn't actually check for pending amendments before trying to preview upgrades
 - getSubscriptionStatus is going to be deleted as it is surplus to requirements


tomverran:
@paulbrown1982 would you mind taking a look at this?

To reproduce the current issue 

 - sign up as a member
 - cancel your subscription
 - attempt to change tier

paulbrown1982:
:+1:

PR #1058, opened 2016-03-22
English Heritage promo acceptance for Patron

paulbrown1982:
Fix bug where the english heritage promotion code says it's not applicable for a Patron. It's OK because we are giving the offer to all patrons/partners anyway, but it might put someone off completing if they see the error.


AWare:
:+1:

PR #1059, opened 2016-03-22
Checkout price discount promos

paulbrown1982:
* Added draft config for the new partner promotion. End date still needs to change.
* Where the promotion applies only to one rate plan (payment option) the required option is highlighted green when it is selected - this corresponds with the tick colour.
* Moved Promotion section above card so that there is no jump, and the thought flow seems much more logical.
* Tidied up the options.js accordingly to the above - these 2 sections are now more modular.
* Would still like there to be an "event bus" or to have the modules watch their models for differences, so that I don't have to import ongoingCardPayments into billingPeriodChoice (or have ongoingCardPayments bind to billingPeriodChoice's elements).
* This has been tested on all tiers and upgrade pages.
* Some copy has changed under the payment options which I'll doubly confirm with Julia.

https://trello.com/c/U7y3vQ8q/425-promo-codes-checkout-page-changes-for-price-discount-promotions
cc @tomverran @rtyley @joelochlann @AWare


frontend/app/configuration/Config.scala
```diff
@@ -20,52 +20,26 @@ import services._
 import scala.util.Try
 
 object Config {
```
paulbrown1982: Sorry, my IDE reorganised the fields and methods alphabetically!

tomverran:
So yes this is quite an overwhelmingly large set of changes but it does seem to work!

I think in the future we could have the model being a requireJS module with functions to modify the state so we can easily subscribe different things to changes. I'm not sure why I didn't think of this earlier.

Doing the above would mean we'd basically be writing our own JS framework so I think we should schedule in a spike next sprint to explore integrating _something_ that does this work for us, whether react / flux or a different technology.

:+1:

PR #1062, opened 2016-03-23
Remove promos when upgrading

tomverran:
Currently if you sign up with a promotion then upgrade it'll be applied again.
We probably don't want this to happen for discounts, so for now I have made it possible to remove promotions when upgrading.

Before release we'll want to improve this behaviour so you don't lose out if you're half way through a promotion but this should be adequate for now.


AWare:
:+1: 

I think this is a good time to stop overloading promocode as tracking code. 

Please?

PR #1065, opened 2016-03-24
Convert TierController from a load of traits into an object

tomverran:
TierController consisted of a bunch of traits:

 - DowngradeTier
 - UpgradeTier
 - CancelTier

It is the only controller we have constructed in this way and since none of these traits are tested they don't actually serve any purpose.

As such we might as well make it one object to reduce the initial what-on-earth-is-going-on-here reaction that I experience every time I look at the file

I might split out upgrades into a separate controller later today

### DIFF
--- a/frontend/app/controllers/TierController.scala
+++ b/frontend/app/controllers/TierController.scala
@@ -1,5 +1,6 @@
 package controllers
 
+import controllers._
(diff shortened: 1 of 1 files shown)


AWare:
:+1:

PR #1066, opened 2016-03-24
Remove the old EH roundel, and change index to use the new text.

AWare:
Use the new roundel on index as well as the promo page.

Third time lucky. :(


tomverran:
:+1:

PR #1054, opened 2016-03-21
50off landing

AWare:
This turned into a bit more work than intended for a landing page, so I've split out the quote changes into their own branch- which I'll drop a WIP for soon.

Roundel: Deployed  as #1060 

**Package Promo:**
-Discount pricing!
image
-A new, clean look we can use!
image
-Darker buttons!

Landing Page:
(in comment)

PR #1068, opened 2016-03-24
Use new roundel syntax

AWare:
#1066 wasn't off the latest master, and I'd changed Roundel.scala in #1054 so this wouldn't compile when merged. :(


tomverran:
fourth time lucky

tomverran:
:+1:

PR #1070, opened 2016-03-24
Fix text within Pay button on checkout page

paulbrown1982:
Fix bug where the Pay £xx button on the checkout page did not change in line with the payment options.
We could probably refactor away the horrible options.js now!

cc @rtyley @tomverran

### DIFF
--- a/frontend/app/views/fragments/pricing/billingPeriodChoice.scala.html
+++ b/frontend/app/views/fragments/pricing/billingPeriodChoice.scala.html
@@ -47,6 +47,9 @@
             generateChargeAmount: function() {
                 return formattedDiscountedCharge(checkoutForm.currency, this.amount, this.discount);
             },
+            generateDisplayAmount: function() {
+                return formattedDiscountedPrice(checkoutForm.currency, this.amount, this.discount);
+            },
             noun: '@year.noun',
             discount: 0
         },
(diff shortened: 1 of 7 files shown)


frontend/app/views/fragments/pricing/billingPeriodChoice.scala.html
```diff
@@ -47,6 +47,9 @@
             generateChargeAmount: function() {
                 return formattedDiscountedCharge(checkoutForm.currency, this.amount, this.discount);
             },
+            generateDisplayAmount: function() {
```
rtyley: There's starting to be a bit of repetition here, duplicated between monthly & annual billing. Perhaps some of it is removable?! (in a subsequent PR, not necessarily for this one)

rtyley:
I'm not the best at javascript, but this looks :+1: to me.

frontend/app/views/fragments/pricing/billingPeriodChoice.scala.html
paulbrown1982: Yeah we should remove them from the choices array too. It was there to make the template cleaner, so it doesn't have to call back up the object model to find the correct generateDisplayAmount function, it's there on the choice it's looped over.

rtyley:
cc @glockett

PR #1072, opened 2016-03-30
Build images after CSS

AWare:
One weird trick for faster frontend builds!
@joelochlann


AWare:
(But only when you have CSS problems)

joelochlann:
:+1:

PR #1075, opened 2016-04-01
Log error response from Identity

mario-galic:
@tomverran

Log the actual response from identity so we can debug the following frequent error:

[Trello card](https://trello.com/c/5j3VDDAE/451-investigate-frequent-failures-during-membership-checkout)
[Sentry event](https://app.getsentry.com/the-guardian/membership/issues/69009792/events/4096693298/)


AWare:
:+1:

PR #1083, opened 2016-04-04
Fix to social share (?)

AWare:
This should fix the image that facebook et al show.
But their parsing of our html is a little unpredictable- which we'll need to fix sometime.


joelochlann:
:+1:

PR #1091, opened 2016-04-05
We are outputting UTC rather than Europe/London currently

tomverran:
This fixes eventbrite times by using local times!


AWare:
image
+1 (hour)

PR #1092, opened 2016-04-05
Fix styling and social links on thank you page

paulbrown1982:
Tweak open graph links so the sharing of the /contribute page is correct.
Ensure the correct, trackable URLs are in the Twitter and FB share buttons on the /thank-you page.
Fix thank you page header font.

cc @AWare


AWare:
:+1:

PR #1093, opened 2016-04-05
Watch CSS, JS and images separately to speed up dev

joelochlann:
It's ridiculous waiting for JS and images to minify when you've just changed a font-size in CSS. This should speed things up.

My one concern would be that there is a reason why `clean:images` and `build:images` had to go in between the `clean:css` and `build:css` steps. But I can't think of any and it seems to work for me.

I would suggest maybe people pull in this branch and dev against it for a bit to check if there are any issues before merging?

@paulbrown1982 @AWare @tomverran

### DIFF
--- a/frontend/Gruntfile.js
+++ b/frontend/Gruntfile.js
@@ -217,9 +217,17 @@ module.exports = function (grunt) {
          ***********************************************************************/
 
         watch: {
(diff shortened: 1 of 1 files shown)


tomverran:
:+1:

PR #1095, opened 2016-04-05
Fix cross-browser issues and links in giraffe header

joelochlann:
Two things:

1. Points the links on the generic giraffe header to theguardian.com rather than membership.

2. A styling fix for the Guardian logo on Safari, IE and some mobile browsers. The height attributes on the svg element combined with `height: auto` in the CSS (which is supposed to calculate based on width & aspect ratio) were causing these browsers to give the `<svg>` a height that was too large, making it to bottom-align like this:
picture 197
Removing the attributes from the `<svg>` element and setting manual heights fixed the issue:
picture 198

@tomverran since you have grappled with this svg before :stuck_out_tongue_winking_eye:

### DIFF
--- a/frontend/app/controllers/Giraffe.scala
+++ b/frontend/app/controllers/Giraffe.scala
@@ -1,5 +1,6 @@
 package controllers
 import com.gu.stripe.Stripe
+import configuration.Config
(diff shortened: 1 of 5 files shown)


tomverran:
:+1:

PR #1096, opened 2016-04-05
Fix membership doctype

paulbrown1982:
Fixed what is clearly a copy-paste error where the lang attribute was on the doctype rather than the <html> tag. This causes the Facebook crawler not to pick up the open-graph details.

cc @joelochlann


joelochlann:
:+1: :gem: :raised_hands: :+1: :gem: :raised_hands:

PR #1101, opened 2016-04-05
Omniture amount try

AWare:
Surround the call to omniture in a try{} 
 
This lessThanFifty profile needs to be rewritten, as is now horrible.


tomverran:
:+1:

PR #1105, opened 2016-04-06
Fix the card on the paid to paid upgrade summary always being amex

tomverran:
Card type is capitalised now but the CSS classes are not...


AWare:
👍

PR #1106, opened 2016-04-06
Put giraffe bottom text inside the grey container

joelochlann:
#### before
picture 200

#### after
picture 199

@tomverran


tomverran:
:+1:

PR #1108, opened 2016-04-06
Update libsass and add jasmine-core

AWare:
karma-jasmine used to provide jasmine-core- and no longer does; this causes some errors on dev when npm installing, must have been provided transitivly/unused

libsass 1.0.0 had some sourcemap issues, so bumping to latest stable version


tomverran:
:+1:

PR #1109, opened 2016-04-06
[WIP] New quote

AWare:
This breaks the patron page, which needs rework anyway...

Old patron*s*
image
New (broken):
image

Old int Supporter:
image
New:
image

Old US:
image

New US:
image
 
TODO:
Fix the patrons page.
Fix top padding.


AWare:
Once we have a solution to this breaking the patrons page, I'll get full screenshots and would like to get this pushed. https://github.com/guardian/membership-frontend/pull/1113

PR #1115, opened 2016-04-07
Update Selenium to 2.52.0

mario-galic:
@AWare 

See https://github.com/guardian/subscriptions-frontend/pull/410


AWare:
:+1: As before, will need to review when we upgrade.

PR #1117, opened 2016-04-07
Fix SBT test command semantics

mario-galic:
@AWare @tomverran 

Semantics of `sbt test` is to execute all tests [by design](http://www.scala-sbt.org/0.13/docs/Testing.html), and should not be redefined by us. To run just unit tests execute `sbt fast-test` in the same way as we do for [subscription-frontend](https://github.com/guardian/subscriptions-frontend/blob/065e56a3f2dc053a63558cd4ee647021214378b7/build.sbt#L76).

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,11 +125,6 @@ Note that we read the shared secret for these from the `identity.test.users.secr
 
 ### Automated
 
-
-#### Scala unit tests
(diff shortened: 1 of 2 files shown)


AWare:
👍

PR #1119, opened 2016-04-08
Add australia contribute and thank you pages

joelochlann:
We now have three copy-pasted variants of the contribute & thank you pages. This is because the three variants are still subject to unpredictable change at very short notice and we need the flexibility to be able to quickly modify any part of them independently. Once the dust is settled and we have a sense of what probably will and won't vary between countries, we'll refactor to eliminate this duplication.

##### Contribute page
* Aussie dollars instead of pounds
* Marketing preferences unticked

##### Thank you page
* Links to Apple App Store and Guardian Daily Email updated to be Australia-specific

### What's likely to change?
* Terms & Conditions link (currently gives 404 since it's just a guess as to what the URL from Central Production might be)
* Disclaimer text at the bottom

screencapture-mem-thegulocal-com-au-contribute-1460113430025


AWare:
👍 🇦🇺

joelochlann:
giphy

PR #1120, opened 2016-04-08
Disambiguate aussie dollars

joelochlann:
Graham suggested making it clear that it's Australian dollars. According to [wikipedia](https://en.wikipedia.org/wiki/Australian_dollar) A$ is a standard symbol for distinguishing Australian dollars from other dollar currencies.

#### before
picture 208

#### after
picture 209


joelochlann:
Apparently we've decided against this

> given this is being targeted to Australian readers, there is no need to include the A$ to distinguish Australian dollars

PR #1122, opened 2016-04-11
Fix so that billing address correctly changes displayed currency

paulbrown1982:
Fixed a bug where changing the country of a billing address did not change the payment options display, and worse the CTA. It did change the Ongoing payments section and actual backend billing, just fine as it always has.

I also moved the paymentOptions to being below the billing address, so that the render change does not happen above the customer's viewport. This makes it very clear that the change of billing address has changed the charged currency.

The promotion validity remains determined by the delivery country.

cc @AWare @tomverran


tomverran:
👍

PR #1123, opened 2016-04-11
endpoint to preview upgrades with promo codes

tomverran:
So what I had to do to get this to work is the following

 - Split the _creation_ of a PaymentSummary out from the display of the preview page
 - Add a JSON version of the preview page that calls the above method then JSONifies it

Then to pass a promo code through to the preview API call I basically had to do the same thing:

 - Split the creation of the Amend command out from the actual sending of it to Zuora
 - Call that method, then set preview mode to true on the resulting Amend before running it


AWare:
👍

PR #1125, opened 2016-04-12
Fix promotion text on thank you page when no description

paulbrown1982:
Fix that the promotion text is not presented on the thank you page if it has no description - solution is to show the promotion title (or headline) if there's no description.

cc @AWare


AWare:
:+1:

PR #1126, opened 2016-04-12
four logos enter, one logo leaves

AWare:


rtyley:
:+1: will be interesting to see if anyone complains...

PR #1129, opened 2016-04-12
Redirect invalid promotion page requests

paulbrown1982:
Just like with subs, we should redirect non-existing or expired promotion pages to the homepage, along with a campaign code for tracking.

https://trello.com/c/FRqt0kpa/448-add-redirect-for-https-membership-theguardian-com-p-eh2016

cc @AWare


AWare:
Would prefer to redirect to one of our landing pages: 👍

PR #1130, opened 2016-04-13
Log auth context when getFullUserDetails fails

mario-galic:
@tomverran @AWare 

Temporarily enabled to get more info around the following error: [Trello](https://trello.com/c/FC6n1b5o/475-users-cannot-join-paid-membership)


AWare:
:+1:

PR #1131, opened 2016-04-13
Ability to load ES6 modules with requirejs-babel

tomverran:
This allows you to write ES6 modules and import them with requirejs by prefixing their path with es6!

i.e.

```javascript
// filepath: my/es6/module.js
import * as template from 'lodash/string/template'
import * as ajax from 'ajax'

export function init() {
    //nice es6 code
}
```

```javascript
// filepath: main.js
require(['es6!my/es6/module'], function(module)) {
    module.init()
}
```

RequireJS then uses the es6 module loader to resolve the ES6 modules when compiling all the modules into a single file. Because the babelising is a compile time step both the es6 loader and babel then get stubbed out in the resulting JS bundle to stop it becoming massive.

I'm going to do the frontend for paid to paid upgrades as ES6 but I thought I'd make this a separate PR so we can ensure nothing breaks

Note that we have no polyfills yet, this only allows ES6 syntax for modules etc.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -5,6 +5,10 @@
         "browser": true,
         "node": true
     },
+    "parserOptions": {
+        "ecmaVersion": 6,
+        "sourceType": "module",
+    },
     "globals"   : {
         "guardian": true
     },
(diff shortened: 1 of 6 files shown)


joelochlann:
:raised_hands: :gem: :100: :+1:

PR #1133, opened 2016-04-13
Send updated user details to identity

AWare:
Update address details held in identity when a friend becomes a really good friend.


paulbrown1982:
👍

PR #1134, opened 2016-04-13
Tracking promotions

paulbrown1982:
- Visiting https://membership.theguardian.com/p/TRACK01 will redirect to the homepage (setting an INTCMP query string to track in GA/Omniture) and TRACK01 will be stored in the session.
- When the user arrives on the checkout or upgrade form the hidden trackingPromoCode field will be filled with the session value.
- The main promo code field will remain blank and can be used as normal - any valid promo code added in that field overrides the trackingPromoCode when the subscription is made.

https://trello.com/c/3sMxyCLI/485-tracking-codes-for-membership

cc @tomverran

### DIFF
--- a/frontend/app/controllers/Promotions.scala
+++ b/frontend/app/controllers/Promotions.scala
@@ -102,6 +102,7 @@ object Promotions extends Controller {
       for {
         promotion <- promoService.findPromotion(promoCode) \/> notFound //if we can't find the promotion fail out of the for comprehension with a notFound
         _ <- failWhen(promoCodeStr.toUpperCase != promoCodeStr,redirectToUpperCase)
+        _ <- failWhen(promotion.starts.isAfterNow,notFound)
         _ <- failWhen(promotion.expires.isBeforeNow,notFound)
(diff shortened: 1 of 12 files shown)


philwills:
Anything with a `cmp` or `intcmp` parameter will also be tracked as having that campaign code in Ophan dashboard and lake.

frontend/app/controllers/Promotions.scala
```diff
@@ -102,7 +104,9 @@ object Promotions extends Controller {
       for {
         promotion <- promoService.findPromotion(promoCode) \/> notFound //if we can't find the promotion fail out of the for comprehension with a notFound
         _ <- failWhen(promoCodeStr.toUpperCase != promoCodeStr,redirectToUpperCase)
+        _ <- failWhen(promotion.starts.isAfterNow,notFound)
         _ <- failWhen(promotion.expires.isBeforeNow,notFound)
+        _ <- failWhen(promotion.whenTracking.isDefined,redirectToHomePage)
```
philwills: Am I right that this uses `failWhen` for a valid behaviour? If so that feels a bit odd.

tomverran:
👍

PR #1135, opened 2016-04-14
Update content-api-client to v8.5

rtyley:
This was kind of prompted by noting that [when Frontend upgraded their content-api-dependency, it broke a small bit of Member's Area functionality](https://github.com/guardian/frontend/pull/11466#commitcomment-17085986).

When a non-member tries to access a piece of Member content, the redirect is sending them to:

https://membership.theguardian.com/choose-tier?membershipAccess=MembersOnly

rather than:

https://membership.theguardian.com/choose-tier?membershipAccess=members-only

...because "MembersOnly" is the `name` value of `com.gu.contentapi.client.model.v1.MembershipTier.MembersOnly` (note that this class is generated from [the master Thrift file](https://github.com/guardian/content-api-models/blob/caaa8ce0a/src/main/thrift/content/v1.thrift#L177))

I could have just changed [the strings that we match on](https://github.com/guardian/membership-frontend/blob/9b031888/frontend/app/model/MembershipAccess.scala#L4-L5), but I thought _maybe_ it would be better to use the content-API values.

cc @tomverran

### DIFF
--- a/frontend/app/controllers/Info.scala
+++ b/frontend/app/controllers/Info.scala
@@ -228,7 +228,7 @@ trait Info extends Controller {
 
     val results =
       GuardianContentService.offersAndCompetitionsContent.map(ContentItemOffer).filter(item =>
-        item.content.fields.map(_("membershipAccess")).isEmpty && ! item.content.webTitle.startsWith("EXPIRED") && item.imgOpt.nonEmpty)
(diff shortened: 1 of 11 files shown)


tomverran:
👍

philwills:
Burn the hideous strings with righteous fire!

rtyley:
Verified on PROD...

Offers and Comps articles still found:
image

Masterclasses articles still found:
image

Linked articles still found:
image

PR #1136, opened 2016-04-14
Fix display of members-content signup message

rtyley:
We want to let users know why they're seeing a 'Join Membership' page when they click on members-only content, like this:

image

This got broke a little while ago: https://github.com/guardian/frontend/pull/11466#commitcomment-17085986 ...and this new commit Membership-Frontend to read the new value that dotcom-Frontend is sending us, ie:

https://membership.theguardian.com/choose-tier?membershipAccess=MembersOnly

Sidenote: I feel that the `MembershipTier` class is rather misleadingly named for us (because it doesn't actually supply the Membership tier, which would be 'Partner', 'Friend', etc - instead it supplies`MembersOnly` / `PaidMembersOnly`, so I've used an import rename of `MembershipTier=>ContentAccess` to make it's purpose clearer in our codebase.

This PR follows on from https://github.com/guardian/membership-frontend/pull/1135

cc @tomverran

### DIFF
--- a/frontend/app/controllers/Joiner.scala
+++ b/frontend/app/controllers/Joiner.scala
@@ -3,6 +3,7 @@ package controllers
 import actions.ActionRefiners._
 import actions.{RichAuthRequest, _}
 import com.github.nscala_time.time.Imports._
+import com.gu.contentapi.client.model.v1.{MembershipTier=>ContentAccess}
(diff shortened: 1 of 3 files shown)


paulbrown1982:
👍

rtyley:
Verified on PROD

image

PR #1137, opened 2016-04-14
Refactor giraffe.js to use ES6 and be shorter

joelochlann:
Can be considered both a small step in migrating to ES6, and part of post-Giraffe cleanup to pave the way for future one-off donations work.

@AWare


AWare:
www.jquery.org

AWare:
This shows just how verbose the DOM apis are, but we really shouldn't use any more bonzo. 
👍

joelochlann:
Yeah. `bonzo` to `jQuery should be a simple enough switch (I think the `$` wrapper that's `bonzo` + `qwery` gives basically a subset of jQuery). If we want to ditch `bean` and `reqwest` that's a bit more work

PR #1140, opened 2016-04-14
Fix inconsistencies between different giraffe pages

joelochlann:
In #1137 I removed the `hidden` attribute from the currency selector on the UK page by mistake. It should still be hidden.

I also added some classes to the buttons & button parents which are now necessary for "option button" functionality (a group of buttons which function like radio buttons). I only added these to the UK page - we need them for other countries too.

@AWare

### DIFF
--- a/frontend/app/views/giraffe/contribute.scala.html
+++ b/frontend/app/views/giraffe/contribute.scala.html
@@ -60,7 +60,7 @@
             <div class="form__column">
                 <h3 class="form__column-heading headline-text--bolder"><span class="num">2</span> Your amount</h3>
                 <ul class="u-unstyled">
-                    <li id="currency_field" class="form-field">
+                    <li id="currency_field" class="form-field" hidden>
                         <label for="currency" class="label u-h">Currency</label>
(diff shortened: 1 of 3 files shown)


frontend/app/views/giraffe/contribute.scala.html
```diff
@@ -60,7 +60,7 @@ <h3 class="form__column-heading form__column-heading--first headline-text--bolde
             <div class="form__column">
                 <h3 class="form__column-heading headline-text--bolder"><span class="num">2</span> Your amount</h3>
                 <ul class="u-unstyled">
-                    <li id="currency_field" class="form-field">
+                    <li id="currency_field" class="form-field" hidden>
```
paulbrown1982: Shouldn't hidden go inside the class attribute?

joelochlann: yes, if we can imagine it being turned on/off via JavaScript - but this was simply a shortcut to "disable" this functionality altogether so I think it's fine as an attribute here

PR #1047, opened 2016-03-18
Update sbt 0.13.9

rtyley:
I don't know why sbt 0.13.11 is failing (as with https://github.com/guardian/membership-frontend/pull/1028 & https://github.com/guardian/membership-frontend/pull/1045) - the previous version of sbt was 0.13.9, so updating to that.

All versions of sbt work fine for me when executing the packaged artifact locally on my Ubuntu dev box:

```
cd ~/development/membership-frontend
sbt clean play-artifact
git clean -fd
unzip /home/rtyley/development/membership-frontend/frontend/target/artifacts.zip packages/frontend/app.zip
unzip packages/frontend/app.zip
cd ~/development/membership-frontend/frontend-1.0-SNAPSHOT
bin/frontend -mem 1024 -Dconfig.resource=dev.conf
```

PR #1143, opened 2016-04-15
Add sbt-assembly to run as a CI task to identify classpath clashes

rtyley:
This is related to the various issues we've had recently that have been due to classpath conflicts - several of these issues (eg https://github.com/guardian/membership-frontend/pull/1008 and https://github.com/guardian/membership-frontend/pull/1047#issuecomment-198331201) could have been detected by running `sbt assembly`, which fails the build if there are classpath conflicts. It does make build time a little longer, and I'd like to stress that we_don't deploy_ the resulting uber-jar - the deployable is still the `native-packager` zip of library jars, we just build the uber-jar with `assembly` to check that we _can_ do so without clashes.

In order for `sbt assembly` to work, this change addresses all the classpath clashes that currently occur in our build - normally by excluding the unwanted ones:

https://github.com/sbt/sbt-assembly#exclude-specific-transitive-deps

There's a nasty classpath clash brought in by Play-WS, but unfortunately we can't exclude that dependency just yet... https://github.com/guardian/membership-frontend/pull/1052

Note that the sbt `assembly` task also runs `test`.

One issue that `assembly` would not have been able to detect would have been the dependency diamond problem we had recently with membership-common.

@AWare  @blackyjnz

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,4 +13,4 @@ before_cache:
   - find $HOME/.ivy2/cache -name "ivydata-*.properties" -type f -delete -print
 before_install:
   - export TZ=Europe/London
-script: sbt ++$TRAVIS_SCALA_VERSION test
+script: sbt ++$TRAVIS_SCALA_VERSION assembly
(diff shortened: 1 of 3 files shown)


AWare:
👍

PR #1144, opened 2016-04-15
Tracking promo code on friend signup

paulbrown1982:
Added the trackingPromoCode field to the Friend signup form, which gets filled from the promoCode stored in the session, iff it's a code for a Tracking-only promotion.

cc @tomverran @AWare


tomverran:
👍

PR #1145, opened 2016-04-15
"fix" requirejs timeout

AWare:
Babel is ever so slightly slow, and require is ever so slightly impatient. 

Make it wait ages until @tomverran's got webpack working.


paulbrown1982:
👍

PR #1147, opened 2016-04-19
Don't cancel the free subscription until the paid one has succeeded.

AWare:
Log when we get an error that's misparsed as a success.

Only delete the free sub once the paid one has been created.


rtyley:
Sounds fair enough :+1:

PR #1148, opened 2016-04-19
Update to the latest Membership Common version.

paulbrown1982:
Handle the fact that the country inside memcommon's Address case class has been changed to an option. In most cases we fall back to the Address.countryName string, which the optional Address.country is derived from anyway.


AWare:
:+1:

PR #1150, opened 2016-04-20
Log free to paid upgrades as before

AWare:
I've changed the analytics for free to paid to be the same as what we did before. 
Will tidy up once we've okayed this approach. :)


tomverran:
👍

PR #1151, opened 2016-04-21
show currency names in payment chooser

AWare:
image
image
image

Easiest fix to https://trello.com/c/762rvdhe/484-specifiy-ca-whenever-we-refer-to-the-canadian-dollar-price-on-our-site otherwise, we'll need to take a closer look at how we display currencies.


paulbrown1982:
👍

PR #1152, opened 2016-04-21
Only respect Identity Cookie for authentication, use Token for name only

rtyley:
We seem to be getting the Mobile apps hitting us with only an Identity token, and no valid Identity cookie.

This change means that, at the very least, those users will be prompted to sign in when they try to hit the checkout page, rather than an error.

https://github.com/guardian/identity-play-auth/commit/e716abb0549810d87851839f0f616e6e7997a44b

cc @tomverran @mario-galic @DiegoVazquezNanini @alfavata

### DIFF
--- a/frontend/app/services/AuthenticationService.scala
+++ b/frontend/app/services/AuthenticationService.scala
@@ -1,17 +1,15 @@
 package services
 
-import com.gu.identity.play.{AccessCredentials, AuthenticatedIdUser}
+import com.gu.identity.play.AccessCredentials.{Cookies, Token}
+import com.gu.identity.play.AuthenticatedIdUser.Provider
 import configuration.Config
-import play.api.mvc.RequestHeader
 
 object AuthenticationService extends com.gu.identity.play.AuthenticationService {
   def idWebAppSigninUrl(returnUrl: String) = Config.idWebAppSigninUrl(returnUrl)
 
   val identityKeys = Config.idKeys
 
-  override lazy val authenticatedIdUserProvider: (RequestHeader) => Option[AuthenticatedIdUser] = AuthenticatedIdUser.provider(
-    AccessCredentials.Cookies.authProvider(identityKeys),
-    AccessCredentials.Token.authProvider(identityKeys, "membership")
-  )
+  override lazy val authenticatedIdUserProvider: Provider =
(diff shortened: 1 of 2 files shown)


DiegoVazquezNanini:
I am going to ask the client side devs why this happens.
:+1:  Thanks @rtyley

alfavata:
@DiegoVazquezNanini we can have a look together but @mario-galic and I were looking at our logs yesterday and everything seemed in order on our side.

DiegoVazquezNanini:
@alfavata sure. @rtyley do you know which app is doing the calls without the cookies? 
Android? Ios? and do we know the version ? I should be on the User-Agent header.

mario-galic:
@DiegoVazquezNanini It seems only iOS is doing the calls without the cookies. Here are few examples:

```
SC_GU_U=false GU-IdentityToken=true trackingParamters=List((trackingUserAgent,Mozilla/5.0 (iPhone; CPU iPhone OS 9_3_1 like Mac OS X) AppleWebKit/601.1.46 (KHTML, like Gecko) Mobile/13E238)

SC_GU_U=false GU-IdentityToken=true trackingParamters=List((trackingUserAgent,Mozilla/5.0 (iPhone; CPU iPhone OS 9_3 like Mac OS X) AppleWebKit/601.1.46 (KHTML, like Gecko) Mobile/13E230)

SC_GU_U=false GU-IdentityToken=true trackingParamters=List((trackingUserAgent,Mozilla/5.0 (iPad; CPU OS 8_4_1 like Mac OS X) AppleWebKit/600.1.4 (KHTML, like Gecko) Mobile/12H321)
```

alfavata:
We only load that web view once the cookies **and** the token are received. The only thing I can think of would be adding a check if the `cookies` object we receive actually contains any value before continuing.

rtyley:
This is deployed to PROD https://riffraff.gutools.co.uk/deployment/view/26d24a34-e594-4d61-836a-c883f016cdf9# - looks good to me in the web. @alfavata, can you validate that the flow still works in app?

alfavata:
It seems so, @rtyley, but I'd still be keen on adding the check I mentioned and some more logging (working with @mario-galic on the latter).

alfavata:
Actually, that check is already in place (just in a different part of the codebase). I really have no idea what was causing that error 😭

PR #1155, opened 2016-04-26
Fix sign out return url

jranks123:
.Previously, if you signed out from membership you were redirected
to the home page. This seems to be due to a missing negation in
the sign out url builder, namely that instead of not adding
a return URL to anything namedin a sequence of excpetions,
it was not adding the return URL to anything not named in the
sequence, meaning all but the welcome page


tomverran:
👍

PR #1156, opened 2016-04-26
Update Salesforce tier on free to paid upgrades

rtyley:
Following on from guardian/members-data-api#95 ("Get accurate Membership status from Zuora on webhook update") we noticed that we were not updating Salesforce with the new Tier when doing a
free-to-paid upgrade (we _do_ do it for all other join/upgrade paths that I can see). This seemed to stop the Salesforce outgoing webhook from firing, even though I _think_ Z360 should have taken care of that.

So this change makes the free-to-paid upgrade fall into line with the other code paths.

Note that it is really sad that we can't just have this bit of code in *one* place:

```
salesforceService.updateMemberStatus(idUser, tier, customer)
```

...but that's due for a bigger refactor.

cc @tomverran @AWare

### DIFF
--- a/frontend/app/services/MemberService.scala
+++ b/frontend/app/services/MemberService.scala
@@ -187,7 +187,6 @@ class MemberService(identityService: IdentityService,
         contact = sub.contact,
         planChoice = PaidPlanChoice(newTier, sub.subscription.plan.billingPeriod),
         form = form,
-        customerOpt = None,
         campaignCode = code
       ))
     } yield {
@@ -476,7 +475,6 @@ class MemberService(identityService: IdentityService,
                                   contact: Contact,
                                   planChoice: PlanChoice,
                                   form: MemberChangeForm,
-                                  customerOpt: Option[Customer],
                                   campaignCode: Option[CampaignCode])(implicit r: IdentityRequest): Future[MemberError \/ ContactId] = {
 
     val addressDetails = form.addressDetails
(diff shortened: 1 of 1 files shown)


frontend/app/services/MemberService.scala
```diff
@@ -475,7 +475,6 @@ class MemberService(identityService: IdentityService,
                                   contact: Contact,
                                   planChoice: PlanChoice,
                                   form: MemberChangeForm,
-                                  customerOpt: Option[Customer],
```
rtyley: This argument was always `None`, so removing it.

AWare:
:+1:

rtyley:
Blast, looks like our Zuora code is resolving Zuora subscriptions in a bad way when we have a same day join-as-friend-then-upgrade-to-supporter.

I joined as a Friend, then immediately upgraded to Supporter:

image

But woe, this was recorded in the logs:

```
2016-04-26/i-d43e5e5e/membership-attribute-service.log
2016-04-26 15:35:55,450 [application-akka.actor.default-dispatcher-15] INFO controllers.SalesforceHookController - Differing tier info for Contact(16158552,...) : webhook=Supporter zuora=Friend
```

Our Zuora code is saying the user is a friend - and we're trusting it - while the SF webhook correctly (because we've _told_ SF) says the user is a Supporter.

Damn. cc @tomverran

tomverran:
https://github.com/guardian/membership-common/commit/154d2ce621c6136b5b04b2e331ed3951b04381a0 fixed the parsing of multiple Zuora subscriptions so a membership-common update should sort this out, shall do now

PR #1159, opened 2016-04-27
Webpack!

AWare:
This is all the handiwork of @tomverran 

New features: 
-Webpack instead of require
-NPM instead of bower (for js)
-Clean asset map and recompile on watch
-Future async loading to replace curl


AWare:
Needs a new node install. 

Will be installing nvm

AWare:
Teamcity server updated.

paulbrown1982:
I get a JS error in comparisonTable.es6, the $ variable is an object not a function, $.default appears to be the function that does the selector.

paulbrown1982:
Everything else works just fine, including contribute.

jranks123:
screenshot at may 10 12-24-10

The columns of this table don't appear to be populated

frontend/assets/javascripts/src/modules/comparisonTable.es6
```diff
@@ -28,6 +28,7 @@ function tierSelector(tier) {
 // Instead they are split between rows, so we use data-tier to group them together.
 function addHoverListeners() {
     bean.on(COMPARISON_TABLE, 'mouseenter', CLICKABLE, e => {
+        console.log($);
```
paulbrown1982: --

AWare:
The build will fail if a non webpack build has been run on that agent, and will need this to be checked. 
image

tomverran:
👍

PR #1160, opened 2016-04-28
Only prefill promo code when relevant

paulbrown1982:
Use the promoService.validate to pre-validate a promo code before displaying a checkout page. This stops a Partner-only promo from appearing [with an invalid warning] on a Supporter checkout page.

https://trello.com/c/ar6GxFSu/501-bugfix-only-prefill-promo-code-when-valid-on-join-upgrade-page

cc @tomverran

### DIFF
--- a/frontend/app/controllers/Joiner.scala
+++ b/frontend/app/controllers/Joiner.scala
@@ -6,7 +6,8 @@ import com.github.nscala_time.time.Imports._
 import com.gu.contentapi.client.model.v1.{MembershipTier=>ContentAccess}
 import com.gu.i18n.CountryGroup.UK
 import com.gu.i18n.{CountryGroup, GBP}
-import com.gu.memsub.promo.{Tracking, PromoCode}
+import com.gu.memsub.BillingPeriod
+import com.gu.memsub.promo.{NewUsers, PromoCode}
 import com.gu.salesforce._
 import com.gu.stripe.Stripe
 import com.gu.stripe.Stripe.Serializer._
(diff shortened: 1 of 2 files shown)


frontend/app/controllers/TierController.scala
```diff
@@ -156,7 +156,7 @@ object TierController extends Controller with ActivityTracking
 
     // Preselect the country from Identity fields
     // but the currency from Zuora account
-    def pageInfo(pf: PrivateFields, billingPeriod: BillingPeriod): PageInfo = {
+    def getPageInfo(pf: PrivateFields, billingPeriod: BillingPeriod): PageInfo = {
       val selectedCountry = pf.billingCountry.orElse(pf.country).flatMap { name =>
         CountryGroup.countries.find(_.name == name)
```
tomverran: If this is being used by multiple controllers I'd put it in some function somewhere common rather than having controllers depend on other controllers, maybe in the views.support.PageInfo?

frontend/app/controllers/Joiner.scala
```diff
@@ -6,14 +6,16 @@ import com.github.nscala_time.time.Imports._
...
+import com.gu.memsub.BillingPeriod
+import com.gu.memsub.promo.{NewUsers, PromoCode}
 import com.gu.salesforce._
 import com.gu.stripe.Stripe
 import com.gu.stripe.Stripe.Serializer._
 import com.gu.zuora.soap.models.errors._
 import com.netaporter.uri.dsl._
 import com.typesafe.scalalogging.LazyLogging
 import configuration.{Config, CopyConfig}
+import controllers.TierController._
```
paulbrown1982: This shouldn't have happened. Will see why IntelliJ imported it.

frontend/app/controllers/TierController.scala
paulbrown1982: It's not used by any other class.

tomverran:
So yes I'd be happier if we were able to reduce the code duplication involved in getting valid tracking / non tracking promotions but that can be done later 👍

PR #1162, opened 2016-04-29
Fix giraffe cap

AWare:
Returns cap to 500.
Longer input fields. 
Fix to js bug noticed by @paulbrown1982 
Obey cap in stripe code.


markjamesbutler:
LGTM 👍

PR #1163, opened 2016-05-03
Only have one giraffe template

AWare:
I'm not happy with the mix of approaches I took here, I'd like to talk it over with someone. This works, but briefly displays all content to all users.


AWare:
@joelochlann Took css based approach we discussed, much nicer now. 

@markjamesbutler

frontend/app/views/giraffe/contribute.scala.html
```diff
@@ -225,6 +230,30 @@ <h3 class="cvc__heading">American Express</h3>
...
+        </span>
+        <span data-shown="usd">
+            <p>If you have any questions about contributing to the Guardian, please
+                <a href="mailto:contribution.support@@theguardian.com" class="support-thanks__link">contact us here</a>.
+            </p>
+            <p>
+                The ultimate owner of the Guardian is The Scott Trust Limited, whose role it is to secure the editorial and financial independence of the Guardian in perpetuity. Reader contributions support the Guardian’s journalism.
+            </p>
+            <p>
+                Please note that the Guardian is not a charity, will not use your support for charitable
```
markjamesbutler: Could this be pushed into PageInfo as an additional attribute thus removing the duplication here?

frontend/app/views/giraffe/contribute.scala.html
```diff
@@ -207,7 +208,10 @@ <h3 class="cvc__heading">American Express</h3>
 
                     <div class="fieldset__note">
                         <p>By proceeding, you are agreeing to our
-                            <a href="@Links.giraffeTerms" class="text-link" target="_blank">Terms and Conditions</a> and
+                            <a href="@Links.giraffeTerms" class="text-link" target="_blank" data-shown="gbp">Terms and Conditions</a>
+                            <a href="@Links.giraffeTermsAustralia" class="text-link" target="_blank" data-shown="aud">Terms and Conditions</a>
```
markjamesbutler: It should be possible to pass this into the view rather than selecting it in the template using CSS etc.  I.e. Add to PageInfo etc for given country.

markjamesbutler:
LGTM given refactoring chat and fixing of build.

PR #1166, opened 2016-05-04
Set the optimizely flag to false

AWare:
Optimizely is massive (~70kb), and we don't seem to have any tests running with them at the moment. 

- Do we actually have any tests set up?
- Do we want to keep using them for A/B tests?
- Is it right that optimizely only runs on prod?

@blackyjnz


markjamesbutler:
👍

AWare:
Chatted with @blackyjnz, no current tests. Turning off for now, will delete if we don't use it.

PR #1170, opened 2016-05-11
Update Raven

AWare:
Update raven and turn sourcemaps on.
Should provide useful sentry errors. 
Will check sourcemap with https://sourcemaps.io/

Using local raven instead of globals everywhere.


mario-galic:
👍

PR #1171, opened 2016-05-11
Update facebook tracking account for Nick M Miller

philwills:
See https://trello.com/c/RfGPOR2c/4-facebook-pixels-on-supporter-landing-pages

The Trello card talks about a new tracking pixel, but as far as I can tell, it's identical to what is already included.


tomverran:
👍 assuming that new number is correct!

paulbrown1982:
👍 too assuming the old number is not used by anyone.

PR #1172, opened 2016-05-11
More CORS headers on AJAX calls

paulbrown1982:
Fixing the CORS header on our AJAX calls so that www.theguardian.com/membership can go HTTPS.

cc @tomverran


tomverran:
👍

PR #1174, opened 2016-05-12
Use correct raven module

paulbrown1982:
- Ensure our raven.js module uses the webpack-provided raven-js dependency.
- Remove non-installed aliased (bower) version of raven and ensure our modules are using our wrapped raven module instead.

cc @AWare @tomverran


tomverran:
👍 Thanks for the verbal explanation!

So the summary is

 - the alias 'raven' mistakenly pointed to the actual raven object, not the wrapper we have
 - Since loading commonJS modules is so easy there's no point in aliasing actual raven at all

PR #1176, opened 2016-05-13
Giraffe uat switch

jranks123:
This pull request turns contribute into a non-cached page which will react in the following way. 

If the user is not logged in, the real touchpoint backend is used
If the user is logged in and is not a test user, the real touchpoint backend is used
If the user is logged in and is a test user, the test touchpoint backend is used (this is shown visually on the pay button, which has (UAT) appended to the text)

screenshot at may 13 17-14-00

### DIFF
--- a/frontend/app/controllers/Giraffe.scala
+++ b/frontend/app/controllers/Giraffe.scala
@@ -23,19 +23,16 @@ object Giraffe extends Controller {
     Facebook("http://www.theguardian.com/news/series/panama-papers?CMP=fb_contribute")
   )
 
-  val stripe = TouchpointBackend.Normal.giraffeStripeService
-  val identity = TouchpointBackend.Normal.identityService
+
   val chargeId = "charge_id"
   val maxAmount: Option[Int] = 500.some
-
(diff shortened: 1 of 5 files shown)


AWare:
Let's display a warning when we're UATing.

frontend/app/controllers/Giraffe.scala
```diff
@@ -41,10 +42,12 @@ object Giraffe extends Controller {
       navigation = Seq.empty,
       customSignInUrl = Some((Config.idWebAppUrl / "signin") ? ("skipConfirmation" -> "true"))
     )
-    Ok(views.html.giraffe.contribute(pageInfo,maxAmount,CountryGroup.UK))
+    Ok(views.html.giraffe.contribute(pageInfo,maxAmount,CountryGroup.UK,isUAT))
   }
 
-  def contributeUSA = CachedAction { implicit request =>
```
AWare: I think we're at the point where we need to consider the viability of three separate, identical controllers. (But not as part of this PR)

AWare:
👍

AWare:
Seen in prod. :)

PR #1178, opened 2016-05-16
split giraffe and normal js

AWare:
Had some real problems when both of these were es6 modules, not sure where they crept in from. 

Going to attempt to strip giraffeMain down further before deciding what's common.


tomverran:
👍

PR #1180, opened 2016-05-16
Refactored Giraffe thank you controllers

jranks123:
- removes seperate scala.html pages for UK, US and AU thank you pages
- refactors thank you controller logic so it's all in one function


AWare:
👍

PR #1186, opened 2016-05-19
Genericise promo landing pages

paulbrown1982:
- Genericised the old Partner 99 Discount and English Heritage Incentive landing pages into generic versions for those two promotion types.
- There's a UX constraint that we can't explain what's going on when yearly rate plans have a durationMonths discount. Basically for Zuora we divide the discount percentage by 12/durationMonths, then apply that to the year price.
- Still got to update the index.scala.html page to use the section template for its sections in order to gain better code reuse. Also this will force us to template each of those section listings and makes it easier to use on the promotion pages if needed (like I did with the video)
- More landing page tweaks and ensured time-limited discount promotions calculate correctly for annual terms.

cc @tomverran


tomverran:
This looks okay to me 👍

PR #1190, opened 2016-05-23
fix giraffe cvc cta

AWare:
Whoops, over zealous with the js lib removal. Readd toggle.


philwills:
Despite very little context, I'm pretty confident this is A-OK.

AWare:
Thanks Phil

PR #1191, opened 2016-05-24
replace full name field with first and last name fields

jranks123:
replace full name field with first and last name fields, and also fix contribute amounts

screenshot at may 24 08-07-52
screenshot at may 24 08-08-18


AWare:
Can you post some screenshots please?
Looks fine, if the name is going into stripe okay then it just needs a rebase to get rid of the merge commit.

AWare:
👎  Data doesn't get to stripe and prefill from identity is broken.

PR #1192, opened 2016-05-24
remove explicit panama sharing section from thankyou page

jranks123:
remove explicit panama sharing section from thankyou page

screenshot at may 24 08-07-06


AWare:
I've got new copy for this, have emailed it over.

PR #1197, opened 2016-05-24
Rename logback config file to what Play uses

mario-galic:
@tomverran 

[Play expects logback.xml](https://www.playframework.com/documentation/2.4.x/SettingsLogger#Using-a-configuration-file-from-project-source)


nlindblad:
👍

PR #1200, opened 2016-05-25
Landing page for tracking-only promo codes

paulbrown1982:
- Show a single tier landing page for tracking promotions
- Rename the discountLandingPage to singlePricePlanDiscountLandingPage
- Fixed bug with wrong CTA buttons showing
- Removed the non-defined "blockquote--feature" class

cc @tomverran


tomverran:
👍

PR #1205, opened 2016-05-25
Update FB share image to new Support The Guardian image

jranks123:
screenshot at may 25 16-04-40


AWare:
Again, we need to look at the non UK user experience for this, but that's a bigger job. 
👍

jranks123:
@AWare Can you explain what you mean?

PR #1206, opened 2016-05-25
update ophan.js to use an amd module

AWare:
Depends on https://github.com/guardian/ophan/pull/1480 
This will let us call ophan's record function in our code, albeit as:
`            ophan.ophan.then(function(ophan){ 
                ophan.record(???)}); 
`
(NB, this has four wheels so is stable ;) )

@joelochlann @philwills


philwills:
Might it make sense to expose `record` rather than `ophan`?

AWare:
I'm not exposing ophan, I'm exposing curls ophan promise. But agreed, it's inelegant.

philwills:
👍

PR #1214, opened 2016-05-26
When other amount field is in focus, buttons become inactive.

jranks123:
If you then click on a button, the other amount field blanks out


tomverran:
👍

PR #1215, opened 2016-05-27
Retry getFullUserDetails call on failure

mario-galic:
@paulbrown1982 

[ID API connection timeout during Membership checkout](https://trello.com/c/lPzQbTGX/15-id-api-connection-timeout-during-membership-checkout)


paulbrown1982:
👍 but I think you should change the delay to 2 seconds rather than 3.

PR #1216, opened 2016-05-31
Retry Salesforce upsert on failure during checkout

mario-galic:
@paulbrown1982 @tomverran 

[Salesforce connection timeout during checkout](https://trello.com/c/gO3yzyAP/16-salesforce-connection-timeout-during-checkout)


paulbrown1982:
👍 but as discussed, change the number of retries to 2 rather than 3.

PR #1222, opened 2016-06-02
Hero banner responsive images

tomverran:
### What

This PR updates _all_ the membership hero banners to use fixed position `<picture>` tags with content hosted on The Grid rather than:

 - background-attachment: fixed in some places
 - @AWare and his fixed position `::before` element in others
 - non fixed images on mobile due to iOS bugs with background-attachment

The reason for doing this is that we need to pull background images for promotion landing pages from the promotion tool, so we have to pass in image URLs from the Scala somehow rather than defining them in CSS classes.

The picture element lets us support different crops for mobile & desktop as well as allowing the browser to select which resolution of image is most appropriate for the current device.

### How

We've basically reimplemented `background-attachment: fixed` in much the same manner that alex's fixed position `::before` element did but we use flexboxes to allow the hero banner image to be anchored to the top, the middle(ish) or the bottom of the hero element.

This JSFiddle is what we used to figure out what we were doing with flexboxes: https://jsfiddle.net/Lzztmbsk/

### Alternatives

If we go back to using background-attachment: fixed in the future we can support the same behaviour by generating CSS rather than picture elements - there's a CSS equivalent to srcset for background-images but it has much less browser support currently

### Screenshots

Some of the hero images are cropped slightly differently now so here are some screenshots

#### New

image

image

#### Old

image

image


SiAdcock:
👍

paulbrown1982:
👍

nlindblad:
👍

PR #1224, opened 2016-06-02
lets send the ophan id to stripe

AWare:

### DIFF
--- a/frontend/assets/javascripts/src/modules/giraffe.es6
+++ b/frontend/assets/javascripts/src/modules/giraffe.es6
@@ -2,7 +2,7 @@ import * as display from 'src/modules/form/validation/display'
 import * as helper from 'src/utils/helper'
 import * as ajax from 'ajax'
 import $ from '$'
-import * as ophan from 'src/modules/analytics/ophan';
+import {ophan} from 'src/modules/analytics/ophan';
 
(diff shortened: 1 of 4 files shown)


frontend/assets/javascripts/src/modules/giraffe.es6
```diff
@@ -2,6 +2,7 @@ import * as display from 'src/modules/form/validation/display'
 import * as helper from 'src/utils/helper'
 import * as ajax from 'ajax'
 import $ from '$'
+import * as ophan from 'src/modules/analytics/ophan';
```
tomverran: Since you're using ophan.ophan I think you can import {ophan} to pull the ophan property out of the object exported by src/moduls/analytics/ophan

frontend/assets/javascripts/src/modules/giraffe.es6
```diff
@@ -103,3 +106,11 @@ function getStuffFromIdentity() {
         }
     })
 }
+
+function ophanId(){
+
+    var $OPHAN = $('.js-ophan-id');
+    ophan.ophan.then(function(o){
```
tomverran: You can write this as an arrow function

tomverran:
👍

PR #1235, opened 2016-06-10
Add the change members status permission page

jranks123:
Adds a page to membership front end that controls whether a users membership status should display when they comment 

Currently only accessible to guardian staff members

screenshot at jun 10 13-10-18


paulbrown1982:
👍

PR #1237, opened 2016-06-13
Disable preselect test

AWare:
Disable preselected variant for contribution test. Use £25.


markjamesbutler:
👍  although an unrelated test is failing.  :(

PR #1241, opened 2016-06-27
Use membership common 222

tomverran:
Updates for membership common 222

Bit odd that membership needs to change for the paper catalog work but there you go


tomverran:
👍  from Niklas

PR #1248, opened 2016-06-29
CMP Tracking

markjamesbutler:
Add AB test to allow tracking of CMP.  This is a work-around until we can implement campaign codes properly.

?CMP=ema-3325
?CMP=ema-3324
?CMP=ema-3323
?CMP=ema-3322
?CMP=ema-3321


AWare:
:+1: let's fix this ASAP

PR #1249, opened 2016-06-29
Automatic geo-relocation on Contribute

jranks123:
- Checks the users location from the fastly header and if they are from UK, US or AU, redirects to the appropriate page, else defaults to the UK page

- Also makes sure to redirect with existing parameters

### DIFF
--- a/frontend/app/controllers/Giraffe.scala
+++ b/frontend/app/controllers/Giraffe.scala
@@ -71,10 +71,8 @@ object Giraffe extends Controller {
     val countryGroup = request.getFastlyCountry.getOrElse(CountryGroup.RestOfTheWorld)
 
     val baseUrl: Uri = redirectToGiraffe(countryGroup).absoluteURL
-    val paramsToPropagate = Seq("INTCMP", "CMP")
+    val paramsToPropagate = request.queryString.map { case (k,v) => k }.toSeq
     val urlWithParams = paramsToPropagate.foldLeft[Uri](baseUrl) { (url, param) =>
-      // No need to filter out the params that aren't present because if the `?` method gets a key-value tuple
-      // with value of None, that parameter will not be rendered when toString is called
       url ? (param -> request.getQueryString(param))
     }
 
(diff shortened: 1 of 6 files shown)


AWare:
Added my euro changes to this pr.

frontend/app/controllers/Giraffe.scala
```diff
@@ -62,14 +67,27 @@ object Giraffe extends Controller {
     )
   }
 
+  def contributeRedirect = NoCacheAction { implicit request =>
+    val countryGroup = request.getFastlyCountry.getOrElse(CountryGroup.RestOfTheWorld)
+
+    val baseUrl: Uri = redirectToGiraffe(countryGroup).absoluteURL
+    val paramsToPropagate = request.queryString.map { case (k,v) => k }.toSeq
```
markjamesbutler: It might be wise to extract the querystring concat as a function and write a unit test around it, simply so we can verify it produces valid url's with multiple query params.

frontend/app/controllers/Giraffe.scala
```diff
@@ -62,14 +67,27 @@ object Giraffe extends Controller {
...
 
   def contributeUK = contribute(CountryGroup.UK)
   def contributeUSA = contribute(CountryGroup.US)
   def contributeAustralia = contribute(CountryGroup.Australia)
+  def contributeEurope = contribute(CountryGroup.Europe)
 
-  def thanksUK = thanks(CountryGroup.UK, routes.Giraffe.contributeUK().url)
-  def thanksUSA = thanks(CountryGroup.US, routes.Giraffe.contributeUSA().url)
-  def thanksAustralia = thanks(CountryGroup.Australia, routes.Giraffe.contributeAustralia().url)
+  def thanksUK = thanks(CountryGroup.UK, routes.Giraffe.contributeRedirect().url)
```
markjamesbutler: routes.Giraffe.contributeRedirect().url could be a default param in thanks or simply called in the function rather than specified each time.

markjamesbutler:
👍

frontend/app/controllers/Giraffe.scala
philwills: Also, I think you can replace `request.queryString.map { case (k,v) => k }.toSeq` with `request.keys`

dominickendrick: Maybe overkill, but https://github.com/NET-A-PORTER/scala-uri makes uri handling fun again!

PR #1255, opened 2016-06-30
Giraffe autorelocate

jranks123:
- Implements auto relocation for visitors to the contributions page, based on their IP.
- Sends European, US, AU and UK visitors to their specific currency pages, defaults to US dollars


frontend/app/controllers/Giraffe.scala
```diff
@@ -110,3 +118,15 @@ object Giraffe extends Controller {
     })
   }
 }
+
+object MakeURL {
```
paulbrown1982: Have you seen the netaporter Scala Uri API? 

import com.netaporter.uri.dsl._
import com.netaporter.uri.Uri

With that you can do something like this:

var x = Uri.parse(request.uri).withScheme("https")
val url = x.copy(pathParts = Seq(PathPart(countryGroup.id)) ++ x.pathParts)

jranks123: Thanks Paul, this is much better

paulbrown1982:
👍

markjamesbutler:
:shipit:

PR #1257, opened 2016-06-30
Remove an unnecessary redirect for the geo-ip negotiation for /supporter

paulbrown1982:
Useful to save an additional redirect here from http to https.
cc @nlindblad @tomverran


AWare:
:+1:

PR #1260, opened 2016-07-04
Added extra query parameter to supporter redirect

paulbrown1982:
cc @AWare @JuliaBellis


JuliaBellis:
👍  Looks fine. We can accept mcopy.

AWare:
This is probably too late, but we've started accepting cmp and intcmp on contributions for this reason. mcopy wasn't really intended as a general purpose tracking code.

PR #1267, opened 2016-07-06
Revert "Use membership common 227"

tomverran:
Reverts guardian/membership-frontend#1266

As of membership-common 226 (which was never pushed to membership frontend) we now can't parse the production membership product catalog, I am reproducing in a test and sorting out now...


mario-galic:
👍

tomverran:
Going forward to v228 that fixes this

PR #1269, opened 2016-07-07
Dynamic credit card expiry dropdown

markjamesbutler:
The credit card expiry date will now show the current year plus 9 years.  

screen shot 2016-07-07 at 15 20 14


AWare:
This could probably be useful in the main codebase too. :) :+1:

PR #1272, opened 2016-07-13
Remove competitions from nav

markjamesbutler:
Competitions are currently on hold therefore removing them from the nav for now.

screen shot 2016-07-13 at 15 24 44


tomverran:
👍

PR #1273, opened 2016-07-14
Use membership common 233

tomverran:
I think in the future we should probably avoid membership depending on the existence of paper / digipack rate plan IDs but for now this sorts it all out


AWare:
👍

PR #1274, opened 2016-07-14
Endpoint for returning paid Membership plans with prices

agulowaty:
Mobile apps are implementing native membership flows and think these information will be helpful.

Compared to https://membership.theguardian.com/staff/catalog#, free, legacy and tests plans are excluded.

Endpoint will be accessible under `/rate-plans`, here is sample json output:
rate-plans-api

cc @chrisjowen @rtyley

### DIFF
--- a/frontend/app/controllers/PricingApi.scala
+++ b/frontend/app/controllers/PricingApi.scala
@@ -1,12 +1,18 @@
 package controllers
 
 import com.gu.i18n._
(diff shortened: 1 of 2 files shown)


tomverran:
:+1: though I think ideally we'd move this stuff to members-data-api

agulowaty:
rate-plans-prod

PR #1276, opened 2016-07-18
A more friendly page for members only content

philwills:
image

This is a new page to show if we have content that is only for members and the reader is not a member.

I believe the styling works, but I'm also pretty confident I've not got everything in sensible places, or  followed conventions properly, so feedback would be very welcome.


frontend/app/controllers/MemberOnlyContent.scala
```diff
@@ -0,0 +1,41 @@
...
+import services.{GuardianContentService, _}
+import tracking.ActivityTracking
+import views.support.PageInfo
+
+object MemberOnlyContent extends Controller with ActivityTracking
+                                 with LazyLogging {
+
+  val contentApiService = GuardianContentService
+
+  def membershipContent(referringContent: String) = NoCacheAction.async { implicit request =>
```
philwills: I need to have a bit more thought about whether we could actually cache this, which would obviously be preferable.

philwills:
One thing I was thinking was to redirect back to the content if the member only flag isn't set to avoid people fiddling with the URL to make it look like we're restricting content that we're not.

frontend/assets/stylesheets/components/_panels.scss
```diff
@@ -99,3 +99,20 @@
     padding: rem($gs-baseline * 2);
     margin-bottom: rem($gs-baseline * 2);
 }
+
+/* Challenge for member content */
+.challenge--panel {
```
joelochlann: Normally `--` is used for a modifier class in BEM syntax. i.e. it would only be used in combination with a base class, e.g. `.challenge`. https://github.com/guardian/membership-frontend/blob/master/docs/client-side-principles.md#unintended-modifiers

frontend/app/controllers/Joiner.scala
```diff
@@ -72,8 +73,13 @@ object Joiner extends Controller with ActivityTracking
       customSignInUrl=Some(signInUrl)
     )
 
-    Ok(views.html.joiner.tierChooser(TouchpointBackend.Normal.catalog, pageInfo, eventOpt, accessOpt, signInUrl))
-      .withSession(request.session.copy(data = request.session.data ++ contentRefererOpt.map(JoinReferrer -> _)))
+    contentRefererOpt.filter(Uri.parse(_).host.exists(_ == "www.theguardian.com")).map { contentURL =>
```
joelochlann: I might be reading this wrong, but does this mean that if there was a link on theguardian.com to the tier chooser page and it was clicked by the user, they would see the restricted content message?

joelochlann: Might be OK for now... but I could see that really coming back to bite someone in the future!

philwills: Now only redirects if the `membershipAccess` query param is present, which feels less prone to confusion.

joelochlann:
👍  once that caching question has been considered

PR #1278, opened 2016-07-18
Double Spaces Removal

markjamesbutler:
Removed accidental double spaces in variant names.  See Jack email:

----

Mark,

Some of the Amount VariantNames coming through to Stripe have double spaces in. e.g.

Amount - 50 highlight
Amount -  50 highlight

are both showing up (the former is preferable because it's simpler).

It's not crucial because it can be cleared up a bit in the analysis but it would make life a bit easier if it was all uniform.

Thanks,
Jack


AWare:
:+1:

PR #1279, opened 2016-07-19
Fix the jdk version

alexduf:


AWare:
It might be worth removing cloud-formation/gen-ami.sh at the same time. @rtyley?

PR #1284, opened 2016-07-20
pin grunt asset hash

AWare:
There's a new version of grunt-asset-hash that's compatible with new versions of grunt....
...and our ^ised version number matches it and breaks.


markjamesbutler:
👍

AWare:
@paulbrown1982

AWare:
(discussed with Paul, merging)

PR #1287, opened 2016-07-28
app message for supporter landing pages

AWare:
On suggestion from Graham:
image

Might need a slight copy tweak?

### DIFF
--- a/frontend/app/views/fragments/tier/packagePromoSupporter.scala.html
+++ b/frontend/app/views/fragments/tier/packagePromoSupporter.scala.html
@@ -31,6 +31,7 @@
             <ul class="o-bulleted-list">
                 <li>Welcome pack, card and gift</li>
                 <li>Regular Members updates</li>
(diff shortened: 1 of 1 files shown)


alexduf:
I got confused and thought it would be ad free on the website. I had a chat with @AWare who confirmed it was app only (which makes sense now I'm reading it again)
Maybe there's a another way to formulate that sentence?

frontend/app/views/fragments/tier/packagePromoSupporter.scala.html
```diff
@@ -31,6 +31,7 @@ <h4 class="u-margin-top">Benefits:</h4>
             <ul class="o-bulleted-list">
                 <li>Welcome pack, card and gift</li>
                 <li>Regular Members updates</li>
+                <li>Enjoy an ad-free experience with free access to the premium tier of the app</li>
```
tomverran: Consider adding this to Benefits.scala maybe
and perhaps powering this view off it

AWare:
closing needs refactor

PR #1300, opened 2016-08-03
/join to go to Supporter landing page

paulbrown1982:
Let's make /join (which is linked to from the MMA page amongst other pages) go to the relevant Supporter page for that account, rather than the homepage. There's a second button in the UI 
which goes to /about which also goes to the homepage, which we can leave alone - and the customer can self-select based on how much they know already about membership.

picture 159

The redirect was previously a 301, so this might not be 100% reliable, but neither scenario are a bad thing.

Cc @AWare @rtyley


AWare:
👍

PR #1302, opened 2016-08-04
Change social thank you to link to supporter url

AWare:
Currently the social supporter link goes to the homepage, which is a bit UK centric. 
Maybe this should point to the supporter page. (Need to build this at work)


paulbrown1982:
👍

AWare:
Tested on prod.

PR #1303, opened 2016-08-05
Use the Subs Google Analytics property

paulbrown1982:
Use the Subs Google Analytics property, as it'll be that property which will be converted next week into the umbrella one for the Membership Stream.

cc @rtyley @AWare @tomverran


AWare:
:+1:

PR #1307, opened 2016-08-12
more ga tracking

AWare:
track the additional required fields for ga
@ajosephides @paulbrown1982

### DIFF
--- a/frontend/app/views/joiner/thankyou.scala.html
+++ b/frontend/app/views/joiner/thankyou.scala.html
@@ -52,9 +52,9 @@
 
 @main(title) {
 <script type="text/javascript">
-   window.productData = {
-       regNumber: "@member.contact.regNumber.getOrElse("")",
-       tier: "@member.subscription.plan.tier.name"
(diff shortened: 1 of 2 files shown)


ajosephides:
@AWare - will you also add the internal campaign code dimension (12). Something I might need to run through with you.

frontend/app/views/joiner/thankyou.scala.html
```diff
@@ -51,7 +51,12 @@ <h2 class="page-section__headline">@title</h2>
 }
 
 @main(title) {
-
+<script type="text/javascript">
+   window.productData = {
```
paulbrown1982: Maybe add this to window.guardian.productData instead? That would be more consistent with subs.

AWare: window.guardian isn't instantiated until the foot.

AWare:
Changed as per @paulbrown1982  and @ajosephides

paulbrown1982:
👍

PR #1309, opened 2016-08-16
Eventbrite cross domain ga

paulbrown1982:
In order for Google Analytics to track user sessions through the EventBrite screens, we need to pass in the user's client id into the _eboga query parameter. 

[See EventBrite docs here](https://www.eventbrite.co.uk/support/articles/en_US/Troubleshooting/how-to-enable-cross-domain-and-ecommerce-tracking-with-google-universal-analytics)

cc @ajosephides @AWare


ajosephides:
Eventbrite has been added to the Referral exclusion list on the Membership GA Property.
With this method of Cross Domain Tracking being used instead I don't believe we will need the Linker plugin and the eventbrite domain that is already in place.

Can we initially remove this plugin also.

paulbrown1982:
https://trello.com/c/i7ZzAUyw/597-google-analytics-remove-linker-plugin

AWare:
👍

PR #1312, opened 2016-08-26
Add redirect to members only content page

AWare:
Allow users to keep reading once they have signed up from a members only content link.


nlindblad:
👍

PR #1313, opened 2016-08-26
quick fix for member content links

AWare:
This fixes the non-display of member content links in thank you pages. 
We'll need to look at the problems in dev next week.


AWare:
(The Salesforce key rotation kerfuffle)

nlindblad:
👍

PR #1314, opened 2016-08-26
fix sign in redirect

AWare:
Make the sign in link redirect to the right place.


nlindblad:
👍 Let's fix it properly next week.

PR #1315, opened 2016-08-30
fix members only content link (don't break the session)

AWare:
- Add some brackets so we don't make the session some kind of hellish map[map[map[map[
- Use the config host
- This fixes the link to the article
@tomverran


tomverran:
👍  🌍 🔍

PR #1318, opened 2016-09-02
Make the opengraph stuff for members only content a bit better

AWare:
Make the description the trail and the image the mainPicture from the imageGroup. 
When the codefreeze is done, maybe figure out how to add the nice blue bottom border. 
@tomverran


tomverran:
👍

AWare:
This isn't great. 
image

PR #1321, opened 2016-09-15
Make longer section names show properly

philwills:
Before:

image

After:

image

### DIFF
--- a/frontend/assets/stylesheets/components/_articles.scss
+++ b/frontend/assets/stylesheets/components/_articles.scss
@@ -244,9 +244,16 @@
     float: left;
     padding-right: $gs-gutter/3;
 
-    @include mq(tablet) {
+    @include mq(wide) {
         @include fs-header(4);
         line-height: get-line-height(header, 2);
+        float: none;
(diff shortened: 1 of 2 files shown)


frontend/assets/stylesheets/components/_articles.scss
```diff
@@ -244,12 +244,19 @@
     float: left;
     padding-right: $gs-gutter/3;
 
-    @include mq(tablet) {
+    @include mq(wide) {
         @include fs-header(4);
         line-height: get-line-height(header, 2);
```
AWare: Might be worth just explicitly setting font size here and using the font-only mixin?

philwills: I'm not sure about the font-only mix-in, but when trying to do mq queries without the repetition, the precedence rules were confusing me and I couldn't get a consistent result.

If you've got time to show me, I'd be happy to try though.

AWare: `font-size: 22px;
line-height: 24px;
@include f-header
`

PR #1323, opened 2016-09-20
Convert DestinationService into a class

tomverran:
`DestinationService` is a relatively small service that given a session figures out where you came from when you signed up to become a member (either an event or some restricted content) and then shows you some content on the thank you page (either an eventbrite iframe or a link back to the article)

It was implemented as a trait for testability but because it depended on some quite serious objects (most notably _the entirety of TouchpointBackend_) the test was written with mockito. This is the only test we have which uses mockito and I'd rather we ensure that code depends on the minimum possible and is generic enough that we don't need to use it.

The test also depended on an entire play app running, so I got rid of that.

Also `DestinationService` deals with Futures of Options but was written before OptionT was available so I simplified the implementation a bit


tomverran:
image

PR #1329, opened 2016-09-29
Fix issue where path was displayed as text on error page.

paulbrown1982:
Fixed security issue where path was displayed as text on error page.

cc @adamnfish @philwills @JuliaBellis 

https://trello.com/c/vZHtNn6V/15-fix-error-content-injection-in-membership-frontend

Bad:

picture 197

Good:

picture 198


tomverran:
👍

PR #1332, opened 2016-10-04
New subs model

johnduffell:
Update to the latest membership common!

This gives us the new refactored models with the new readers to allow guardian weekly.  Also okhttp3 and some small changes.

@tomverran


tomverran:
> some small changes

image

😄

tomverran:
👍

johnduffell:
>What's gone wrong?

😢 https://github.com/guardian/membership-frontend/pull/1334

frontend/app/views/support/MembershipCompat.scala
```diff
@@ -0,0 +1,103 @@
...
+      case Tier.Partner() => in.partner
+      case Tier.Patron() => in.patron
+    }
+
+    def findPaid(p: ProductRatePlanId): Option[CatalogPlan.PaidMember[BillingPeriod]] =
+      Try(unsafeFindPaid(p)).toOption
+
+    def findFree(f: FreeTier) = f match {
+      case Tier.Friend() => in.friend
+      case Tier.Staff() => in.friend // fix
```
rtyley: The bigger PRs are, the more likely something like this will slip through :cry:  See PR https://github.com/guardian/membership-frontend/pull/1344 for the fix.

johnduffell: oh dear! no idea why that was there, it was all a race to deploy before tom left so I don't think it got reviewed at all. thanks for fixing!

tomverran: Ah gosh sorry this must've been done before staff was in the new Catalog
And yes the whole situation was a bit less than ideal

johnduffell: if that's the worst problem that happens then we're in the clear ;)

PR #1334, opened 2016-10-04
prevent NPE caused by initialisation order

johnduffell:
Problem due to how I refactored for OKHTTP3, now fixed and working locally

@jacobwinch cc/ @tomverran


philwills:
Repeat after me, "I will not use brittle trait hierarchies".

PR #1336, opened 2016-10-05
Extend timeout for createSubscription call

jacobwinch:


johnduffell:
"No description provided." 😢 
I'm sure it's a great idea, convince me to give you a +1! 😛

jacobwinch:
Apologies for the lack of description! We re-factored the Zuora Soap Client in membership-common to take an extendedHttpClient (i.e. a customisable OkHttpClient instance), as well as the default client function. This is passed through so that we can make specific requests with a timeout greater than 10s (the default). 

The result of this change is that the createSubscription call (issued when a user tries to pay us) gets a timeout of 20s. This behaviour has been implemented in subscriptions-frontend too, because this call has been regularly failing in production when using the default timeout.

johnduffell:
fab sounds great! 🚀 👍 thanks 😉

PR #1338, opened 2016-10-07
Add membership profile

svillafe:
Help new starters by letting know they need to add profile membership  into the config command cc: @rtyley @mario-galic


rtyley:
:+1:

PR #1341, opened 2016-10-12
Removed a GA validation that was always false.

svillafe:
Due to the fact that the line [#40 of setup.js](https://github.com/guardian/membership-frontend/blob/master/frontend/assets/javascripts/src/modules/analytics/setup.js#L40) was setting up google analytics as the second step of a promise, this condition was always false and therefore the listeners were never bound and we were not tracking events.

Please @paulbrown1982 @AWare @rtyley review this change before merging it into master.

### DIFF
--- a/frontend/assets/javascripts/src/modules/metrics/triggers.js
+++ b/frontend/assets/javascripts/src/modules/metrics/triggers.js
@@ -13,7 +13,7 @@ define([
 
     function init() {
         var trackingElems = document.querySelectorAll('[' + METRIC_ATTRS.trigger + ']');
-        if (window.ga && trackingElems) {
+        if (trackingElems) {
             configureListeners(trackingElems);
         }
     }


frontend/assets/javascripts/src/modules/metrics/triggers.js
```diff
@@ -13,7 +13,7 @@ define([
 
     function init() {
         var trackingElems = document.querySelectorAll('[' + METRIC_ATTRS.trigger + ']');
-        if (window.ga && trackingElems) {
+        if (trackingElems) {
```
rtyley: I'm not sure this is the right fix for this problem? `configureListeners(trackingElems);` presumably does want to wait until *both* `window.ga` and `trackingElems` are ready?

It ultimately calls the stuff in https://github.com/guardian/membership-frontend/blob/1ff87757b/frontend/assets/javascripts/src/modules/metrics/logger.js - which needs `ga` to be present by that point, otherwise it will be as ineffective as the current code?

I'd guess what you really need is a callback that gets called when `ga` loads?

paulbrown1982: The logger function also has its own window.ga presence check before actually calling the send function on window.ga, so the event handler bindings can still happen regardless of whether Google is loaded or not.  Hence a 👍 from me.

svillafe:
Waiting for @AWare review before we merge it.

PR #1342, opened 2016-10-13
analytics triggers

AWare:
@svillafe and I worked on fixing some bugs that I caused in the ga upgrade 

This reenables some of the ga events that we used to send to the old account, so will need to chat with @paulbrown1982, @JuliaBellis and @ajosephides about whether we want to be sending all of these events. cc @rtyley

### DIFF
--- a/frontend/app/views/event/eventDetail.scala.html
+++ b/frontend/app/views/event/eventDetail.scala.html
@@ -79,9 +79,6 @@
                     @for(highlight <- event.metadata.highlightsOpt) {
                         <a class="action action--external"
                            href="@highlight.url"
-                           data-metric-trigger="click"
-                           data-metric-category="events"
-                           data-metric-action="highlights"
                         >
                            @highlight.title
                            @fragments.actionIcon("arrow-right")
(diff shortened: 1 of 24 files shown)


ajosephides:
I agree with having a review of the events being sent. Previously this would not have had a cost impact but since we have moved to a Premium GA we will be billed for each server call / event sent.

svillafe:
The complete list of **events** that we are tracking can be located in: https://docs.google.com/a/guardian.co.uk/spreadsheets/d/1nkagwA0fRXcqejwqv-V0wv--FIGb5dl9D_m7YUS4jhY/edit?usp=sharing

rtyley:
:+1: I think that it's okay to merge this, let it run for a day, and see how expensive the traffic is. If the level is too bad, we can always turn it off again tomorrow.

AWare:
I think the new GA account has some namespace conventions that we should probably adopt before merging.

ajosephides:
@rtyley - as opposed to throwing everything up can we get an idea of what is being reported on first.
@AWare - as a standard we have events being sent to GA with noninteraction = true

svillafe:
I deleted all the events that we were tracking in the Membership project. In addition, I added trackers only for the three events which take place in the supporter's landing page. These events were requested by @JuliaBellis . Please @ajosephides @AWare @paulbrown1982 @rtyley review this changes.

AWare:
Todo: 

- rexamine whether wrappedGa is the right way to set the tracker and proxy ga
- set noninteraction=true before we start using this

svillafe:
The events are being seen in production under the realtime feature. picture 14

PR #1344, opened 2016-10-14
Fix Staff tier - I think this is all we need to do!

paulbrown1982:
cc @rtyley @svillafe @Ap0c @johnduffell

### DIFF
--- a/frontend/app/views/support/MembershipCompat.scala
+++ b/frontend/app/views/support/MembershipCompat.scala
@@ -96,7 +96,7 @@ object MembershipCompat {
 
     def findFree(f: FreeTier) = f match {
       case Tier.Friend() => in.friend
-      case Tier.Staff() => in.friend // fix
+      case Tier.Staff() => in.staff
     }
   }
 


Ap0c:
Excellent fix!

One thing that still seems odd: when you first set up a staff membership you initially get an email saying thanks for joining as a Partner, and then immediately get a second email saying your tier has been changed to Friend... This seems to explain the Friend part, but is there a reason it would assign a Partner tier first?

paulbrown1982:
def joinStaff = AuthenticatedNonMemberAction.async { implicit request =>
    staffJoinForm.bindFromRequest.fold(redirectToUnsupportedBrowserInfo,
        makeMember(Tier.partner, Redirect(routes.Joiner.thankyouStaff())) )
  }

FWICT It's always been set way. Staff get Partner benefits - and I think this setting makes Zuora set up usable benefits on the staff sub - i.e. 6 free event tickets.

johnduffell:
👍 good spot! not sure why it was left like that, hopefully there wasn't a good reason.

svillafe:
It is working picture 12

paulbrown1982:
From Zuora: picture 209

PR #1345, opened 2016-10-14
Fix Campaign Querystring When Coming From Supporter Landing Page

Ap0c:
# Problem

When a user is on the supporter landing page, and they click on the button to become a supporter, they are sent to the `enter-details` page. However, if they are not logged in they are first sent to the identity `registration` page. This page is given the `enter-details` page as a return url, so they can continue once they've signed up:

```
Supporter Landing Page
         |
         V
Enter-Details Endpoint (not signed-in)
         |
         V
Registration (Return Url = Enter Details Page)
         |
         V
Enter Details Page (signed in)
```

This return url was being constructed incorrectly, and was dropping the campaign (`INTCMP`) query param after signup (it wasn't encoded, and was thus being interpreted as a query param by the `registration` page).

This is probably what has led to campaigns not being reported properly in the data lake.

# Fix

Add the campaign code param to the return url, rather than the registration page url. It is now encoded properly and is passed on to the `enter-details` page once the user has signed up.

@guardian/membership-and-subscriptions @rtyley

### DIFF
--- a/frontend/app/actions/RegistrationUri.scala
+++ b/frontend/app/actions/RegistrationUri.scala
@@ -4,6 +4,7 @@ import com.gu.i18n.CountryGroup
 import com.gu.i18n.CountryGroup.RestOfTheWorld
 import com.gu.salesforce.Tier
 import com.netaporter.uri.Uri
+import com.netaporter.uri.dsl._
 import configuration.Config
 import play.api.mvc.RequestHeader
 
@@ -33,10 +34,10 @@ object RegistrationUri {
     controllers.routes.Joiner.joinFriend().path -> "FRI"
   )
 
-  def parse(request: RequestHeader) = {
+  def parse(request: RequestHeader): String = {
     val redirectUrl: String = Config.idWebAppRegisterUrl(request.uri)
     val campaignCode = extractCampaignCode(request.headers.get(REFERRER_HEADER), request.path)
-    redirectUrl.concat(s"&INTCMP=$campaignCode")
+    (redirectUrl ? ("INTCMP" -> campaignCode)).toString
   }
 
   private def extractCampaignCode(refererOpt: Option[String], path: String) : String = {
--- a/frontend/app/actions/RegistrationUri.scala
+++ b/frontend/app/actions/RegistrationUri.scala
@@ -35,8 +35,9 @@ object RegistrationUri {
   )
 
   def parse(request: RequestHeader): String = {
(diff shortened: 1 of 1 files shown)


frontend/app/actions/RegistrationUri.scala
```diff
@@ -33,10 +34,11 @@ object RegistrationUri {
     controllers.routes.Joiner.joinFriend().path -> "FRI"
   )
 
-  def parse(request: RequestHeader) = {
-    val redirectUrl: String = Config.idWebAppRegisterUrl(request.uri)
+  def parse(request: RequestHeader): String = {
     val campaignCode = extractCampaignCode(request.headers.get(REFERRER_HEADER), request.path)
-    redirectUrl.concat(s"&INTCMP=$campaignCode")
+    val returnUri: String = (request.uri ? ("INTCMP" -> campaignCode))
```
johnduffell: what's this `?` thing, is it a function?

frontend/app/actions/RegistrationUri.scala
```diff
@@ -33,10 +34,11 @@ object RegistrationUri {
...
+    val redirectUrl: String = Config.idWebAppRegisterUrl(returnUri)
+    (redirectUrl ? ("INTCMP" -> campaignCode)).toString
```
johnduffell: so we need the campaign code on the id url still, as well as adding it to the return url?

frontend/app/actions/RegistrationUri.scala
johnduffell: I'd rename `returnUri` to `returnHereUrlAfterLogin` and `redirectUrl` to `identityRegistrationUrl` (or similar) otherwise it's hard to know what's going on as they both look too similar and not concrete

johnduffell:
apart from the above looks great 👍 good find!

frontend/app/actions/RegistrationUri.scala
Ap0c: Ok renamed the variables (actually, got rid of one from your comment below): 019179d.

frontend/app/actions/RegistrationUri.scala
Ap0c: Nope, that was left over from my first change, good spot, removed it: 019179d.

johnduffell:
thanks! and btw excellent description on the PR, it's really clear to know what and why you're doing it so makes it simple to review 👍 🚀 🎉

PR #1347, opened 2016-10-17
Update Membership Header and Footer to Guardian Blue Colours

Ap0c:
**Merge Notice:** We are not merging this until the mobile layout changes are ready.

# Reason

The intention is that if we change the membership brand colours to be the Guardian colour palette then our users are more likely to convert, because they will less confused by the product and who they're giving their money to.

# Changes

- Updated masthead (header) png to version with new colours.
- Updated brandbar (footer) png to version with new colours.

# Screenshots

**Original:**

membership-colours-original

**Updated:**

membership-colours-updated


johnduffell:
> The intention is

no need for an a/b test?

paulbrown1982:
I feel that a change to fundamental Membership site branding should be driven by the Commercial dept. Given there's a user lifecycle within the membership microsite, any A-B test would ideally be sticky to that user forever. Also the Identity pages during the checkout flow are styled to match membership and would need to be updated and be in A/B sync too to gain the confusion reduction. Happy to discuss but I feel there's a reputation/customer experience risk in just rolling this out without more thought.

chriswilk:
@paulbrown1982 I've raised this with Natalie / David - they have given us the OK

paulbrown1982:
Cool, I've spoken to Chris, Julia etc, so this is a 👍 from me too.

Ap0c:
Minus a fastly cache issue that's now been fixed (thanks @paulbrown1982!), this has gone through ok.

PR #1349, opened 2016-10-18
Prevent Campaign Code Overwrite

Ap0c:
# Problem

As one thing is fixed, another thing breaks. Fixing the supporter campaign codes in #1345 has led to loss of reporting for the engagement banner campaign codes. Because the banner links to the supporter landing page, the campaign codes that are set when redirecting to sign-in from the landing page overwrite the original ones set by the engagement banner:

```
Engagement Banner (cc: mem_uk_banner)
          |
          V
Supporter Landing Page (cc: mem_uk_banner)
          |
          V
Sign-In (cc: mem_supuk_sup)
          |
          V
Checkout (reported cc: mem_supuk_sup)
```

# Solution

Existing campaign codes take priority. If a campaign code already exists in a cookie (i.e. if it has already been set by the engagement banner), it will be passed through on the sign-in redirect. Otherwise, the supporter page campaign code will be set.

# Changes

- Update the `CampaignCode.fromRequest` method to take the more generic `RequestHeader` type, as it still allows us to retrieve the cookie information.
- Set up the redirect uri generator to check the cookie for a campaign code first, before setting the supporter code.

# Longevity

This is a short-term fix to get the reporting working again correctly (needed urgently). Long-term, this code relies on omniture, so we'll be stripping some of it out and replacing with our own solution soon.

@rtyley @guardian/membership-and-subscriptions

### DIFF
--- a/frontend/app/actions/RegistrationUri.scala
+++ b/frontend/app/actions/RegistrationUri.scala
@@ -7,7 +7,6 @@ import com.netaporter.uri.Uri
 import com.netaporter.uri.dsl._
 import configuration.Config
 import play.api.mvc.RequestHeader
-import play.utils.UriEncoding
 import utils.CampaignCode
 
 import scala.util.Try
--- a/frontend/app/actions/RegistrationUri.scala
+++ b/frontend/app/actions/RegistrationUri.scala
@@ -7,6 +7,8 @@ import com.netaporter.uri.Uri
 import com.netaporter.uri.dsl._
 import configuration.Config
 import play.api.mvc.RequestHeader
+import play.utils.UriEncoding
+import utils.CampaignCode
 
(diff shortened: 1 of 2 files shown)


frontend/app/actions/RegistrationUri.scala
```diff
@@ -35,7 +36,12 @@ object RegistrationUri {
   )
 
   def parse(request: RequestHeader): String = {
```
rtyley: An observation - `parse()` is a really bad name for this method (I realise you didn't introduce this!) - half of the work it does is actually constructing the redirect url.

rtyley:
:+1:

Ap0c:
Looks like it's working ok! Banner codes being passed through in the query param.

PR #1352, opened 2016-10-20
Stop Sending INTCMP to Identity

Ap0c:
# Plan

Currently we're sending internal campaign codes to identity every time a user registers or logs in via membership. Previously introduced in [this PR](https://github.com/guardian/membership-frontend/pull/818). This functionality is not needed any more, so we're removing it. Also, all the information we pass in the `INTCMP` query param is already present in the return URL itself.

# Additional Info

Currently when a user attempts to become a supporter (or partner or patron) they are passed through an authentication action. If they are logged in, they are sent straight to the checkout page. If not, they are redirected to identity for signup (with an option to log in). A return url is given to identity, to send them on to the checkout once they have signed up/logged in. Currently an internal campaign code it also sent.

# Changes

- Remove `RegistrationUri` module, as the entire thing exists to send this code.
- Update `chooseRegister` to build the registration url straight from the config, rather than going via `RegistrationUri`.
- Remove tests for `RegistrationUri` module.

@rtyley @markjamesbutler @paulbrown1982


rtyley:
:+1:

paulbrown1982:
👍

Ap0c:
INTCMP param not being sent to identity:

intcmp-identity-gone

Identity not passing it on to checkout:
intcmp-gone

This seems to be working!

PR #1354, opened 2016-10-24
Fix Graphical Bugs On Membership Homepage

Ap0c:
# Problem

Two graphical bugs affecting membership homepage.

## Bug 1

"Become a Friend" button appearing blank:

blank-friend-button

Introduced in #1342, missing `>` on the anchor tag.

## Bug 2

A number of package action buttons wrapping weirdly in Firefox:

firefox-button-wrapping

possibly caused by [this Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=488725). We already have code dealing with another case of this [here](https://github.com/guardian/membership-frontend/blob/a799ebade9eb94868e18ad0a90981ccb2d9f36de/frontend/assets/stylesheets/components/_packages.scss#L35).

# Solution

## Bug 1

Add `>` to anchor tag:

fixed-friend-button

## Bug 2

Remove `white-space: nowrap` property, which prevents the conflict with `float`. As far as I can tell, the default of `white-space: normal` looks fine across different browsers and breakpoints, and fixes the problem in Firefox:

fixed-firefox-buttons

### DIFF
--- a/frontend/app/views/fragments/tier/friendTeaser.scala.html
+++ b/frontend/app/views/fragments/tier/friendTeaser.scala.html
@@ -12,7 +12,7 @@
     <div class="package-teaser__action">
         <a href="@routes.Joiner.enterFriendDetails().url"
             class="action action--secondary qa-friend-join"
-            aria-title="Join as a Friend for free"
+            aria-title="Join as a Friend for free">
             Become a Friend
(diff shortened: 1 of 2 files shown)


Ap0c:
Looks like it's fixed now in PROD 👍.

PR #1356, opened 2016-10-25
Removing benefits which are no longer available.

paulbrown1982:
Offers and competitions, and live streams are no longer available on the membership site, so removing them at Commercial's request.

cc @rtyley


rtyley:
:+1:  I see that the link to competitions was removed back in July with #1272, so this change naturally follows on from that.

PR #1358, opened 2016-10-26
Switch to using build.sbt, revel in auto-plugin happiness

rtyley:
Using .sbt build definitions (rather than .scala definitions) has been the recommended way of defining SBT projects since SBT v0.13.12, for all project types, including multi-project builds.

Using auto-plugins is much nicer with .sbt build definitions - the necessary imports are all automatically made for you. It's become fiddlier to do .scala-based build definitions because plugin authors no longer document what those necessary imports are!

I've deliberately tried to not shift too many files round with this commit, but at some point it would be nice to bring the sole `frontend` subproject up to the root level.

http://notes.implicit.ly/post/147554513094/sbt-01312
http://www.scala-sbt.org/0.13/docs/Basic-Def.html
https://groups.google.com/d/msg/sbt-dev/S-pr2nZQ3KY/iX9q3cbrHVQJ

cc @paulbrown1982 @AWare


rtyley:
> Overdue on PROD (merged by @rtyley 14 minutes and 3 seconds ago) What's gone wrong?

This deploy has failed due to https://github.com/sbt/sbt-native-packager/issues/759 - it's affected two successive builds (10075 & 10077)

I notice that [we seemed to find a fix for the issue](https://github.com/guardian/members-data-api/pull/100) with `useNativeZip`, I'll open a separate PR to make that change here.

PR #1360, opened 2016-10-26
Use official sbt-riffraff-artifact plugin, rather than the copy-pasta

rtyley:
The official `sbt-riffraff-artifact` plugin also supports the newer riff-raff formats which we'd probably want to take advantage of (no need for nesting a zip file inside a zip file, etc).

Note the TeamCity build against this PR will fail, because it runs the `play-artifact` sbt task - which is removed with this change! We will need to change our TeamCity config at: https://teamcity-aws.gutools.co.uk/admin/editRunType.html?id=buildType:memsub_membership_frontend&runnerId=RUNNER_13 ..the current TeamCity sbt commands are:

```
clean fast-test play-artifact
```

...we'll want to change that to:

```
clean fast-test riffRaffNotifyTeamcity
```

The generated artifact is also slightly different (`app.zip` is now named `frontend-1.0-SNAPSHOT.tgz`), so we need to change our cloud-formation also, to unzip the correct file when we start up a box.

https://github.com/guardian/sbt-riffraff-artifact/

cc @AWare @Ap0c @philwills


cloud-formation/membership-app.cf.json
```diff
@@ -108,9 +108,9 @@
 
                         "adduser --system --home /membership --disabled-password membership",
 
-                        {"Fn::Join":["", ["aws --region ", { "Ref": "AWS::Region" }, " s3 cp s3://membership-dist/", { "Ref" : "Stack" }, "/", { "Ref" : "Stage" }, "/frontend/app.zip /tmp"]]},
+                        {"Fn::Join":["", ["aws --region ", { "Ref": "AWS::Region" }, " s3 cp s3://membership-dist/", { "Ref" : "Stack" }, "/", { "Ref" : "Stage" }, "/frontend/frontend-1.0-SNAPSHOT.tgz/tmp"]]},
 
-                        "unzip -d /membership /tmp/app.zip",
+                        "unzip -d /membership /tmp/frontend-1.0-SNAPSHOT.tgz",
```
philwills: You can avoid this change by overriding the `packageName` value from the sbt-native-packager plugin. Also, I didn't know `unzip` could handle `tgz` packages.

rtyley: Good catch! I still don't know how to make it generate a zip file rather than a tgz (so I still need a name change), but I should have remembered that I also need a change of decompression command.

rtyley:
I managed to botch the deploy (editing cloud-formation while a deploy was going on - seemed to result in the 3 healthy boxes getting terminated too, soon), and take down the site for 8m, but I've managed to clean up my mess -everything's working again now.

PR #1361, opened 2016-10-28
Add GA Metric Tracking To Record Membership Conversion

Ap0c:
# Plan

We would like to use GA to provide data on users either joining membership, or upgrading their membership tier. We are currently using a goal-based implementation for tracking in GA on Membership. This has some issues, including the fact that it appears to be inaccurate.

The plan is to switch to using the metric-based implementation that is currently used by Jobs and Subscriptions (while keeping the goal option running in parallel for a while to compare the numbers).

This change will cover joining as a Friend, Supporter, Partner or Patron. It will also cover upgrading to a Supporter, Partner or Patron.

# Implementation

- Update the `ga` script on the client-side to send an event to GA when a user arrives on a 'thank you' page.
- Use this event to increment a metric in GA to signify joining/upgrading to a given tier.
- Only run this on the 'thank you' page.
- Tweak the `guardian` page object on the 'thank you' page, so the client-side knows if the user is upgrading or joining for the first time.

@ajosephides @AWare @svillafe @rtyley @paulbrown1982


ajosephides:
👍

Ap0c:
Looks like it's working as expected, will check the GA results on Monday when it will hopefully have accumulated a few sign-ups. If there are any problems, we're still running and creating reports based upon goals, so the data shouldn't be affected.

PR #1363, opened 2016-11-01
Update sbt-buildinfo

rtyley:
We can now call it like an autoplugin - saves 1 line of sbt conf!

cc @Ap0c


rtyley:
Looks like that built just fine - Prout was able to get the buildinfo, as expected.

PR #1364, opened 2016-11-01
Update to latest membership-common

paulbrown1982:
The latest Membership-common separated out Ireland as its own country group, rather than it being part of Europe. This was so that the commercial team could have a specific European country to reference on the checkout page - a selected country is required by Subs and Membership promotion logic.

This change updates Membership to handle requests from Ireland like it always has done. Also I tweaked the sitemap page to handle alternate URLs differently, please let me know if that's OK.

cc @rtyley @Ap0c @svillafe

### DIFF
--- a/frontend/app/controllers/SiteMap.scala
+++ b/frontend/app/controllers/SiteMap.scala
@@ -16,12 +16,13 @@ object SiteMap extends Controller with LazyLogging {
       <url>
         <loc>{routes.FrontPage.index.absoluteURL(secure = true)}</loc>
         <priority>0.8</priority>
+        {alternatePage(routes.FrontPage.index.absoluteURL(secure = true), "x-default")}
       </url>
     </urlset>
     Ok(foo)
   }
 
-  def supporterPages()(implicit req: RequestHeader): Iterable[Elem] = for {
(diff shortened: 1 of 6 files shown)


Ap0c:
Looks like the membership flow is working ok after these changes. The sitemap changes are building fine, and main checkout flow works as expected for Ireland (routes through EU pages).

@rtyley has some comments, but apart from that LGTM.

frontend/app/controllers/SiteMap.scala
```diff
@@ -10,41 +9,41 @@ import scala.xml.Elem
...
+    countryGroup <- ActiveCountryGroups.all
   } yield {
     <url>
       <loc>
         {routes.Info.supporterFor(countryGroup).absoluteURL(secure = true)}
       </loc>
       <priority>1.0</priority>
-      {alternatePages()}
+      {alternatePages(ActiveCountryGroups.all)}
+      {alternatePage(routes.Info.supporterFor(countryGroup).absoluteURL(secure = true), "x-default")}
```
paulbrown1982: No you're right, there is no true un-localised page, so I'll remove the x-default alternate.

paulbrown1982:
Closing as I have backed out there being an Ireland country group in membership-common.

PR #1367, opened 2016-11-02
Refactor config to prepare for deployable .debs & systemd-based AMIs

rtyley:
The config include-chain is now this (each file contains an include for the next):

- `application.conf`
- `/etc/gu/membership.private.conf` (downloaded from S3)
- `[STAGE].public.conf` (from a stage-specific include in `membership.private.conf`

You can download the private credentials like this:

```
$ aws s3 cp s3://membership-private/DEV/membership.private.conf /etc/gu/membership.private.conf --profile membership
```

This layout of config was adopted for Contributions in July 2016 with https://github.com/guardian/contributions-frontend/pull/30, and it allows us to use Debian packages, rather than zips, for our deployable artifacts. Debian installation doesn't support the `-Dconfig.resource` parameter we've used with the previous Upstart-based scripts, prompting this refactor. Also, if we want to use [newer Ubuntu AMIs from Amigo](https://amigo.gutools.co.uk/recipes/xenial-membership), we need to move to systemd, which is also supported by the Debian-based artifacts.

Note that `membership.private.conf` is now performing a double-duty:

1. Storing your private credentials!
2. Including the correct public conf for your stage: `[STAGE].public.conf`

The config files for DEV & PROD should be kept very similar in structure, so that examining them with a diffing tool can easily show any missing keys or mistakes, ie:

- `DEV.public.conf` & `PROD.public.conf` should diff well
- `DEV/membership.private.conf` & `PROD/membership.private.conf` also

cc @Ap0c @svillafe @AWare @mario-galic


rtyley:
With the addition of [blank.private.conf](https://github.com/guardian/membership-frontend/blob/80272dd2/frontend/test/acceptance/conf/blank.private.conf) the CI tests now pass :grinning:

rtyley:
Post deployment testing passed, but there was an error in the `google.oauth.callback` config values that broke Google authentication, fixed with c2fe5c1, deploying now.

rtyley:
This is now deployed to PROD, access to https://membership.theguardian.com/test-users works, and I was able to sign up as a supporter using a test-user...

image

rtyley:
Looking at https://dashboard.ophan.co.uk/graph/breakdown?hours=1&platform=membership&path=/join/supporter/thankyou&by=referrer also shows successful real-person sign-ups since the change went out too, so looks like we're good here, phew.

PR #1371, opened 2016-11-04
Maintenance notice for planned outage due to Salesforce

rtyley:
Salesforce having a 2h45m read-only maintenance period on Sunday means we can't register new membership sign-ups for that period. So, we'll redirect any page that leads to needing sign-up functionality (eg Supporter Landing Pages, Checkout pages, but not event pages) to a new Maintenance page (https://membership.theguardian.com/maintenance):

image

...this directs people to Contributions: **https://contribute.theguardian.com/?INTCMP=mem_maint**

**I know the page could look more beautiful, but bear in mind that it will only be shown for 3 hours, and we need to get this sorted before the weekend! So please review this PR based on whether the code/approach is good, and then suggest style changes in a separate PR if the mood takes you.** It would probably be nice to put an amusing Guardian quote in there if anyone can think of one.

These are the urls that will be affected:

* https://membership.theguardian.com/uk/supporter (all locales)
* https://membership.theguardian.com/join/supporter/enter-details
* All other tiers (Partner, Patron)

As this page needs to go up in the middle of the night (for our London-based dev team), the `PlannedOutage` class holds the times that the Maintenance page will be in effect, and for increased confidence that it's going to work I've added a staff-only page:

https://membership.theguardian.com/planned-outages

...which just says:

image

```
Next planned outage is
6 November 2016, 2am - 6 November 2016, 4.45am
starting in
38 hours, 1 minute and 40 seconds
```

Salesforce planned outage:
----
https://status.salesforce.com/status/maintenances/9
```
RELATED INSTANCES: EU0
RELATED SERVICES: Core Service
PLANNED START TIME: 02:00, Nov 06 GMT
PLANNED END TIME: 04:45, Nov 06 GMT
TYPE: Scheduled Maintenance
AVAILABILITY: This instance will be read only during this maintenance window.
```

cc @paulbrown1982 @philwills @Ap0c @svillafe

### DIFF
--- a/frontend/app/actions/ActionRefiners.scala
+++ b/frontend/app/actions/ActionRefiners.scala
@@ -15,6 +15,7 @@ import play.api.libs.concurrent.Execution.Implicits.defaultContext
 import play.api.mvc.Results._
 import play.api.mvc.Security.AuthenticatedBuilder
 import play.api.mvc._
+import utils.PlannedOutage
 import views.support.MembershipCompat._
 
 import scala.concurrent.Future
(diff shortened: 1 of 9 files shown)


rtyley:
Maintenance page is live on PROD:
image

...and as intended, urls like https://membership.theguardian.com/us/supporter are _not_ currently redirecting.

The planned outages page is also showing the correct info:

image

https://membership.theguardian.com/planned-outages

rtyley:
This is corresponding PR for subscribe.theguardian.com : https://github.com/guardian/subscriptions-frontend/pull/670

PR #1373, opened 2016-11-04
Autoplay/Loop Video on Supporter Landing Page

Ap0c:
# Proposal

We want the second video on the supporter landing page to autoplay and loop. If autoplay is not possible (iOS), we want the default behaviour to occur (i.e. the poster appears over the top, and the user clicks to play).

# Changes

- Added `autoplay` and `loop` parameters to the scala `Video` object, to allow any video to be set to autoplay or loop.
- Set classes on the YouTube iframe that signify autoplaying or looping functionality.
- When loop is set, add event listener for end of video that restarts playback.
- When autoplay is set, fire playback immediately rather than waiting for click.

**Note**: The diff for `videoOverlay.js` makes the changes look a lot more complicated than they are, it might be easier to look at the two versions of the file side-by-side.

**Note Part II**: I will test these changes across browsers and devices on Monday (want to double-check iOS behaviour).

@rtyley @svillafe @rupertbates @JustinPinner

### DIFF
--- a/frontend/assets/javascripts/src/modules/videoOverlay.js
+++ b/frontend/assets/javascripts/src/modules/videoOverlay.js
@@ -64,13 +64,14 @@ define(['src/modules/raven'],function(raven) {
 
     }
 
+    // Sets up click-to-play on overlay, or autoplays if applicable.
     function playerReady(player, playerApi, playerOverlay, autoplay) {
 
         playerOverlay.addEventListener('click', function(event) {
 
             event.preventDefault();
             playVideo(player, playerApi, playerOverlay);
-            
+
         });
 
         if (autoplay) {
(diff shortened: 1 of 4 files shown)


frontend/assets/javascripts/src/modules/videoOverlay.js
```diff
@@ -27,36 +27,66 @@ define(['src/modules/raven'],function(raven) {
             var playerApi;
 
             if(playerIframe && playerOverlay) {
-                playerApi = new YT.Player(playerIframe, {
-                    events: {
-                        'onReady': function() {
-                            playerReady(player, playerApi, playerOverlay);
+
+                var autoplay = playerIframe.classList.contains('autoplay');
+                var loop = playerIframe.classList.contains('autoplay');
```
rupertbates: Shouldn't this be 'loop' not 'autoplay'?

Ap0c: Yup, it definitely should 😮.

In fact the use of `classList` will break in older versions of IE. I know that's a small chunk of the audience, but I've added a workaround just to be safe.

42f3de0

Ap0c:
When I tested this on iOS, I noticed that the poster was disappearing, and added a fix to remedy this: 61aff8f. Apart from that it ran ok.

Ap0c:
Videos look like they're playing ok after the changes, including on iOS 👍.

PR #1374, opened 2016-11-08
update copy on supporter home page

JuliaBellis:
Updated line about ad revenues falling fast to say 'ad revenues across the media' are falling fast. This change has outperformed the original message for contributions.


rtyley:
That affects here, right:

image

:+1: , will merge this now.

rtyley:
Verified, it's in place on https://membership.theguardian.com/uk/supporter : 

image

PR #1375, opened 2016-11-16
Make Terms & Conditions region sensitive

JustinPinner:
We have to supply different T&C text to different territories. This PR attempts that by detecting which countryGroup is in effect and writing suitable links into the footer of the page.

Also, in the US (only) we have to display very specific text at the point of committing to a recurring payment arrangement.

The changes so far make these things work, for the specific pages discussed in the corresponding Trello card. However, I still have to adjust the T&C links that appear in the footer of the `/join` page.

### Australia
screen shot 2016-11-15 at 16 58 15
--

### US
screen shot 2016-11-15 at 16 57 28
--

### International
screen shot 2016-11-15 at 16 58 58
--

### UK
screen shot 2016-11-15 at 16 57 01

### Australia
screen shot 2016-11-14 at 17 23 52
--

### US
screen shot 2016-11-14 at 17 23 31
--

### International
screen shot 2016-11-14 at 17 22 46
--

### UK
screen shot 2016-11-14 at 17 23 10

### Mobile
screen shot 2016-11-16 at 15 03 17
--

### Desktop
screen shot 2016-11-16 at 15 04 07

## Changing the billing country on the form
Regardless of which 'edition' of the page you land on, if the billing address country code is changed, the T&Cs are updated accordingly. This is easier to show in a video:

territorialt cs


AWare:
It's possible to change the country group on the page itself. People that come to the page from the international landing page, or a shared country specific landing page would not see the right t&c.

frontend/app/configuration/Links.scala
```diff
@@ -12,7 +13,15 @@ object Links {
 
   val membershipFront = "http://www.theguardian.com/membership"
 
-  val membershipTerms = "http://www.theguardian.com/info/2014/sep/09/guardian-membership-terms-and-conditions"
+  def membershipTerms(countryGroup: Option[CountryGroup] = None) = {
+    countryGroup match {
+      case Some(cg) if cg.id.toLowerCase == "uk" => "https://www.theguardian.com/info/2014/sep/09/guardian-membership-terms-and-conditions"
```
johnduffell: why not just case Some(UK) etc (for extra compile safety and clarity)

frontend/app/views/fragments/form/submit.scala.html
```diff
@@ -44,3 +46,16 @@
         </div>
     </div>
 </fieldset>
+
+@countryGroup.map{ cg =>
+    @if(cg.id.toLowerCase == "us") {
+        <div class="us-legal">
+            NOTICE REGARDING AUTOMATIC RENEWAL: IF YOU DO NOT CANCEL YOUR MEMBERSHIP PRIOR TO THE END OF YOUR INITIAL
```
paulbrown1982: It's a shame this has to be in capitals. Did we query back to double check this is absolutely required?

JustinPinner: That question is going around again to double-check, but current understanding is that it's a legal requirement to be upper case in the US :(

frontend/app/configuration/Links.scala
JustinPinner: Can you compare a CountryGroup to a Country without assessing either the DefaultCountry or Countries props of the group?

JustinPinner:
Ok, time for 👍s @AWare @rtyley @johnduffell? Thanks!

johnduffell:
I have no objection! 👍 not sure about the capitals either though, but if we're supposed to be legally bound, then so be it

paulbrown1982:
👍 LGTM

JustinPinner:
Updated to accommodate the new checkout form after merge conflict resolution. cc @svillafe

JustinPinner:
Post deploy evidence (switching between US & UK):
territorialt cslive

frontend/app/views/fragments/form/submit.scala.html
```diff
@@ -44,3 +47,11 @@
         </div>
     </div>
 </fieldset>
+
+@countryGroup.map{ cg =>
+    @if(Seq("us","um","vi").contains(cg.id.toLowerCase)) {
```
rtyley: This line doesn't work quite as might have been intended- the `"um"` and `"vi"` terms will never be activated, because we're looking at the id of a `CountryGroup` here, not a `Country` - there are only 6 CountryGroups  _(UK, US, Canada, Australia, NewZealand, Europe, RestOfTheWorld)_, and none of them have `"um"` and `"vi"` as their id.

However, the line _will_ match the `US` country group, so works reasonably well - and when the user selects their country in the form, the javascript will update the visibility of the line appropriately.

On a related note, when [`CountryGroup` was introduced](https://github.com/guardian/membership-common/commit/8d0d5972accd3a78fa166947cefbcb50a02ac3b3#diff-e19ee1b21c6b25cad28d96066988ab86R261), these countries should have been included in the `US` CountryGroup, not _RestOfTheWorld_:

* `Country("VI", "United States Virgin Islands")`
* `Country("VC", "Saint Vincent & The Grenadines")`

paulbrown1982: @rtyley that might have been fortunate in the case of Guardian Weekly pricing - as it costs more to post to those places and so any pages which resolve prices to readers from those places would have to quote the international (RestOfTheWorld) prices.

PR #1376, opened 2016-11-16
Switch to using deb packaging and systemd

rtyley:
Further to #1367, this removes the old `upstart` configuration in favour of `systemd` and means we
can use modern [Amigo-generated AMIs](https://amigo.gutools.co.uk/recipes/xenial-membership).

The configuration here is based upon the similar config found in Contributions frontend:

* https://github.com/guardian/contributions-frontend/blob/9830dbf1/build.sbt#L77-L101
* https://github.com/guardian/contributions-frontend/blob/9830dbf1/cloud-formation/cfn.yml#L105-L156

The relevant changes in Contributions that resulted in this config include:

* guardian/contributions-frontend@ec38edc
* guardian/contributions-frontend#198

cc @guardian/membership-and-subscriptions

PR #1378, opened 2016-11-18
Fix AWS Cloudwatch config

jacobwinch:
@paulbrown1982 @rtyley 

Tested on a single box (after removing from the auto-scaling group) and can see logs are now being picked up by Cloudwatch.

Note that I removed the __BUILD from the log stream name, as I could see that Roberto had removed this from the Cloudformation file.


johnduffell:
fair enough! 👍 🚤

rtyley:
Damn, I thought I did this stuff with #1376, but I guess that change got lost before I merged the branch :angry:  - thanks for fixing it!

PR #1381, opened 2016-11-23
Update PhantomJS to v2

Ap0c:
# Problem

The old version of PhantomJS (pre-v2) does not run on macOS sierra. This is a problem because we have pre-commit hooks that run PhantomJS tests locally.

# Solution

Upgrade PhantomJS to latest version (2.1), which is now called `phantomjs-prebuilt` in npm. Also update the corresponding Karma launcher to the most recent version, as the old version did not support PhantomJS v2.

These changes will also take effect in the test runner, i.e. TeamCity will now use the most recent versions as well.

# To Update Locally

You should just be able to `cd` into `frontend` and run:

```
npm update
```

@svillafe  @guardian/membership-and-subscriptions


rtyley:
:+1:

Ap0c:
Change to `package.json`, so user-facing site shouldn't be affected.

Looks like test-runners are happy with the package updates.

PR #1383, opened 2016-11-25
Allow removal of legacy rate plans as part of upgrade

jacobwinch:
When processing a membership upgrade, we should always remove the 'current rate plan' from the Sub before adding the new rate plan. This PR fixes a bug which failed to remove the 'current rate plan' if it was a legacy plan (i.e. if it was not in the current catalog).

Failure to remove the current rate plan means that Subs end up with overlapping rate plans; at best (in the case that the original subscription was 'Friend'), this is creates mess in Zuora and Salesforce. If the original plan was a paid legacy plan, this could lead to us double-charging a customer.

This PR also includes some improvements to logging in upgrade scenarios.

@johnduffell @paulbrown1982

### DIFF
--- a/frontend/app/services/MemberService.scala
+++ b/frontend/app/services/MemberService.scala
@@ -73,11 +73,8 @@ object MemberService {
     }
   }
 
-  def getRatePlanIdsToRemove(current: Seq[SubIds],
-                             planFinder: ProductRatePlanId => Boolean,
-                             discounts: DiscountRatePlanIds): Seq[RatePlanId] = current.collect {
+  def getDiscountRatePlanIdsToRemove(current: Seq[SubIds], discounts: DiscountRatePlanIds): Seq[RatePlanId] = current.collect {
     case discount if discount.productRatePlanId == discounts.percentageDiscount.planId => discount.ratePlanId
-    case plan if planFinder(plan.productRatePlanId) => plan.ratePlanId
   }
 }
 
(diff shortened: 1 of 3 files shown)


frontend/app/services/MemberService.scala
```diff
@@ -450,10 +447,14 @@ class MemberService(identityService: IdentityService,
...
+    val currentRatePlan = sub.plan.id
+    logger.info(s"Current Rate Plan is: ${sub.plan.productName}. Plan to remove is: ${currentRatePlan}")
 
     (newRatePlan |@| ids) { case (newPln, restSub) =>
-
-      val plansToRemove = getRatePlanIdsToRemove(restSub, catalog.find(_).isDefined, discountIds)
+      val discountsToRemove = getDiscountRatePlanIdsToRemove(restSub, discountIds)
+      if (!discountsToRemove.isEmpty) logger.info(s"Discount Rate Plan ids to remove when upgrading are: ${discountsToRemove}")
+      val plansToRemove = currentRatePlan +: discountsToRemove
+      if (plansToRemove.isEmpty) logger.error(s"plansToRemove is empty - this could lead to overlapping rate plans on the Zuora sub: ${sub.id}")
```
johnduffell: 😕

johnduffell:
good work! 👍 🚀

jacobwinch:
For the record I think this must have got introduced with #1332 when we started using the new catalog.

johnduffell:
actually not, the catalog.find call was there before that: https://github.com/guardian/membership-frontend/pull/1332/files#diff-abb6c51f77945af0f3a1eca1252ea19bL471

johnduffell:
I think we discussed this after I made the above comment, and it turned out the new catalog.find was being more strict than the old one, so Jacob's comment was actually correct.

PR #1385, opened 2016-11-25
AB Test Logic for the "New Checkout Flow"

svillafe:
## Description

Currently, there is two version of the checkout flow on production, but only one has traffic. This PR implement the logic of the  A/B Test

## Flow's details

1 - The user clicks the "become a supporter button"

2 - We check if the user has the cookie: "ab-checkout-flow"

3 - If not we select A or B randomly and we store that value in the cookie.

4 - If the user is not log in and the test selected is B, we redirect the user to **http://profile...clientId=membersB**. If the test selected is A, we redirect the user to **http://profile...clientId=members**

5 - After the user creates his or her account, we check the cookie again and show him/her the correct checkout page.

Trello card: https://trello.com/c/PpnCnR0N/150-implement-the-ab-test-for-the-new-checkout-flow

### DIFF
--- a/frontend/app/actions/Fallbacks.scala
+++ b/frontend/app/actions/Fallbacks.scala
@@ -1,11 +1,48 @@
 package actions
 
+import com.gu.memsub.subsv2.CatalogPlan.PaidMember
+import com.gu.memsub.subsv2.MonthYearPlans
 import com.gu.salesforce.PaidTier
 import play.api.mvc.Results._
-import play.api.mvc.{Call, RequestHeader}
+import play.api.mvc.{Call, Cookie, RequestHeader}
 import play.twirl.api.Html
 import configuration.Config
+import views.html.joiner.form.paymentB
+import views.support.{CountryWithCurrency, IdentityUser, PageInfo}
 
+import scala.util.Random
+
+
+sealed trait CheckoutFlowVariant {
+  val testId: String
+  val identitySkin: String
+}
+
+object CheckoutFlowVariant {
+  val cookieName = "ab-checkout-flow"
+
+  val all = Seq[CheckoutFlowVariant](A,B)
+
+  def deriveFlowVariant(implicit request: RequestHeader): CheckoutFlowVariant =
+    getFlowVariantFromRequestCookie(request).getOrElse(CheckoutFlowVariant.all(Random.nextInt(CheckoutFlowVariant.all.size)))
+
+  def getFlowVariantFromRequestCookie(request: RequestHeader): Option[CheckoutFlowVariant] = for {
+    cookieValue <- request.cookies.get(cookieName)
+    variant <- CheckoutFlowVariant.lookup(cookieValue.value)
+  } yield variant
+
+  case object A extends CheckoutFlowVariant {
+    override val testId: String = "test-A"
+    override val identitySkin: String = "members"
+  }
+
+  case object B extends CheckoutFlowVariant {
+    override val testId: String = "test-B"
+    override val identitySkin: String = "membersB"
+  }
+
+  def lookup(name: String): Option[CheckoutFlowVariant] = all.find(_.testId == name)
+}
 
 object Fallbacks {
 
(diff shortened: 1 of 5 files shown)


frontend/app/controllers/Joiner.scala
```diff
@@ -33,8 +33,9 @@ import views.support.Pricing._
 import views.support.TierPlans._
 import views.support.{CheckoutForm, CountryWithCurrency, PageInfo}
 import views.support.MembershipCompat._
+
 import scala.concurrent.Future
-import scala.util.Failure
+import scala.util.{Failure, Random}
```
rtyley: Unnecessary import, please clean your imports with IntelliJ!

frontend/app/actions/Fallbacks.scala
```diff
@@ -1,11 +1,48 @@
...
 
+import scala.util.Random
+
+
+sealed trait CheckoutFlowVariant {
+  val testId: String
+  val identitySkin: String
+}
+
+object CheckoutFlowVariant {
```
rtyley: These `CheckoutFlowVariant ` classes probably don't belong in `package actions`. Perhaps use IntelliJ to move them to a new `abtests` package?

frontend/app/configuration/Config.scala
```diff
@@ -47,8 +49,13 @@ object Config {
   def idWebAppSigninUrl(uri: String): String =
     (idWebAppUrl / "signin") ? ("returnUrl" -> s"$membershipUrl$uri") & idSkipConfirmation & idMember
 
-  def idWebAppRegisterUrl(uri: String): String =
+  def idWebAppRegisterUrl(uri: String, abTestVariant: CheckoutFlowVariant = A): String ={
+
+    val idMember = "clientId" -> abTestVariant.identitySkin
+
     (idWebAppUrl / "register") ? ("returnUrl" -> s"$membershipUrl$uri") & idSkipConfirmation & idMember
```
rtyley: Note that `idMember` is hiding the wider-scoped `idMember` variable here. Perhaps better to inline this variable, or use a different name?

frontend/app/configuration/Config.scala
```diff
@@ -49,14 +49,9 @@ object Config {
...
-  def idWebAppRegisterUrl(uri: String, abTestVariant: CheckoutFlowVariant = A): String ={
-
-    val idMember = "clientId" -> abTestVariant.identitySkin
-
-    (idWebAppUrl / "register") ? ("returnUrl" -> s"$membershipUrl$uri") & idSkipConfirmation & idMember
-  }
-
-
+  def idWebAppRegisterUrl(uri: String, abTestVariant: CheckoutFlowVariant = A): String =
+    (idWebAppUrl / "register") ? ("returnUrl" -> s"$membershipUrl$uri") & idSkipConfirmation & "clientId" -> abTestVariant.identitySkin
```
rtyley: I'm not sure what the evaluation precedence is on this, so just check you stil get the right value, or surround `"clientId" -> abTestVariant.identitySkin` with parentheses!

svillafe: @rtyley I tested in dev and the evaluation was correct. Just in case (and to improve the 	
legibility of the code), I added a pair of parentheses.

rtyley:
:+1:

svillafe:
The code is working in PROD. 
The cookie is being set when the user clicks the "Become a Supporter Button"

picture 27

frontend/app/actions/Fallbacks.scala
```diff
@@ -23,7 +23,11 @@ object Fallbacks {
   def notYetAMemberOn(implicit request: RequestHeader) =
     redirectTo(controllers.routes.Joiner.tierChooser()).addingToSession("preJoinReturnUrl" -> request.uri)
 
-  def chooseRegister(implicit request: RequestHeader) = SeeOther(Config.idWebAppRegisterUrl(request.uri))
+  def chooseRegister(implicit request: RequestHeader) = {
+    val flowSelected = CheckoutFlowVariant.deriveFlowVariant(request)
```
svillafe: @rupertbates Here we choose the variant randomly.

PR #1386, opened 2016-11-25
Give a better message when users can not upgrade due to locale

rtyley:
If a user is a Supporter and currently paying in non-GBP currency, they can't upgrade to a higher tier. We don't have prices set in non-GBP currency set for any tier other than Supporter.

We've had some emails from users who are disappointed about this, so this message aims to help them out, and allow them to support the Guardian in another way.

We hit Contributions with the url **https://contribute.theguardian.com/?INTCMP=mem_tier_change_unavailable**

image

cc @philwills


svillafe:
👍

PR #1388, opened 2016-11-29
CSS bug fix for the checkout page, flow B

svillafe:
Bug Fix for the checkout page in flow B.

## Before
picture 30

## After

picture 31


svillafe:
It is working on PROD

picture 32

PR #1389, opened 2016-11-29
Signout-to-register flow for "switch user" scenario at supporter checkout

JustinPinner:
When someone is signed in and completing the payment (checkout) form to become a supporter, they may realise they're logged in with the wrong account if they use a shared device.

We should provide a route by which they can log out and then log back in (or create a new account) and resume their journey into supporterdom.

The identity service rejects signin as an onward journey after signout so now we'll send the user to the membership register page (which also includes a sign in link), with a return url copied through from the point at which they were completing the supporter checkout form.

The flow can therefore be as follows;

`sign in -> become a supporter -> join (checkout) -> sign out -> register -> join (checkout)`
or
`sign in -> become a supporter -> join (checkout) -> sign out -> sign in (via the link on the register page) -> join (checkout)`

@svillafe I've passed the checkout flow variant through, but I wonder if this is needed. What do you think?

I've also made this an additional feature, rather than replacing the current behaviour whereby clicking sign out from the same link but in other places (e.g. friend sign up or free to paid upgrade) returns the reader to the dotcom front page as per current functionality.

cc @Ap0c @rtyley @rupertbates 

 https://trello.com/c/TEaX9RUB/133-fix-sign-out-link-on-membership-checkout-page - 
upload_4_11_2016_at_14_31_30

### DIFF
--- a/frontend/app/views/fragments/joiner/signedInAs.scala.html
+++ b/frontend/app/views/fragments/joiner/signedInAs.scala.html
@@ -1,8 +1,8 @@
 @import abtests.CheckoutFlowVariant
 
-@(returnUri: String, inline: Boolean = false, switchUser: Boolean = false, flowVariant: CheckoutFlowVariant = CheckoutFlowVariant.A)
+@(returnUri: String, inline: Boolean = false, flowVariant: CheckoutFlowVariant = CheckoutFlowVariant.A)
 
-@import configuration.Config.{idWebAppSignOutThenInUrl,idWebAppSignOutThenRegisterUrl}
+@import configuration.Config.idWebAppSignOutThenRegisterUrl
 
 <p class="text-note">
     You're signed in as <strong class="js-user-displayname"></strong>.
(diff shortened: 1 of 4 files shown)


frontend/app/views/fragments/joiner/signedInAs.scala.html
```diff
@@ -1,12 +1,18 @@
-@(returnUri: String, inline: Boolean = false)
+@import abtests.CheckoutFlowVariant
 
-@import configuration.Config.idWebAppSignOutThenInUrl
+@(returnUri: String, inline: Boolean = false, switchUser: Boolean = false, flowVariant: CheckoutFlowVariant = CheckoutFlowVariant.A)
```
rtyley: You can remove the `switchUser: Boolean` parameter, and use `idWebAppSignOutThenRegisterUrl()` below only - this template fragment (`fragments.joiner.signedInAs()`) is only used in [4 places in our codebase](https://github.com/guardian/membership-frontend/search?utf8=%E2%9C%93&q=fragments.joiner.signedInAs) - friend-signup, payment A/B, & free-paid upgrade - and all of those are fine to use this new behaviour, returning the user to where they came from once they're signed in!

JustinPinner: Ah good. That'll simplify things.

rtyley:
Cool, didn't realise we could get rid of the old `idWebAppSignOutThenInUrl` as well!

rtyley:
@JustinPinner could you check if this works in production? I just tried, but got the old behaviour - I was immediately redirected to https://www.theguardian.com/uk, without even going to profile.theguardian.com/register

JustinPinner:
Identity also block /register. Just not when you run locally. Mark Butler is removing that limitation at the moment. Should begin working soon.

PR #1393, opened 2016-11-30
remove prod dep on UAT zuora

johnduffell:
So previously the UAT catalog had to read successfully for prod boxes to come up.  Since membership-frontend has some kind of leak, it eventually goes down (usually in the middle of the night) when the uat catalog becomes invalid.

We had a load of actions from the retro to solve this but we never actually did it, resulting in it happening again.

This should solve that problem, although @JustinPinner had a related problem of dev being invalid stopping people develop.  That would still need addressing after this.

I've tested this and I can sign up fine as long as I don't use any test users.  The healthcheck passes.  As soon as I use a test user the page returned a 500 error.  If prod zuora fails to read, then the healthcheck falls over straight away with a 500 server error, so we are still safe from that.


frontend/app/services/TouchpointBackend.scala
```diff
@@ -107,10 +107,8 @@ object TouchpointBackend {
     )
   }
 
-  val Normal = TouchpointBackend(BackendType.Default)
-  val TestUser = TouchpointBackend(BackendType.Testing)
-
-  val All = Seq(Normal, TestUser)
+  lazy val Normal = TouchpointBackend(BackendType.Default)
```
rtyley: Does `Normal` want to be a lazy val here? I'm guess that we always want it?

frontend/app/services/TouchpointBackend.scala
```diff
@@ -107,10 +107,8 @@ object TouchpointBackend {
...
+  lazy val TestUser = TouchpointBackend(BackendType.Testing)
```
philwills: I'd suggest at a minimum there should be a comment explaining why it's important that this is lazy.

rtyley:
Just to explain this fix to myself, making `TouchpointBackend.TestUser` lazy means that even if UAT is badly configured, the `object TouchpointBackend` can still complete it's class initialisation, and thus the box can come up.

Since changes that appear in UAT can progress into changes to PROD, I think it's good to be aware of problems with them.

Post-deploy acceptance testing we run after PRs are merged and deployed uses test-users, so I'm guessing although a deploy may succeed, the post-deploy acceptance testing will fail if UAT is bad, @mario-galic ?

johnduffell:
@philwills : @jacobwinch @pvighi and I were discussing that we should add custom fields to the catalog so we can put hints on the things we should be looking for.  Then we can ignore anything extra in the catalog, as the main problem is when something new is added it all falls over.
I think the main surprises that catch people out is 1) prod membership needs UAT zuora to start, 2) adding something new stops the existing stuff being readable and possible 3) membership depends on subscription products being readable and vice versa.
This PR solves 1, the hints is one way to solve 2, and 3 could be solved by breaking the catalog up and storing things as eithers, or maybe another solution.  It's definitely something we need to bring into the next quarter, as that will be themed around costs and health.

@rtyley yes it would indeed break post deploy testing, however if it's the middle of the night that's less important.  If you've just shipped something that's broken something, you should notice fairly quickly and hopefully roll back if necessary, so I think the tradeoff is mostly in our favour.  What do you think?

frontend/app/services/TouchpointBackend.scala
johnduffell: good call, done that phil!

frontend/app/services/TouchpointBackend.scala
johnduffell: we do, but the health check checks that.  And if you're developing locally and only using UAT, do you really want to insist that DEV is valid?

johnduffell: although I should point out that even with this change, if DEV zuora is invalid, you can't use UAT only, as going to mem.thegulocal with a test identity user logged in still tries to access the non UAT zuora for some reason.

rtyley: > if you're developing locally and only using UAT, do you really want to insist that DEV is valid?

When developing locally, you're normally using Zuora DEV, so I think you would?


In Membership PROD:
Normal=Zuora PROD
Test = Zuora UAT

In Membership DEV:
Normal=Zuora DEV
Test=Zuora UAT

rtyley:
> yes it would indeed break post deploy testing, however if it's the middle of the night that's less important. If you've just shipped something that's broken something, you should notice fairly quickly and hopefully roll back if necessary, so I think the tradeoff is mostly in our favour. What do you think?

Agreed, I just wanted to describe what the expected behaviour is. I think it's a little unfortunate that someone who merges their PR may get a slightly misleading impression from a post-deploy testing failure that their PR caused the problem - when in fact it was a separate Zuora issue.

johnduffell:
Yeah that's fine I wasn't disagreeing ;) but yes totally agree we can make things smarter and more robust, I thought about a periodic validation of all the catalogs on a timer and an email (or pagerduty for prod) if they are invalid.  But this is cheaper in the short term.

frontend/app/services/TouchpointBackend.scala
johnduffell: well, as soon as you try to use it, it will fall over anyway., I don't see there's a big benefit to blowing up the whole app (although in reality that's what happens anyway due to other parts of the code)
I think only the thing you are actually using should need to work, unless you are needing as a gate/test in which case you should have something more specific than just blowing up the app as a side effect.  The healthcheck does check the normal so we're good to go there.
Although I guess it's more a balance, just which way we'd prefer to err on the side.  But if someone breaks dev it would still be nice to have the ability to test locally in UAT in order to get changes out.

rtyley:
:+1:

johnduffell:
seems prod had some instability around when the boxes were coming up, tested partner signup manually and it worked

PR #1368, opened 2016-11-03
Remove Navigation On Checkout

Ap0c:
# Proposal

It's believed that the nav elements on the checkout page are a source of users bouncing on and not completing the form. Removing these may reduce this behaviour and **increase conversion**. This PR removes the link to the main site at the top of the header, and the primary navigation bar at the bottom of the header.

### Before - Desktop

checkout-nav-desktop-before

### Before - Mobile

checkout-nav-mobile-before

### After - Desktop

checkout-nav-desktop-after

### After - Mobile

checkout-nav-mobile-after

# Changes

- Add a shorter masthead image (`72px` vs `130px`).
- Add a new members logo SVG that's more tightly cropped `guardian-members-tightcrop` (the current one has surrounding whitespace).
- Add a new `simpleHeader` template for the new header.
- Add a new sass file `_simpler-header` to style the new header template.
- Tweak the `controlNavigation` template to allow hiding of the link to the main site.
- Tweak the `main` and `payment` templates to include a `simpleHeader` parameter, allowing a choice between the full header and the new shorter one.

@rtyley @svillafe @JustinPinner @rupertbates

PR #1394, opened 2016-12-01
Remove Profile Icon from Checkout Page

Ap0c:
# Plan

We want to remove the profile icon from the checkout page, while leaving it intact for the rest of the site. This includes removing the identity menu from this page.

Involves building on #1368.

# Changes

- Remove control navigation template from the simple header, we're now not using anything from it (`simpleHeader.scala.html`).
- No need for passing `mainSiteLink` through to control navigation now that we don't use it (`controlNavigation.scala.html`).
- Shifted identity menu code into a submodule (`main.js`, `identityPopup.js`, `identityPopupDetails.js`, `setup.es6`, `identityPopup.spec.js`).
- Added presence of simple header to guardian page object, and check for it before trying to initialise identity menu code (`simpleHeader.scala.html`, `setup.es6`).

# Screenshots

**Before**:

with-profile

**After**:

without-profile

**Normal Header Outside Checkout**:

normal-page

@svillafe @rupertbates @rtyley @JustinPinner


frontend/app/views/fragments/global/controlNavigation.scala.html
```diff
@@ -75,16 +74,14 @@
...
-                <span class="control__item">
-                    @fragments.inlineIcon("g-mark", List("icon-inline--large"))
-                </span>
-                <span class="control__item">
-                    <span class="control__info">the guardian</span>
-                </span>
-            </a>
-        </li>
-    }
+    <li class="nav-control__item nav-control__item--last hidden-mobile">
```
rupertbates: Shouldn't this code be deleted now?

Ap0c: Ah, no, sorry, I should have made that clearer; this code isn't specific to the simple header. It was always shown in the main header, I just [added a flag](https://github.com/guardian/membership-frontend/pull/1368/files#diff-a50ebb05028e3a9ba0785d1ba9dbc24b) so we could have the option of hiding it for the simple one.

However, now that the simple header no longer uses anything from that file, the flag is no longer necessary 😄.

Ap0c:
Looks to be working ok...

**Supporter Page Without Profile Icon:**

supporter-flow

**Successful Partner Sign-Up (also works for Supporter):**
partner-success

Will investigate why post-prod tests are failing (again).

Ap0c:
Temporarily broke post-deploy tests by deleting the display name element. Fixed in #1399.

PR #1395, opened 2016-12-01
Ignore weekly one off plan instead of crashing on startup

pvighi:
Update membership-common to the latest version which ignores weekly one off plans in the zuora catalog instead of crashing when reading it
@paulbrown1982 @jacobwinch @johnduffell


johnduffell:
although hopefully it shouldn't actually fall over after the other change to make prod not depend on UAT!
👍 🚀

PR #1396, opened 2016-12-01
upwrap email address from its option

johnduffell:
When I made email address optional in our model reading from salesforce, I didn't spot that membership uses it in its thank you page

As such we were telling users we have emailed them on Some(joe.bloggs@guardian.co.uk) 😨 

I've fixed it up and made it all helpful too if the option is not filled, but since this would have 500'd before and it's basically an unlikely error, I'm not sure if `.get` would be good enough? thoughts?

image


rtyley:
:+1:

PR #1399, opened 2016-12-05
Fix Post-Deploy Tests - Display Name Check

Ap0c:
# Problem

I broke the post-deploy tests in #1394 (sorry!). By removing the profile display name field on the checkout page, the test that checks the display name against the test user fails ("And should be logged in with their Identity account.").

# Solution

Fortunately, there is another location on the checkout page that shows the display name, the only difference being that it is not shown in lower case. Removing the `toLowerCase` method call allows the test to pass again 😃 (thanks Mario for figuring this out!).

@mario-galic @rupertbates @svillafe @JustinPinner


Ap0c:
The tests are working again!

PR #1400, opened 2016-12-05
Tracking billing period (monthly/annual) in the payment button.

svillafe:
# Tracking billing periods in the checkout page

## Why?
In order to A/B test the price of the pay button, we should be able to track if the users are converting using the monthly price or the annual price.

## Changes
* `billingPeriodChoice.js`: After a change event takes place, I am sending the type of pricing as a payload to the listeners.
* `submitButton.js`: The submit module receives the payload and appends it to the button's `data-metric-label`

## Images

We are sending events labels correctly.

picture 39


svillafe:
It is working on PROD.

PR #1402, opened 2016-12-06
Initialise TouchpointBackend.TestUser early, so it's ready when the post-deployment testing happens

svillafe:
# UAT settings don't block deployment

## Problem
When a developer wants to deploy code on PROD, sometimes a race condition takes place. The test run so fast that the UAT environment is not initialised at that point and as a consequence, it fails.

## Proposed solution
Using a **Future** we initialized the UAT environment and in this way it will be up and running at the moment the test will be executed.


svillafe:
Please @rtyley review this PR.

frontend/app/services/TouchpointBackend.scala
```diff
@@ -111,6 +113,12 @@ object TouchpointBackend {
   lazy val Normal = TouchpointBackend(BackendType.Default)
   lazy val TestUser = TouchpointBackend(BackendType.Testing)
 
+  val f = Future {
```
rtyley: `val f` is unused, so you can just remove that - no needs a reference to this right now, unless you want to somehow tie this into the healthcheck?

svillafe: 😅

rtyley:
> The test run so fast that the UAT environment is not initialised at that point and as a consequence, it fails.

It's not so much that the test is running _fast_ - it's more that the membership-frontend server has not even _tried_ to initialise the `TestUser` Touchpoint backend (which includes reading the Zuora product catalog and Salesforce authentication) until the very moment when the test starts!

See, for instance https://github.com/guardian/membership-frontend/pull/1400#issuecomment-265128197 and https://saucelabs.com/tests/aef3aee5707246e784a994de83770979 :

0008screenshot

and the [error in the logs](https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logEventViewer:group=FrontendLogs-PROD;stream=2016-12-06/i-5a1827b2/frontend.log;refid=33027937935553329732067548636287954586427957001815916559;reftime=1481023958685) that indicates our connection to Salesforce UAT was not fully initialised when the post-deploy test began it's run: 

```
2016-12-06 11:32:37,855 [ForkJoinPool-2-worker-25] com.gu.salesforce.SimpleContactRepository$$anon$1[Scalaforce.scala:149] [34mINFO: Trying to authenticate with Salesforce UAT...
^^ this is happening literally as the test user request for /join/partner/enter-details comes in!
...
2016-12-06 11:32:37,925 [application-akka.actor.default] application[HttpErrorHandler.scala:216] ERROR: 
! @72a6mlp1i - Internal server error, for (GET) [/join/partner/enter-details] ->
play.api.UnexpectedException: Unexpected exception[UnknownHostException: dummy-url.salesforce.com: Name or service not known]
```

This problematic behaviour was introduced with https://github.com/guardian/membership-frontend/pull/1393, which was a fix aimed at making sure that a bad UAT could not take down production. As @johnduffell said when he saw this problem, it looks like we've swapped an infrequent fatal problem with one that's more frequent but merely irritating!

The hope with the change in this PR is that it should mean than `TouchpointBackend.TestUser` is initialised much earlier (ie when the server starts up), and so the TouchpointBackend should be ready when the post-deployment acceptance-test-user comes to call.

There's still potentially a bit of a race condition, because we aren't holding back the healthcheck for the `TouchpointBackend.TestUser` to fully initialise - so the post-deployment acceptance test might still fail, but it's much less likely.

rtyley:
:+1:

frontend/app/services/TouchpointBackend.scala
```diff
@@ -111,6 +113,12 @@ object TouchpointBackend {
   lazy val Normal = TouchpointBackend(BackendType.Default)
   lazy val TestUser = TouchpointBackend(BackendType.Testing)
 
+  Future {
+    logger.info(s"TouchpointBackend.TestUser is lazily initialised to ensure bad UAT settings can not block deployment to PROD. Initalisation starting...")
```
johnduffell: good use of logging and documenting at the same time!

johnduffell:
👍 🚀 thanks!

svillafe:
The `post-deployment` tests are working and the site is working correctly in **PROD**.

johnduffell:
I know where the dummy url comes from now.. so when we have time we can go there to fix properly  https://github.com/guardian/membership-common/blob/1acd1576d3fb7f203c11643a1c0fa8f6e587f156/src/main/scala/com/gu/salesforce/Scalaforce.scala#L64

PR #1403, opened 2016-12-07
Remove progress bar

svillafe:
# Delete the progress bar from flow B

## Why?
Surprisingly after we ran the first A/B Test, the old design performed better than the new design. Therefore we decide to remove the progress bar from the new design, since our hypothesis is that it distracting the final user.

## Related PRs
https://github.com/guardian/identity-frontend/pull/194

### Desired
new-fow-bw 1

### Final result

screencapture

PR #1404, opened 2016-12-07
Turning on the A/B Test

svillafe:
# Turn on the checkout A/B Test

## Why?
Surprisingly after we ran the first A/B Test, the old design performed better than the new design. Therefore we decide to remove the progress bar from the new design since our hypothesis is that it distracting the final user.

## Related PRs
* New design for flow B at identity frontend: https://github.com/guardian/identity-frontend/pull/194
* New design for flow B at membership frontend: https://github.com/guardian/membership-frontend/pull/1403

### Desired
new-fow-bw 1


rtyley:
:+1:

PR #1407, opened 2016-12-12
Fix logic for unavailable upgrade scenarios

jacobwinch:
PR #1386 improved the error message that we show international users who try to upgrade to higher membership tiers (which are not available outside the UK).

However, these changes meant that we would show the same error page under two incorrect scenarios: 

1. When signed-in members followed links to join/current-tier/enter-details - this happens surprisingly often and we should probably address that as a separate issue in itself.
2. When paying members attempt to 'upgrade' to an invalid tier (I'm not convinced this has ever actually happened, but it's still possible).

This PR splits out the logic and shows an appropriate error for all cases mentioned above.

@rtyley @paulbrown1982


frontend/app/services/MemberService.scala
```diff
@@ -274,15 +274,19 @@ class MemberService(identityService: IdentityService,
...
 
-  def subscriptionUpgradableTo(sub: Subscription[SubscriptionPlan.Member], newTier: PaidTier): Boolean = {
-    import model.TierOrdering.upgradeOrdering
+  def upgradeTierIsValidForCurrency(sub: Subscription[SubscriptionPlan.Member], newTier: PaidTier): Boolean = {
     val newPlan = catalog.findPaid(newTier)
-
     // The year and month plans are guaranteed to have the same currencies
     val targetCurrencies = newPlan.year.charges.price.prices.map(_.currency).toSet
     val currencyIsAvailable = targetCurrencies.contains(sub.plan.currency)
+    currencyIsAvailable
```
johnduffell: don't need to put into a val and then return it (although it is allowed if you think that's clearer that way)

johnduffell:
👍 great work! 🚀

frontend/app/actions/ActionRefiners.scala
```diff
@@ -91,8 +91,10 @@ object ActionRefiners extends LazyLogging {
 
   def checkTierChangeTo(targetTier: PaidTier) = new ActionFilter[SubReqWithSub] {
     override protected def filter[A](request: SubReqWithSub[A]): Future[Option[Result]] = {
-      if (!request.touchpointBackend.memberService.subscriptionUpgradableTo(request.subscriber.subscription, targetTier)) {
-        Future.successful(Some(Ok(views.html.tier.upgrade.unavailable(request.subscriber.subscription.plan.tier, targetTier))))
+      if (!request.touchpointBackend.memberService.upgradeTierIsValidForCurrency(request.subscriber.subscription, targetTier)) {
+        Future.successful(Some(Ok(views.html.tier.upgrade.unavailableTierForCurrency(request.subscriber.subscription.plan.tier, targetTier))))
```
rtyley: You can extract the `Future.successful()` outside of this `if` expression to reduce repetition.

PR #1411, opened 2016-12-15
Benefits class is refactored.

svillafe:
## Why are you doing this?
Currently, the benefits are hard coded on the landing page. In this PR, we are reading those from the `Benefits.scala `

## Changes
* `Benefits.scala` **[Modified]**: Added the new benefits for the specifics landing pages. Defined the benefits object whenever is necessary.
* `elevatedSupporterUS.scala.html`, `elevatedSupporterUK.scala.html` and `elevatedSupporterAU.scala.html` **[Modified]**: Beneftis are being read from `Benefits.scala`

## Screenshots

Not applicable


rtyley:
Fewer lines, yay :+1: thanks for chasing down the legacy lines so that we could remove them.

PR #1412, opened 2016-12-15
Paypal client side work

rupertbates:
## Why are you doing this?
To enable users to pay for membership using PayPal. 

The hypothesis is that this will improve conversion on mobile where entering credit card details is laborious and time consuming.

There are still a number of things to do before this can go live so the new version of the checkout page is hidden behind a query string flag.

Work still to do:
* For tech stuff see [cards tagged with PayPal here](https://trello.com/b/TjUElW0B/simple-and-coherent-product-membership?menu=filter&filter=label:PayPal)
* Work through the finer points of workflow stuff like how to process refunds.

Depends on [this PR in membership common](https://github.com/guardian/membership-common/pull/381)

## Changes
* Adds a new version of the checkout page with a PayPal button - this will eventually be removed once we release the final version but is there for now to allow testing on PROD
* Adds a new PayPal controller to coordinate requests between the PayPal api and our js code
* Adds a new optional query string parameter to allow switching between the standard checkout page and the new PayPal one
* Added new config to membership.private.conf

## Screenshots
screen shot 2016-12-15 at 16 42 23


frontend/app/controllers/Paypal.scala
```diff
@@ -0,0 +1,105 @@
...
+		"VERSION" -> Config.paypalNVPVersion)
+
+	// Takes a series of parameters, send a request to Paypal, returns response.
+	def nvpRequest (params: Map[String, String]) = {
+
+		val client = new OkHttpClient()
+		val reqBody = new FormBody.Builder()
+
+		for ((param, value) <- defaultNVPParams) reqBody.add(param, value)
+		for ((param, value) <- params) reqBody.add(param, value)
```
paulbrown1982: This does look elegant, but it could also be written a tiny bit more succinctly:

```
defaultNVPParams.map(reqBody.add)
params.map(reqBody.add)
```

or even something like (I forget the real method name):

```
reqBody.addAll(defaultNVParams)
reqBody.addAll(params)
```

paulbrown1982: Ahh sorry, reqBody is not a Scala Map. No worries, the for loop is good then.

frontend/app/controllers/Joiner.scala
```diff
@@ -144,26 +144,39 @@ object Joiner extends Controller with ActivityTracking
...
-          countryCurrencyWhitelist,
-          identityUser,
-          pageInfo,
-          trackingPromoCode = validTrackingPromoCode,
-          promoCodeToDisplay = validDisplayablePromoCode,
-          Some(countryGroup))
-        case CheckoutFlowVariant.B => views.html.joiner.form.paymentB(
+
+      Ok(
+      if(paypalTest){
```
rupertbates: This big if statement is ugly, but it is just to support testing and will disappear when we release this stuff

frontend/app/views/fragments/form/cardDetailPayPal.scala.html
```diff
@@ -0,0 +1,189 @@
+
+@import com.gu.salesforce.PaidTier
```
rupertbates: This file duplicates cardDetail.scala.html but will be removed when we finally release PayPal

frontend/app/views/fragments/form/submitPayPal.scala.html
```diff
@@ -0,0 +1,66 @@
+@import views.support.Pricing._
```
rupertbates: This file duplicates submit.scala.html but will be removed when we finally release PayPal

Ap0c:
Tests are passing locally, the failures here are due to Paypal config not being available in the CI environment.

Ap0c:
Now there's only one test failing, and it's because this PR updates the `netaporter` library to the latest version, which breaks the test.

frontend/test/model/EBEventTest.scala
```diff
@@ -79,10 +79,6 @@ class EBEventTest extends PlaySpecification {
       updateEventWithDescription(ebLiveEvent, "<!-- main-image: https://media.gutools.co.uk/images/e5b35b1f20588172b24960314cbf8e9c8482d3bf?crop=27_0_1800_1080 ! -->")
         .mainImageUrl must beSome(Uri.parse("https://media.gutools.co.uk/images/e5b35b1f20588172b24960314cbf8e9c8482d3bf?crop=27_0_1800_1080"))
     }
-    "should not blow up even if the url is really bizarre" in {
-      updateEventWithDescription(ebLiveEvent, "<!-- main-image: http://test.net/## -->")
-        .mainImageUrl must beNone
-    }
```
Ap0c: This test was failing because the `scalaUri` library was updated to the latest version in this PR. As a result, [the code being tested](https://github.com/guardian/membership-frontend/blob/paypal-nvp-and-client-side/frontend/app/model/Eventbrite.scala#L287) no longer gives parse errors for the test url given here, and `mainImageUrl` thus returns `Some` rather than `None`.

The latest version of `Uri.parse` seems to be a lot more lenient, and I cannot find a url that will break the parser (I tried some complete garbage urls and they came back fine). As a result, I can't think of an alteration to the test that will activate the error handling code being tested. Therefore I've removed the test.

@rtyley @rupertbates please let me know if you think this is a bad idea, or if there's a better solution that I've missed 🐳.

rtyley: Rather than removing the test, just remove the assertion, ie the `.mainImageUrl must beNone` part.

The history of this test is that the event-admin staff were occasionally accidentally entering malformed urls, which would crash the whole site. All we really care about is that parsing weird urls doesn't cause an exception  - we don't care whether it evaluates as `None` or not.

Ap0c: Makes sense 🐔. Added the test back in here: e63a0d8.

PR #1418, opened 2016-12-21
stop spam to membership feedback

johnduffell:
Note: view with [files?w=0](https://github.com/guardian/membership-frontend/pull/1418/files?w=0) for better diff

## Why are you doing this?
Look at membership feedback and you'll know!!  Loads of emails we can't use in any meaningful way.

## Screenshots


rtyley:
Fair enough, though to avoid explicitly including addresses, you might consider taking a hash of the address instead, eg something like `if (hash(feedback.email.toLowerCase)) == "abc1212df...")`

johnduffell:
ooh that's a sensible idea

PR #1419, opened 2016-12-21
stop spam

johnduffell:
Why are you doing this?

Look at membership feedback and you'll know!! Loads of emails we can't use in any meaningful way.

Screenshots

### DIFF
--- a/frontend/app/services/EmailService.scala
+++ b/frontend/app/services/EmailService.scala
@@ -1,5 +1,8 @@
 package services
 
+import java.math.BigInteger
+import java.security.MessageDigest
+
 import com.amazonaws.regions.{Region, Regions}
 import com.typesafe.scalalogging.LazyLogging
 import com.gu.identity.play.IdMinimalUser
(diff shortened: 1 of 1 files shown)


frontend/app/services/EmailService.scala
```diff
@@ -14,18 +17,32 @@ trait EmailService extends LazyLogging {
...
   val client = {
-    val c= new AmazonSimpleEmailServiceClient()
+    val c = new AmazonSimpleEmailServiceClient()
     c.setRegion(Region.getRegion(Regions.EU_WEST_1))
     c
   }
 
+  private val digest = MessageDigest.getInstance("MD5")
+
+  def md5(salt: String, input: String): Option[String] = {
```
rtyley: I'd be tempted to use Guava's Hash functions here - Guava is already in Membership-Frontend, and the methods look a bit prettier than having to do `try/catch`? 

```
HashFunction hf = Hashing.md5();
HashCode hc = hf.newHasher().putString("foo@bar.com", Charsets.UTF_8).hash();
```

https://github.com/google/guava/wiki/HashingExplained

frontend/app/services/EmailService.scala
```diff
@@ -14,18 +17,32 @@ trait EmailService extends LazyLogging {
...
+      Option(new BigInteger(1, digest.digest()).toString(16))
+    } catch {
+      case _: Throwable => None
+    }
+  }
   def sendFeedback(feedback: FeedbackForm, userOpt: Option[IdMinimalUser], uaOpt: Option[String]) = {
-    logger.info(s"Sending feedback for ${feedback.name} - Identity $userOpt")
-    val to = new Destination().withToAddresses(feedbackAddress)
-    val subjectContent = new Content("Membership feedback")
+    if (md5("wxEKKfFtz4zQREhodbVvk84IzRR3iXLMCzGaIkxGv7YeLgSwEjqPPNei0e3B0kLo86b9ArS9ZgtLHzMp7qY2asot7feJYT9GQvxza95xif9mJy9kcIfncgKoPG1GloIV", feedback.email).contains("87bc0ff1050d56f06e50809a42f44bfc")) {
```
rtyley: Probably a good idea to extract this salt out to a const variable to make this snippet of CODE a bit clearer.

Personally, I think the aim with hashing here is just obfuscation rather than security, so I think you could probably dispense with the unique salt altogether?

johnduffell:
yeah ok try that!

johnduffell:
🍖

frontend/app/services/EmailService.scala
```diff
@@ -14,18 +19,27 @@ trait EmailService extends LazyLogging {
   val feedbackAddress: String
 
   val client = {
-    val c= new AmazonSimpleEmailServiceClient()
+    val c = new AmazonSimpleEmailServiceClient()
     c.setRegion(Region.getRegion(Regions.EU_WEST_1))
     c
   }
```
rtyley: You can write this as `Hashing.md5().hashString("your input", StandardCharsets.UTF_8).toString` <- note the `toString` method returns a nice hex string...

...or alternatively you could construct the blacklisted email as a `HashCode` object using `HashCode.fromString()`, and pass that around as a more strongly typed alternative to strings..!

frontend/app/services/EmailService.scala
```diff
@@ -14,18 +19,27 @@ trait EmailService extends LazyLogging {
   val feedbackAddress: String
 
   val client = {
-    val c= new AmazonSimpleEmailServiceClient()
```
rtyley: You can get rid of this now!

rtyley:
For some reason all my review comments have been shifted from where they're meant to be, but I hope you get the gist!

frontend/app/services/EmailService.scala
```diff
@@ -14,18 +19,27 @@ trait EmailService extends LazyLogging {
...
+    val c = new AmazonSimpleEmailServiceClient()
     c.setRegion(Region.getRegion(Regions.EU_WEST_1))
     c
   }
 
+  private val digest = MessageDigest.getInstance("MD5")
```
rtyley: You can get rid of this now!

frontend/app/services/EmailService.scala
```diff
@@ -14,18 +19,27 @@ trait EmailService extends LazyLogging {
...
+
+  def md5(input: String): String = {
+    val hf = Hashing.md5()
+    util.Arrays.toString(hf.newHasher().putBytes(input.getBytes("UTF-8")).hash().asBytes())
```
rtyley: You can write this as `Hashing.md5().hashString("your input", StandardCharsets.UTF_8).toString` <- note the `toString` method returns a nice hex string...

...or alternatively you could construct the blacklisted email as a `HashCode` object using `HashCode.fromString()`, and pass that around as a more strongly typed alternative to strings..!

johnduffell:
ok I've fixed the unused var, I don't plan to do anything about the tostring, can I have a +1 or -1 and I'll either ship or close!

paulbrown1982:
👍

PR #1424, opened 2017-01-05
Use latest SBT v0.13.13

rtyley:
## Why are you doing this?

This was mentioned as a massive quick-win by @rorygraves in his _Can ScalaC be 10 times faster?_ talk at Scala eXchange 2016:

https://skillsmatter.com/skillscasts/9042-can-scalac-be-10-times-faster

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=0.13.12
+sbt.version=0.13.13


johnduffell:
no +1?

rtyley:
> no +1?

It was so tiny, and I'm _so_ arrogant, that once it passed CI I didn't think it needed a +1. But we can disagree if you like :smiley:

rorygraves:
I can take no credit for this (other than putting it on a slide) - all credit goes to @retronym

PR #1426, opened 2017-01-06
Fix laggy startup

rupertbates:
## Why are you doing this?
To speed up the start up time of the site when running locally.

The delay was caused by the AddEC2InstanceHeader filter which was unable to complete when running outside of EC2 and so was causing the app to hang until it timed out.

## Changes
Only attempt to get the EC2 instance header when running in Prod

### DIFF
--- a/frontend/app/filters/AddEC2InstanceHeader.scala
+++ b/frontend/app/filters/AddEC2InstanceHeader.scala
@@ -11,18 +11,14 @@ import scala.concurrent.Future
 object AddEC2InstanceHeader extends Filter {
 
   // http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html
-  lazy val instanceIdOptF = getHeader
+  lazy val instanceIdOptF = if (Config.stageProd)
+    WS.url("http://169.254.169.254/latest/meta-data/instance-id").get().map(resp => Some(resp.body).filter(_.nonEmpty)).recover { case _ => None }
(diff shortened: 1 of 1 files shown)


rtyley:
This is absolutely bloody wonderful (plus _really_ embarrassing for me, because I wrote the `AddEC2InstanceHeader`). I measure the time for startup as roughly _11 seconds_ - down from **2m10s**: 

image


Timing test:
---

```
$ sbt compile
$ ./start-frontend.sh
```

Wait for this in the console output:

```
[info] p.c.s.NettyServer - Listening for HTTP on /0:0:0:0:0:0:0:0:9100

(Server started, use Ctrl+D to stop and go back to the console...)
```

Then hit http://localhost:9100/ to get the homepage- the time taken for that page to get delivered- after the server has started listening on port 9100 - is what I've been measuring. 3 separate test runs seen above.

@rupertbates tells me his strategy for finding this issue was to start with the assumption that it was something associated with the Play `Global` object - once there, he just started commenting things out until he got a speed change...

cc @philwills @johnduffell :champagne:

johnduffell:
haha oh dear! build a stool and grab the low hanging fruit 🎉 
I'd be interested in what other "paper cuts" harm our development speed as I wonder if any are worth spending more time on (eg having to double/triple refresh things to get things to pass, and rerunning flaky tests)

Ap0c:
This is fantastic! Nice job 💥

PR #1428, opened 2017-01-09
Paypal email to zuora

rupertbates:
## Why are you doing this?
For PayPal users we need to store the email address they use to log in to PayPal as well as the email they have registered with Identity.

This is specifically for if a payment fails when an email will be sent out (to the Identity email address) with the PayPal email address in it.

## Changes
* Refactored the PayPal controller into a controller and a service class
* Added a method in the PayPalService to retrieve the users PayPal email
* Pass in the PayPal email address when creating the users payment method in Zuora

## Screenshots
NA

### DIFF
--- a/frontend/app/controllers/PayPal.scala
+++ b/frontend/app/controllers/PayPal.scala
@@ -8,6 +8,7 @@ import com.netaporter.uri.Uri.parseQuery
 import configuration.Config
 import com.typesafe.scalalogging.LazyLogging
 import play.api.Logger
+import services.PayPalService
 
 object PayPal extends Controller with LazyLogging {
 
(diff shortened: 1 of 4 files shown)


frontend/app/services/MemberService.scala
```diff
@@ -150,12 +150,18 @@ class MemberService(identityService: IdentityService,
...
+
     formData match {
-      case paid@PaidMemberJoinForm(_, _, PaymentForm(_, _, Some(_)), _, _, _, _, _, _, _, _, _) => //Paid member with PayPal token
+      case paid@PaidMemberJoinForm(_, _, PaymentForm(_, _, Some(baid)), _, _, _, _, _, _, _, _, _) => //Paid member with PayPal token
         for {
           idUser <- getIdentityUserDetails()
           sfContact <- createSalesforceContact(idUser)
-          zuoraSubName <- createPaidZuoraSubscription(sfContact, paid, idUser.primaryEmailAddress, None)
+          email <- retrieveEmail(baid)
+          zuoraSubName <- createPaidZuoraSubscription(sfContact, paid, email, None)
```
Ap0c: Does this mean that Zuora now just stores the PayPal email address, and no longer has a copy of their Identity email address?

frontend/app/services/PayPalService.scala
```diff
@@ -0,0 +1,82 @@
...
+
+    client.newCall(request).execute()
+  }
+
+  // Takes an NVP response and retrieves a given parameter as a string.
+  private def retrieveNVPParam(response: Response, paramName: String) = {
+    val responseBody = response.body().string()
+    val queryParams = parseQuery(responseBody)
+    val result = queryParams.paramMap(paramName).head
+    response.close()
```
Ap0c: I thought `response.body().string()` had already closed the response by this point, but I may well be wrong 🐱.

As far as I can make out from the docs [here](https://square.github.io/okhttp/3.x/okhttp/okhttp3/ResponseBody.html), both the `string` and `bytes` methods implicitly close the response body.

frontend/app/controllers/PayPal.scala
```diff
@@ -18,105 +14,45 @@ object PayPal extends Controller with LazyLogging {
...
-		val agreementParams = Map(
-			"METHOD" -> "CreateBillingAgreement",
-			"TOKEN" -> token.token)
-
-		val response = nvpRequest(agreementParams)
-		Json.toJson(Token(retrieveNVPParam(response, "BILLINGAGREEMENTID")))
-
+	// Wraps the PayPal token and converts it to JSON
+  // for sending back to the client.
+	private def tokenJsonResponse (token : String) = {
```
Ap0c: I know we discussed this when we first implemented the PayPal Controller, but I find this conceptually a little bit confusing. We're using this to serialise both the `token` and the `baid`, but referring throughout to it just as the token. I'm just wondering whether we should name things slightly more generically, or possibly use a separate method for the `baid`.

Obvs this is just a readability thing, functionally it's fine. What do you think @rupertbates?

frontend/app/services/MemberService.scala
rupertbates: That is a good point, I'll tweak it to save both

frontend/app/services/PayPalService.scala
rupertbates: Yes I think you're right. I saw a warning about leaking the connection yesterday which was why I added it but I think maybe that was just with some test code and we don't need it with the final implementation.

frontend/app/controllers/PayPal.scala
rupertbates: Agreed to leave this for now after an offline discussion

Ap0c:
Teamcity build failed due to network error, code is fine. Restarted the build.

PR #1430, opened 2017-01-09
Hide Address From PayPal Checkout

Ap0c:
## Why are you doing this?

We don't want the address to appear in the PayPal checkout screen, as we're handling it separately in our checkout. This NVP parameter hides it from the user.

**Trello Card**: [Here](https://trello.com/c/ZsaDUGFo)

## Changes

- Add NVP param to hide address in PayPal checkout.

## Screenshots

**Note**: Ignore the fact that one is a window and the other is an overlay, that's just the PayPal sandbox being flaky.

**Before**:

paypal-address-before

**After**:

paypal-address-after

@rupertbates @rtyley @svillafe @JustinPinner


Ap0c:
Looks good:

paypal-address-hidden

PR #1379, opened 2016-11-21
New checkout design

svillafe:
## New Design

checkout_css

This PR implements the new design of the checkout step. If you attach the query parameter: `skin=membersB` to the URL you will see the new design.

Please @SiAdcock and @rtyley check this PR.

PR #1431, opened 2017-01-10
Make paypal checkout blue

rupertbates:
## Why are you doing this?
The PayPal checkout flow is currently using the old styling whereas the standard flow has been updated to a lovely blue colour.

This PR makes them look the same.

## Changes
* Copy all the styling changes across the standard flow to the PayPal flow

## Screenshots
Before:
screen shot 2017-01-10 at 15 27 32
After:
screen shot 2017-01-12 at 16 05 08

### DIFF
--- a/frontend/app/views/fragments/form/cardDetailPayPal.scala.html
+++ b/frontend/app/views/fragments/form/cardDetailPayPal.scala.html
@@ -34,6 +34,9 @@
                 <h3 class="credit-card-note__header">Cancellation</h3>
                 <p>You can cancel your membership online through your profile page.</p>
             </div>
+            <div class="credit-card-note">
+                <h3 class="credit-card-note__header"><a class="js-change-payment-method" href="#">Change payment method</a></h3>
+            </div>
         </div>
     </div>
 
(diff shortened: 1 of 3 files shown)


frontend/app/views/fragments/form/cardDetailPayPal.scala.html
```diff
@@ -36,6 +34,9 @@ <h3 class="credit-card-note__header">First payment</h3>
                 <h3 class="credit-card-note__header">Cancellation</h3>
                 <p>You can cancel your membership online through your profile page.</p>
             </div>
+            <div class="credit-card-note">
+                <h3 class="credit-card-note__header"><a class="js-change-payment-method" href="#">Change payment method</a></h3>
+            </div>
```
Ap0c: This doesn't show up on mobile for some reason, weird... Perhaps there's a CSS rule hiding it somewhere:

payment-method-mobile

frontend/assets/stylesheets/components/_payment-b.scss
```diff
@@ -173,6 +173,12 @@
       @include mq($from: tablet) {
         display: block;
       }
+
+      a {
+        color: guss-colour(guardian-brand);
+        display: block;
+        margin-top: 10px;
+      }
```
Ap0c: Do we still need this as a separate file? The `b` refers to the B part of the AB test that's no longer running doesn't it?

Ap0c:
The `_payment-b.scss` file was originally created for the checkout AB in #1379. Now that this is done, and the changes have been applied, is it worth merging this into `_payment.scss` and losing the references to the B flow (`.skin-checkout-b` for example)? @svillafe what do you think?

This is probably not something for this PR, but just a thought.

frontend/app/views/fragments/form/cardDetailPayPal.scala.html
rupertbates: @Ap0c Fixed now if you'd care to have another look

Ap0c:
Yup, it's there on mobile now, all good 😄

PR #1432, opened 2017-01-10
Track incomplete supporter checkouts

JustinPinner:
## Why are we doing this?
We're going to reach out to potential new supporters if they decide not to complete their journey though the payment process.

## Changes
* Make it possible for members-data-api to capture form show events for users
* Make sure the captured event is removed when the user reaches the thank you page after checkout.

## Screenshots
N/A


frontend/conf/DEV.public.conf
```diff
@@ -21,11 +21,11 @@ google.oauth {
 }
 
 membership {
-    url = "https://mem.thegulocal.com"
+    url = "https://mem.thegulocal.com:9100"
```
rupertbates: I don't think we should need these port numbers

frontend/app/controllers/Joiner.scala
```diff
@@ -289,14 +293,20 @@ object Joiner extends Controller with ActivityTracking
...
-      request.subscriber,
-      paymentSummary,
-      card,
-      destination,
-      upgrade,
-      validPromotion.filterNot(_.asTracking.isDefined)
-    )).discardingCookies(TierChangeCookies.deletionCookies: _*)
+    } yield {
+      tier match {
+        case t: Tier.Supporter if !upgrade => MembersDataAPI.Service.removeBehaviour(request)
```
JustinPinner: Yes, we're only interested in new signups at the moment.

frontend/conf/DEV.public.conf
JustinPinner: Ok, I'll remove them. This is one of the many pitfalls I encountered while trying to run this thing end-to-end locally.

svillafe: I would not introduce changes in this file.

frontend/app/controllers/Joiner.scala
```diff
@@ -289,14 +293,20 @@ object Joiner extends Controller with ActivityTracking
...
+        case _ =>
```
svillafe: Does this case have a right side?

JustinPinner: No, it's to stop the compiler emitting a warning about a non-exhaustive match if it's not there.

frontend/app/services/MembersDataAPI.scala
```diff
@@ -46,19 +47,32 @@ object MembersDataAPI {
...
     override val httpClient: LoggingHttpClient[Future] = RequestRunners.loggingRunner(MembersDataAPIMetrics)
+  }
 
+  case class BehaviourHelper(cookies: Seq[Option[Cookie]]) extends WebServiceHelper[Behaviour, ApiError] {
+    override val wsUrl: String = Config.membersDataAPIUrl
+    override def wsPreExecute(req: Request.Builder): Request.Builder = {
+      req.addHeader("Cookie", cookies.flatten.map(c => s"${c.name}=${c.value}").mkString("; "))
+    }
+    override val httpClient: LoggingHttpClient[Future] = RequestRunners.loggingRunner(MembersDataAPIMetrics)
   }
```
rtyley: Try to avoid repetition like this! `AttributeHelper` and `BehaviourHelper` are identical apart from the type parameter of the `WebServiceHelper` they extend, so it should be possible to make one class with that implementation gunk that takes that differing type as it's own type parameter, and have two one-line instances of it that differ only in that parameter.

PR #1433, opened 2017-01-11
Add salesforce auth to healthcheck

AWare:
See: https://github.com/guardian/membership-common/pull/399/

Subs was getting a weird 500 error due to Salesforce auth not being present for the first few requests on startup. Looking into it: (I think)
1) a request that requires salesforce arrives
2) scalaforce is initialised and blocks req 1 until authentication completes
3) a second request that makes a salesforce request arrives while 1 is authenticating
4) nothing is in the authentication agent and a 500 is thrown.

I've not checked for this in membership, but it looks feasible. Fixing it for subscriptions required a breaking change to membership-common; so I've fixed it here too.

This will slightly change startup behaviour in that any salesforce requests before authentication completes will fail, rather than the first request blocking.

### DIFF
--- a/frontend/app/services/SalesforceService.scala
+++ b/frontend/app/services/SalesforceService.scala
@@ -40,6 +40,8 @@ class SalesforceService(salesforceConfig: SalesforceConfig) extends api.Salesfor
   override def updateMemberStatus(user: IdMinimalUser, tier: Tier, customer: Option[Customer]): Future[ContactId] =
     upsert(user.id, memberData(tier, customer))
 
+  override def isAuthenticated = repository.salesforce.isAuthenticated
+
   private def memberData(tier: Tier, customerOpt: Option[Customer]): JsObject = Json.obj(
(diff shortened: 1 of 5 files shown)


frontend/app/services/SalesforceService.scala
```diff
@@ -40,6 +40,8 @@ class SalesforceService(salesforceConfig: SalesforceConfig) extends api.Salesfor
   override def updateMemberStatus(user: IdMinimalUser, tier: Tier, customer: Option[Customer]): Future[ContactId] =
     upsert(user.id, memberData(tier, customer))
 
+  override def isAuthenticated = repository.salesforce.authAgent.get().isDefined
```
rtyley: Probably best to shift this implementation-hiding value into the `repository.salesforce` object, which lives inside membership-common, so we're not reaching across repo boundaries.

rtyley:
:+1:

PR #1434, opened 2017-01-12
Added billingPeriod re rendering when the country is selected

svillafe:
## Why are you doing this?
There was a bug introduced in https://github.com/guardian/membership-frontend/pull/1387. The bug took place because the promo code was bound and updated that price. When we removed the promo code, that event was not being bound anymore.

Please @paulbrown1982 @rtyley review this change.

PR #1435, opened 2017-01-12
Add currency check to acceptance test

mario-galic:
# Why are you doing this?
In order to avoid the same type of bug in the future (https://github.com/guardian/membership-frontend/pull/1434), we added an extra check in the membership frontend's acceptance test.

# Changes
* Check currency can change to non-UK currency
* Switch to Supporter flow from Partner flow

### DIFF
--- a/frontend/test/acceptance/JoinPartnerSpec.scala
+++ b/frontend/test/acceptance/JoinPartnerSpec.scala
@@ -59,6 +59,12 @@ class JoinPartnerSpec extends FeatureSpec with Browser
       And("should be logged in with their Identity account.")
       assert(enterDetails.userIsSignedIn)
 
+      When("users select non-UK country,")
+      enterDetails.changeCountry()
+
+      Then("the currency should change.")
+      assert(enterDetails.currencyHasChanged)
+
       When("Users fill in delivery address details,")
       enterDetails.fillInDeliveryAddress()
 
@@ -72,7 +78,7 @@ class JoinPartnerSpec extends FeatureSpec with Browser
       assert(ThankYou.pageHasLoaded)
 
       And("should be signed in as Partner.")
-      assert(ThankYou.userIsSignedInAsPartner)
+      assert(ThankYou.userIsSignedInAsSupporter)
     }
   }
 }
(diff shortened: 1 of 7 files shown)


frontend/test/acceptance/JoinPartnerSpec.scala
```diff
@@ -72,7 +78,7 @@ class JoinPartnerSpec extends FeatureSpec with Browser
       assert(ThankYou.pageHasLoaded)
 
       And("should be signed in as Partner.")
```
rtyley: 'Partner' should be 'Supporter' now!

frontend/test/acceptance/JoinPartnerSpec.scala
```diff
@@ -59,6 +59,12 @@ class JoinPartnerSpec extends FeatureSpec with Browser
       And("should be logged in with their Identity account.")
       assert(enterDetails.userIsSignedIn)
 
+      When("users select non-UK country,")
+      enterDetails.changeCountry()
+
+      Then("the currency should change.")
```
rtyley: I guess the default currency is GBP, so the currency _will_ have changed? We don't seem to be specifying a `countryGroup` when we go to `/join/supporter/enter-details` - so I'm not sure if there's any chance the currency might already be USD, before the country is set to US.

frontend/test/acceptance/pages/EnterDetails.scala
```diff
@@ -25,6 +25,10 @@ case class EnterDetails(val testUser: TestUser) extends Page with Browser {
 
   def fillInCardDeclinedProcessError() { CreditCard.fillInCardDeclinedProcessError() }
 
+  def changeCountry() { DeliveryAddress.selectUSA() }
```
rtyley: `changeCountry()` is a bit vague here, I think you're better off making this `changeCountryTo("US")` - and then you could change country a couple of times, to checking currency each time.

frontend/test/acceptance/JoinPartnerSpec.scala
svillafe: Also the file should be called JoinSupporterSpec.scala

rtyley:
:+1:

svillafe:
It is working

svillafe:
It is working on PROD

PR #1446, opened 2017-01-23
Paypal Error Handling

Ap0c:
## Why are you doing this?

There are many places where the PayPal flow might break. It would be nice to present clear messages to users when this happens, and to do lots of logging to figure out why.

There's a fair bit of refactoring going on here, so the diff isn't that clear. Probably the easiest way to figure out what's going on is to look at the flow in the `paypal.es6` module, and trace the calls to the other files.

Also, I think I've caught the places that the flow might fail, but there may still be bugs. I'm hoping that it's set up to catch unexpected crashes, but if anyone notices additional possible sad paths please shout!

[**Trello card**](https://trello.com/c/DfTJi91x/232-client-side-error-handling-for-paypal)

## Changes

- Added error message div to the payment section.
- Added a new module (`paymentError.es6`) to handle displaying error messages to user and logging them to sentry.
- Started using bonzo in `paymentMethod.es6`.
- Refactored `paypal.es6` to handle errors better.
- Tweaked calls in `stripe.es6` to match with new payment API.
- Modified `payment.es6` to use new error module, and to provide an interface to other modules.

## Screenshots

payment-errors

@JustinPinner @rtyley @svillafe

### DIFF
--- a/frontend/app/views/joiner/form/paymentPayPal.scala.html
+++ b/frontend/app/views/joiner/form/paymentPayPal.scala.html
@@ -121,6 +121,7 @@
                                 <div class="form-field js-payment-type">                                @fragments.form.stripeCheckout(idUser)
                                 </div>
                             </div>
+                            <div class="js-payment-error is-hidden">Sorry, we weren't able to process your payment this time around. Please try again.</div>
                         </fieldset>
                         @fragments.form.cardDetailPayPal(plans)
                     </div>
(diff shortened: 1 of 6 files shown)


rupertbates:
Do you have any way to force errors for testing?

Ap0c:
It's a little difficult to check that the error handling is working ok in PROD, because I can't break bits of the backend to test it.

However, normal sign-ups on the old flow are working fine, and PayPal sign-ups are working too, so I don't think it's broken anything 🤖.

PR #1448, opened 2017-01-24
Use .nvmrc to ensure everyone uses the same version of Node

rtyley:
## Why are you doing this?

The `.nvmrc` file will help everyone ensure they are using the same version of Node, which should improve reproducability.

Note you need to execute [`$ nvm use`](https://github.com/creationix/nvm#nvmrc) in order for this to take effect.

See also https://github.com/guardian/identity-frontend/pull/202

cc @svillafe @AWare


svillafe:
👍

PR #1449, opened 2017-01-24
Pass the correct billing amount to PayPal

rupertbates:
## Why are you doing this?
So that PayPal charges users the correct amount  when they purchase membership

## Changes
* Pass billing data from the checkout page to the server in the setupPayment call
* Pass the billing data through to PayPal when retrieving the initial payment token

## Screenshots

screen shot 2017-01-24 at 17 00 14
*PayPal now shows the correct price of a US membership, not hard coded values*

### DIFF
--- a/frontend/app/controllers/PayPal.scala
+++ b/frontend/app/controllers/PayPal.scala
@@ -1,8 +1,8 @@
 package controllers
 
 import com.typesafe.scalalogging.LazyLogging
-import play.api.libs.json.{JsError, JsSuccess, Json}
-import play.api.mvc.Controller
+import play.api.libs.json.{JsError, JsSuccess, Json, Reads}
+import play.api.mvc.{AnyContent, Controller, Request}
 import services.PayPalService
 
 object PayPal extends Controller with LazyLogging {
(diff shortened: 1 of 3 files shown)


frontend/app/controllers/PayPal.scala
```diff
@@ -22,22 +25,25 @@ object PayPal extends Controller with LazyLogging {
...
-				case e: JsError => BadRequest(JsError.toJson(e).toString)
-			}
+  //Takes a request, parses it into a type T, passes this into serviceCall to retrieve a token then returns this as json
+  def parseJsonAndRunServiceCall[T](request:Request[AnyContent])(serviceCall : T => String)(implicit fjs: Reads[T]) = {
+    request.body.asJson.map { json =>
 
-		}.getOrElse(BadRequest)
+      Json.fromJson[T](json)(fjs) match {
+        case JsSuccess(parsed, _) => Ok(tokenJsonResponse(serviceCall(parsed)))
+        case e: JsError => BadRequest(JsError.toJson(e).toString)
```
JustinPinner: Is exposing the json just for testing purposes?

JustinPinner: (assuming the trace will expose at least _some_ of the json!)

rupertbates: I'm not sure actually, that was already there. I'll check if it's used

frontend/assets/javascripts/src/modules/form/paypal.es6
```diff
@@ -32,14 +32,21 @@ function setupPayment (resolve, reject) {
 
 	if (payment.validateForm()) {
 
+        const checkoutForm = guardian.membership.checkoutForm;
+        const amount = checkoutForm.billingPeriods[checkoutForm.billingPeriod].amount[checkoutForm.currency];
```
JustinPinner: Is checkoutForm.currency defaulted to something for international (ROW) users? I seem to remember the country code being '--'. (I'm assuming this comes through from the payment details form)

JustinPinner:
I think this looks fine. Might want a second opinion, too though. 👍

frontend/app/controllers/PayPal.scala
rupertbates: Yes I think it's just for debugging but as discussed it is only showing parsing errors from the request body which has come from our page and so is public anyway.
I'm going to leave it as is because it can be useful for debugging during development

frontend/assets/javascripts/src/modules/form/paypal.es6
rupertbates: This is true, the user needs to select a country before they can attempt payment or they get a regular form validation error

PR #1457, opened 2017-01-25
Fix Cloudwatch Thread Spawning

Ap0c:
# Why are you doing this?

There is currently a problem where there are far too many threads being spawned to handle the CloudWatch client, which is causing membership-frontend to crash regularly when running in dev (out of memory). This bumps membership-common to pick up the fix, which is described in more detail in [the PR in that repository](https://github.com/guardian/membership-common/pull/413).

[**Trello Card**](https://trello.com/c/pTJCTynM/292-fix-aws-client-thread-splurge)

# Changes
- Bump membership common to pick up the fix to the `CloudWatch` trait.
- Removed implementation of region field, as the functionality for this has been moved to a companion object.

@svillafe @JustinPinner @rupertbates @paulbrown1982 @mario-galic

### DIFF
--- a/frontend/app/monitoring/Metrics.scala
+++ b/frontend/app/monitoring/Metrics.scala
@@ -5,7 +5,6 @@ import com.gu.monitoring.CloudWatch
 import configuration.Config
 
 trait Metrics extends CloudWatch {
-  val region = Region.getRegion(Regions.EU_WEST_1)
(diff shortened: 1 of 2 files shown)


Ap0c:
It's a little tricky to test the effects of these changes in PROD, but the supporter flow works so hopefully we haven't broken anything!

PR #1458, opened 2017-01-26
Pass item details to paypal

rupertbates:
## Why are you doing this?
To provide more information in the PayPal checkout UI so that the user can tell exactly what they are buying.

## Changes
* Pass the tier to which the user is subscribing into setupPayment call along with the billing details
* Pass line item details through to PayPal in the call to SetExpressCheckout

## Screenshots

screen shot 2017-01-26 at 13 44 15

### DIFF
--- a/frontend/app/services/PayPalService.scala
+++ b/frontend/app/services/PayPalService.scala
@@ -36,6 +36,7 @@ object PayPalService extends LazyLogging {
   // Takes an NVP response and retrieves a given parameter as a string.
   private def retrieveNVPParam(response: Response, paramName: String) = {
     val responseBody = response.body().string()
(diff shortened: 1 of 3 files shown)


frontend/app/services/PayPalService.scala
```diff
@@ -36,6 +36,9 @@ object PayPalService extends LazyLogging {
   // Takes an NVP response and retrieves a given parameter as a string.
   private def retrieveNVPParam(response: Response, paramName: String) = {
     val responseBody = response.body().string()
+    if (Config.stageDev)
+      logger.debug("NVP response body = " + responseBody)
```
Ap0c: Nice! This will be really handy for development ⭐️.

PR #1466, opened 2017-01-27
IE bug fixes

svillafe:
## Why are you doing this?
We are fixing two issues that were not compatible in IE.
* NodeList has no implementation of forEach.
* One of the svg had extra parameter and interfered with the rendering of it in IE.

## Changes
* Newspaper's svg
* bundleLanding.es6

### Before

picture 88

### After

picture 87


frontend/assets/javascripts/src/modules/landingBundles.es6
```diff
@@ -42,8 +42,11 @@ function bindButtonBehaviour() {
 }
 
 function bindOptionsBehaviour() {
-    PRINT_OPTIONS.forEach(function(field) {
-        field.addEventListener('click', function(evt){
+
+    for(var i = 0 ; i < PRINT_OPTIONS.length ; i++) {
+        var field = PRINT_OPTIONS[i];
```
Ap0c: Alternatively:

```
[].forEach.call(PRINT_OPTIONS, field => {
```

PR #1468, opened 2017-01-27
A little copy change

svillafe:
## Why are you doing this?
New text from Nick, in order to be clearer to the final user.


rtyley:
:+1:

svillafe:
It is working on **PROD**.

PR #1472, opened 2017-01-30
Pricing match

svillafe:
## Why are you doing this?
We want to eliminate a compiler warning about an Inexhaustive Match in Pricing.scala,

## Changes
* Added a base case in `Pricing.scala`

## Screenshots

N/A


svillafe:
It is working on **PROD**

PR #1474, opened 2017-01-31
Fix TestUser sensitivity to clock drift

rtyley:
## Why are you doing this?

This is a re-do of https://github.com/guardian/subscriptions-frontend/pull/784 for Membership Frontend.

Acceptance tests on Subscriptions Frontend were [intermittently failing](https://github.com/guardian/subscriptions-frontend/pull/781#issuecomment-276336674) - this was happening when the TestUser token was stamped with a time ahead of the clock time on the `subscriptions-frontend` appserver:

    "IllegalArgumentException: The end instant must be greater the start"

The same fix applies to Membership Frontend (though we don't seem to have the same errors occurring here, probably because we're not using Guest Checkout)

## Changes

See https://github.com/guardian/identity-test-users/commit/410a7fec for the details of the fix, and accompanying tests that check for tolerance of clock drift.

cc @Ap0c

### DIFF
--- a/frontend/app/utils/TestUsers.scala
+++ b/frontend/app/utils/TestUsers.scala
@@ -1,6 +1,7 @@
 package utils
 
-import com.github.nscala_time.time.Imports._
+import java.time.Duration.ofDays
+
 import com.gu.identity.play.IdMinimalUser
 import com.gu.identity.testing.usernames.TestUsernames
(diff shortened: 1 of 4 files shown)


rtyley:
This PR was aiming to improve post-deploy testing stability, so this pass is a Good Thing :-)

image

PR #1475, opened 2017-01-31
Add Fetch Polyfill

Ap0c:
## Why are you doing this?

We're using `fetch` in the PayPal code for our AJAX requests, but it's not supported in all browsers. This adds GitHub's polyfill to handle that.

Note that we could have used `reqwest` here, because we're using it elsewhere under the name `ajax`. However, this library is now outdated, and `fetch` is on the way to being [implemented natively](https://fetch.spec.whatwg.org) in [all browsers](http://caniuse.com/#feat=fetch), so we think using it is preferable. This is probably going to be a precursor to replacing `reqwest` with `fetch` throughout the codebase.

[**Trello Card**](https://trello.com/c/8ZonVpeN/303-fetch-doesn-t-work-in-old-browsers)

## Changes
- Add fetch polyfill to dependencies.
- Import fetch in PayPal module.

@rupertbates @rtyley @JustinPinner

### DIFF
--- a/frontend/assets/javascripts/src/modules/form/paypal.es6
+++ b/frontend/assets/javascripts/src/modules/form/paypal.es6
@@ -2,6 +2,7 @@
 
 import form from 'src/modules/form/helper/formUtil';
 import * as payment from 'src/modules/payment';
+import 'whatwg-fetch';
 
 
 // ----- Functions ----- //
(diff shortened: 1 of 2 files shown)


JustinPinner:
If that's all it takes! 👍

Ap0c:
@svillafe it looks to me like both of those issues are related to limitations of `XMLHttpRequest`, which means that they would also apply to `reqwest`, since it also uses this. In other words, I don't think we would lose functionality relative to that library by using the fetch polyfill.

Ap0c:
Tested on the lovely IE 10, looks ok:

ie-fetch

PR #1478, opened 2017-02-01
Support non-signed-in Test Users

rtyley:
## Why are you doing this?

Soon we're going to A/B test a flow where the user may not be logged in when they visit the checkout page. Unfortunately the [Test User](https://membership.theguardian.com/test-users) functionality has - up till now - **required the user to be signed in** to work out whether they're a Test User or not -and we need to know whether the user is a Test User or not to choose the correct Touchpoint backend.

Now we need to know if a user is Test User, without them being signed in.

## Changes

This change matches the work in https://github.com/guardian/subscriptions-frontend/pull/165, adding to Membership Frontend new methods for us to tell if a non-signed in user is a Test User. Only the original method is valid for all pages on the site:

- `PreSigninTestCookie` <- valid for viewing correct rateplans pre-purchase. You can get this cookie simply by going to https://membership.theguardian.com/test-users - the cookie will be dropped automatically
- `NameEnteredInForm` <- valid for posting the Checkout form - you will need to enter the Test User token as your First Name
- `SignedInUsername` <- valid for all pages

If the user is signed in, only `SignedInUsername` is used and other credentials are irrelevant.

Because this is all hideously complicated, **this change also adds an informative banner on Checkout and Thank you pages**, that confirms you've been verified as a valid test-user, explaining what credential has been used to work that out: _"Using test backend: UAT because you are signed in as a user with a valid Test Username"_

Note that - at the moment - you still have to be signed-in to view the Checkout and Thank You pages, so the `PreSigninTestCookie` & `NameEnteredInForm` methods can not be exercised.

This change requires https://github.com/guardian/membership-common/pull/417 to be released before this PR will build and pass tests.

## Screenshots

https://mem.thegulocal.com/join/supporter/enter-details
image

https://mem.thegulocal.com/join/supporter/thankyou
image

cc @Ap0c

### DIFF
--- a/frontend/app/controllers/Joiner.scala
+++ b/frontend/app/controllers/Joiner.scala
@@ -29,6 +29,7 @@ import services.PromoSessionService.codeFromSession
 import services.{GuardianContentService, _}
 import tracking.ActivityTracking
 import utils.RequestCountry._
+import utils.TestUsers.{NameEnteredInForm, PreSigninTestCookie}
 import utils.{CampaignCode, TierChangeCookies}
 import views.support
 import views.support.MembershipCompat._
@@ -117,7 +118,14 @@ object Joiner extends Controller with ActivityTracking
     pricingType: Option[BillingPeriod],
     paypalTest: Boolean = false) = NonMemberAction(tier).async { implicit request =>
 
-    implicit val backendProvider: BackendProvider = request
+    implicit val resolution: TouchpointBackend.Resolution =
+      TouchpointBackend.forRequest(PreSigninTestCookie, request.cookies)
+
+    implicit val tpBackend = resolution.backend
+
+    implicit val backendProvider: BackendProvider = new BackendProvider {
+      override def touchpointBackend = tpBackend
+    }
     implicit val c = catalog
 
     val identityRequest = IdentityRequest(request)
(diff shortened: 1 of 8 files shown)


frontend/app/controllers/Joiner.scala
```diff
@@ -272,6 +282,7 @@ object Joiner extends Controller with ActivityTracking
   }
 
   def thankyou(tier: Tier, upgrade: Boolean = false) = SubscriptionAction.async { implicit request =>
+    implicit val resolution: TouchpointBackend.Resolution = TouchpointBackend.forRequest(PreSigninTestCookie, request.cookies)
```
Ap0c: This is identical to [this line](https://github.com/guardian/membership-frontend/pull/1478/files#diff-071e948f4be1f1aa847ecc97e7a8110bR121). Could be pulled out into a function if it ends up being used on another page as well 🐔.

rtyley: Fair enough - there will eventually be 3 endpoints that use this method, is it ok if I look for commonality or a refactor when they're all in place?

rtyley:
Banner working on PROD:

image

rtyley:
Production real-live Supporter signups have also continued after deploy!

https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#metricsV2:graph=~(view~'timeSeries~stacked~false~metrics~(~(~'membership~'createMember~'Services~'Member~'Stage~'PROD~(period~300~stat~'Sum))~(~'.~'sign-ups-Supporter~'.~'.~'.~'.~(period~300~stat~'Sum))~(~'.~'failed-sign-up-Supporter~'.~'.~'.~'.~(period~300~stat~'Sum)))~region~'eu-west-1~start~'-PT12H~end~'P0D);search=PROD;namespace=membership;dimensions=Services,Stage

PR #1479, opened 2017-02-02
Merge Patron Benefits Into Single List

Ap0c:
## Why are you doing this?

On the checkout page, the benefits for patron are split into two separate lists. The books and tickets benefits are listed separately from the rest, as they are for partners. However, while this is necessary for partners, because you have to choose between them with a radio button, it makes no sense for patrons, because you get both.

[**Trello Card**](https://trello.com/c/w07SkHr1/316-merge-patron-benefits-into-single-list)

## Changes

- Modify the benefits model, to split tickets and books and provide fuller description.
- Remove the second list of patron benefits, the tickets and books, from the checkout page.

## Screenshots

**Before:**

patron-before

**After:**

patron-after

**Partner Unchanged:**

partner-unchanged

@svillafe @rupertbates @JustinPinner


Ap0c:
It works!

patron-benefits

PR #1480, opened 2017-02-02
Paypal config

rupertbates:
## Why are you doing this?
To enable separate PayPal configurations for different environments and to use the DEV (sandbox) environment for test users in production.

## Dependencies
[This PR](https://github.com/guardian/membership-common/pull/418) in membership common
Requires an update to membership.private.conf files on S3

## Changes
* Add different config blocks for DEV and PROD and read from these according to the current environment
* Use DEV config for test users even in production

## Screenshots
NA


frontend/app/services/PayPalService.scala
```diff
@@ -1,32 +1,39 @@
...
+    "SIGNATURE" -> requestConfig.signature,
+    "VERSION" -> requestConfig.NVPVersion)
 
   // Takes a series of parameters, send a request to PayPal, returns response.
-  private def nvpRequest(params: Map[String, String]) = {
+  private def nvpRequest(params: Map[String, String], user : IdMinimalUser) = {
+    logger.info("Is test user = " + TestUsers.isTestUser(user))
+    val requestConfig = TouchpointBackend.forUser(user).payPalConfig
 
     val client = new OkHttpClient()
```
rtyley: This looks like it'll create a new `OkHttpClient` with each request?

From the docs: 
https://github.com/square/okhttp/blob/71729c39f2/okhttp/src/main/java/okhttp3/OkHttpClient.java#L57

> OkHttpClients should be shared
> 
> OkHttp performs best when you create a single OkHttpClient instance and reuse it for all of your HTTP calls. This is because each client holds its own connection pool and thread pools. Reusing connections and threads reduces latency and saves memory. Conversely, creating a client for each request wastes resources on idle pools.

...I bet we break this rule elsewhere, but we definitely want to cut back here I think!

rupertbates: Well spotted!

tomverran:
FYI there is/was an OkHttpClient used by various Zuora services in an object called something like RequestRunners in membership common, you use it through some functions defined alongside it.

...I still get GitHub notifications about membership 😄

rupertbates:
Nice, thanks Tom

frontend/app/services/PayPalService.scala
```diff
@@ -1,36 +1,43 @@
...
 import com.typesafe.scalalogging.LazyLogging
 import configuration.Config
 import controllers.PayPal.{PayPalBillingDetails, Token}
 import controllers.routes
 import okhttp3.{FormBody, OkHttpClient, Request, Response}
-import play.api.mvc.RequestHeader
+import play.api.mvc.AnyContent
+import utils.TestUsers
 
 object PayPalService extends LazyLogging {
```
rtyley: To fit in with the other services, `PayPalService` shouldn't be a singleton `object` - instead it should be a class like `StripeService`,`ZuoraService`, etc, taking it's backend-specific config in it's constructor. This lets you make it the PayPalService into  a field on 'MemberService' - so that the 'MemberService` uses only services that are configured for the same backend.

frontend/app/services/PayPalService.scala
```diff
@@ -1,36 +1,43 @@
...
+    "USER" -> requestConfig.user,
+    "PWD" -> requestConfig.password,
+    "SIGNATURE" -> requestConfig.signature,
+    "VERSION" -> requestConfig.NVPVersion)
 
   // Takes a series of parameters, send a request to PayPal, returns response.
-  private def nvpRequest(params: Map[String, String]) = {
+  private def nvpRequest(params: Map[String, String], user : IdMinimalUser) = {
+    logger.info("Is test user = " + TestUsers.isTestUser(user))
+    val requestConfig = TouchpointBackend.forUser(user).payPalConfig
```
rtyley: You won't need to hand down the `user`, or derive the `TouchpointBackend.forUser(user).payPalConfig`, if the paypal config is already passed to this class in it's constructor!

PR #1481, opened 2017-02-03
update membership common to use rateplan frontendid

johnduffell:
see https://github.com/guardian/membership-common/pull/419

This is needed so we can tell 6 for 6 guardian weekly apart from normal yearly guardian weekly.  It also makes the catalog process more deterministic and thus less flaky (and easier to debug when it doesn't work)

@guardian/membership-and-subscriptions we will add 6 for 6 in dev and uat next week, so make sure you merge in master when this is shipped otherwise you will see a "too many plans" error when reading the catalog.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -12,7 +12,7 @@ object Dependencies {
     "com.gu" %% "memsub-common-play-auth" % "0.8", // v0.8 is the latest version published for Play 2.4...
     "com.gu" %% "identity-test-users" % "0.6"
   )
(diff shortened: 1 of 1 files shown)


johnduffell:
@paulbrown1982 @rtyley can I get a +1 on this please? 🌟

paulbrown1982:
👍 from me. @rtyley ?

johnduffell:
PS there was another membership common pr going into this (to stop the deadlock ) https://github.com/guardian/membership-common/pull/421

PR #1482, opened 2017-02-06
Turn off price test

svillafe:
## Why are you doing this?
The A/B Test is finished and we need to deactivate this test along with its logic. 

## Changes
* Remove the full logic of the AB Test.


svillafe:
It is working on **PROD**.

PR #1483, opened 2017-02-06
Setup A1 vs A2 test & remove B variants.

JustinPinner:
## Why are you doing this?
Review in conjunction with frontend changes here: https://github.com/guardian/frontend/pull/15767

We've concluded from the results of the A1/B1/B2 test that the A variant wins. Now we're feature-testing between A variants, where the control is ad-free, and the variant is with-ads.

## Changes
* Set up for two variants of A: CONTROL (ad-free) and VARIANT (with-ads)
* Remove B variant pages, css etc.

## Screenshots
**Control**
screen shot 2017-02-06 at 16 05 10

**Variant**
screen shot 2017-02-06 at 16 05 51


JustinPinner:
Good catch, thanks @Ap0c!

Panic over;
screen shot 2017-02-06 at 18 15 58

frontend/app/abtests/BundleVariant.scala
```diff
@@ -49,8 +48,8 @@ object BundleTier {
 
 
 object Distribution {
-  val A = Distribution("A")
-  val B = Distribution("B")
+  val CONTROL = Distribution("CONTROL")
```
svillafe: In order to reuse the existing name, I would keep A1 and A2 as test's name.

frontend/app/abtests/BundleVariant.scala
```diff
@@ -49,8 +48,8 @@ object BundleTier {
 
 
 object Distribution {
```
svillafe: The distribution object was thought to store the existing designs. Before there were A and B, in this PR should exist only A and indicate the different version of A using the `priceindex`.

svillafe: Instead of creating a new design for A2, we should use the hasAdFree variable to show whether the variant has that feature or not.

JustinPinner: Conceptually, using the priceindex to differentiate whether a feature is visible or not is super confusing, especially as the prices are the same on both variants. I know what you're saying, and could call priceindex something else, but that means more changes than are necessary to achieve the results we're after.

frontend/app/abtests/BundleVariant.scala
JustinPinner: The test no longer refers to A1 or A2. Keeping those identifiers here adds confusion. Better to make the code on both sides consistent so someone unfamiliar has a chance to understand what's going on. Because we have to have a `control` group on the frontend side for Abacus, we're going to see it here too.

frontend/app/abtests/BundleVariant.scala
```diff
@@ -9,18 +9,17 @@ object BundleVariant {
...
 
   def lookup(name: String): Option[BundleVariant] = {
     all.find(_.testId == name)
   }
 }
 
-case class BundleVariant(distribution: Distribution, priceIndex: Int, prices: Map[BundleTier, Double], hasAdFree: Boolean = true) {
-  val testId =  s"MEMBERSHIP_AB_THRASHER_UK_${distribution.name}$priceIndex"
+case class BundleVariant(distribution: Distribution, prices: Map[BundleTier, Double], hasAdFree: Boolean = true) {
+  val testId =  s"MEMBERSHIP_A_THRASHER_UK_${distribution.name}"
```
rtyley: We might want to include commenting as part of this test identifier - eg "MEMBERSHIP_A_COMMENT_THRASHER_UK_..." - in case we do any other kinds of tests with the 'A' variant. Otherwise it gets a bit more fiddly to distinguish test results from different tests.

JustinPinner: Fair point. This is the ad-free test, so  I'll add `_ADS_`.

JustinPinner:
🙄 Ok, I'll deal with it via resolve conflicts.

PR #1484, opened 2017-02-07
Adding disallow statements for the landing pages.

svillafe:
## Why are you doing this?

We would like to prevent search engines to crawl our site. Therefore, we add a line in the `robots.txt` file.

## Changes
* Modified the `robots.txt`

### DIFF
--- a/frontend/public/robots.txt
+++ b/frontend/public/robots.txt
@@ -4,7 +4,6 @@ Disallow: /patterns
 Disallow: /whats-on
 Disallow: /join-challenger
 Disallow: /bundles
-Disallow: /bundles/thankyou
 
 Allow: /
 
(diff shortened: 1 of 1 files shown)


frontend/public/robots.txt
```diff
@@ -3,6 +3,8 @@ User-agent: *
 Disallow: /patterns
 Disallow: /whats-on
 Disallow: /join-challenger
+Disallow: /bundles
+Disallow: /bundles/thankyou
```
rtyley: In principal, I think just the first line (`Disallow: /bundles`) is suffient:

```
The "Disallow" field specifies a partial URI that is not to be visited. This can be a full path,
or a partial path; any URI that starts with this value will not be retrieved. For example,

    Disallow: /help disallows both /help.html and /help/index.html, whereas
    Disallow: /help/ would disallow /help/index.html but allow /help.html. 
```

https://www.w3.org/TR/html4/appendix/notes.html#h-B.4.1.1

svillafe: Good Spot 👍

svillafe:
It is working on **PROD**

picture 118

PR #1437, opened 2017-01-17
Landing pages for bundles

svillafe:
## Why are you doing this?
This is the first version of the design of the variant A. We want to do an A/B Test, between design A and design B. In addition in this A/B Test we are adding the logic for the A/B Test. That logic can be found inside the file BundleVariant.scala. In this first version we are not adding the buttons and the icons for each offering.

## Changes
* Added logic for the A/B/C Test
* Added HTML and CSS for the two landing possibilities.

## Desired design
optiona

## First approach
optionaimplemented

PR #1485, opened 2017-02-07
Pay pal ab test

rupertbates:
## Why are you doing this?
When we release PayPal as a payment method we want to AB test it against a checkout flow with Stripe only to ensure that it gives us a better conversion rate. This is important as there is a belief that PayPal will increase the rate of churn (would be nice to confirm or disprove this with data) so if it does not increase conversion then it may be actually hurting overall profitability.

### Things to note:
* This is only set up for the supporter flow from /uk/supporter at the moment
* It is currently disabled so that we can merge it without putting the PayPal checkout live. To enable it uncomment the commented code in the [PayPalTestVariants](https://github.com/guardian/membership-frontend/pull/1485/files#diff-cd6309b149c63816062c9229212e4796) class:

`def allocateVariant() = None //Some(variants(Random.nextInt(2))) //uncomment to go live` 

## Changes
* Allocate two UI variants 'with_paypal' and 'stripe_only' to represent the alternate checkout flows
* Change the paypalTest query param into an option of a String representing the UI variant to use.
* Track the variant using the 'experience' dimension in GA

### With PayPal variant
screen shot 2017-02-07 at 12 30 04

### Stripe only variant
screen shot 2017-02-07 at 12 29 35


frontend/assets/javascripts/src/modules/tools/payPalABTest.es6
```diff
@@ -0,0 +1,27 @@
...
+    guardian.payPalTestVariant = getQueryParameterByName(paramName);
+    if (guardian.payPalTestVariant) {
+        cookie.setCookie(paramName, guardian.payPalTestVariant, 3);
+    } else {
+        guardian.payPalTestVariant = cookie.getCookie(paramName);
+    }
+
+}
+
+function isThankYouPage () {
```
rtyley: Probably can drop this, as there's no harm to setting the `experience` in GA on other pages.

frontend/app/abtests/PayPalTestVariants.scala
```diff
@@ -0,0 +1,10 @@
+package abtests
+
+import scala.util.Random
+
+object PayPalTestVariants {
+  val withPayPal = "with_paypal"
+  val stripeOnly = "stripe_only"
+  val variants = List(withPayPal, stripeOnly)
```
rtyley: It can be good to avoid stringly-typed stuff like this with an enum-style class, like [`BundleVariant`](https://github.com/guardian/membership-frontend/blob/37008909/frontend/app/abtests/BundleVariant.scala#L5-L26).

In this case it's only a bit of nicety tho', as unlike `BundleVariant`, there isn't a lot of extra pricing and behaviour associated, so not essential for this PR. Might be a nice excuse to try out [lloydmeta/enumeratum](https://github.com/lloydmeta/enumeratum) tho'!

svillafe: If you decide to go with the approach of avoiding stringly-typed stuff, you will have to modify `Binders.scala`. See this PR with the same type of changes: https://github.com/guardian/membership-frontend/pull/1437/files#diff-108bb3217f691dfd764ade3ec6ff3196L1

PR #1486, opened 2017-02-07
Deleted some old cache settings. Using Play's aggressively strategy.

svillafe:
## Why are you doing this?

The changes introduced in https://github.com/guardian/membership-old/pull/318, didn't use the play's cache strategy correctly. In the `application.conf` file, it was set as a **default cache time** the same amount of time that play use int the aggressive caching strategy. Therefore, we modified the `routes` file to indicate which route should be cache aggressively and which one should be cache for the _default time_.

## Changes
* Change routes file.
* Change `application.conf`

### /robots.txt

| Before | After |
|--------|-------|
|picture 120 | picture 119 |

### /assets/javascripts/main.js

| Before | After |
|--------|-------|
| picture 123 | picture 122 |

### /uk/supporter

| Before | After |
|--------|-------|
|picture 124 |  picture 125 |

### DIFF
--- a/frontend/conf/routes
+++ b/frontend/conf/routes
@@ -2,8 +2,8 @@
 # This file defines all application routes (Higher priority routes first)
 # ~~~~
 
-GET            /robots.txt                            controllers.CachedAssets.at(path="/public", file="robots.txt", false:Boolean)
-GET            /humans.txt                            controllers.CachedAssets.at(path="/public", file="humans.txt", false:Boolean)
+GET            /robots.txt                            controllers.CachedAssets.at(path="/public", file="robots.txt", aggressiveCaching:Boolean ?= false )
+GET            /humans.txt                            controllers.CachedAssets.at(path="/public", file="humans.txt", aggressiveCaching:Boolean ?= false )
 GET            /favicon.ico                           controllers.CacheBustedAssets.at(path="images/favicons/32x32.ico")
 GET            /sitemap.xml                           controllers.SiteMap.sitemap()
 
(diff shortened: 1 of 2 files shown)


frontend/conf/routes
```diff
@@ -2,8 +2,8 @@
 # This file defines all application routes (Higher priority routes first)
 # ~~~~
 
-GET            /robots.txt                            controllers.CachedAssets.at(path="/public", file="robots.txt")
-GET            /humans.txt                            controllers.CachedAssets.at(path="/public", file="humans.txt")
+GET            /robots.txt                            controllers.CachedAssets.at(path="/public", file="robots.txt", aggressiveCaching:Boolean ?= false )
```
rtyley: Is the question mark necessary?

svillafe: Without the question mark, Play assumes that value will be provided in the left part of the route. Since we don't provide it, it fails with a 500. 

Related resources:

https://www.playframework.com/documentation/2.0.2/ScalaRouting
http://stackoverflow.com/questions/13633089/pass-parameter-value-in-routes

rtyley: It also says:

```
Parameters with fixed values

# Extract the page parameter from the path, or fix the value for /
GET   /                     controllers.Application.show(page = "home")
```

...this just doesn't seem to be working with passing boolean literal true?!
https://www.playframework.com/documentation/2.0.2/ScalaRouting#Parameters-with-fixed-values

svillafe: I tried that approach and unfortunately, it is not working either passing `nameOfVariable = true` or `nameOfVariable = "true"`.

svillafe:
It is working in **PROD**. The CDN's URL has been purged as well.

PR #1487, opened 2017-02-07
Track incomplete checkouts take 2

JustinPinner:
## Why are you doing this?
Initially, the plan was to pass our cookies over to the members data API and let that decode and re-validate them to determine the id we wanted to record against the activity. 

Despite everything working as expected when running locally, deployment to PROD revealed a bunch of error messages returned as a result of the API not being able to complete that cookie revalidation.

So, now we're wrapping the data up in the POST to the API, which records that data directly instead of attempting the revalidation.

Again, this works locally, so we'll see how it goes when it hits PROD.

This should be merged after https://github.com/guardian/members-data-api/pull/155 is deployed and ready.

## Screenshots
See comment below for updated log captures etc.


frontend/app/controllers/Joiner.scala
```diff
@@ -132,6 +132,10 @@ object Joiner extends Controller with ActivityTracking
     (for {
       identityUser <- identityService.getIdentityUserView(request.user, identityRequest)
     } yield {
+      tier match {
+        case t: Tier.Supporter => MembersDataAPI.Service.addBehaviour(request, "enterPaidDetails.show")
+        case _ =>
```
svillafe: Is this a good Scala style? It seems weird to have an empty right side in a case of a pattern match.

JustinPinner:
Ok, here's a little update now that I've updated the API's cloudformation and tested it in PROD from my local membership.thegulocal;

# TL;DR - it works!

On checkout form display;
**PROD members-data-api log**
```
2017-02-09 15:50:37,510 [application-akka.actor.default] controllers.BehaviourController[BehaviourController.scala:41] INFO: recorded enterPaidDetails.show on 2017-02-09T15:50:36.941Z for 2******9
```

**THEGULOCAL membership-frontend log**
```
2017-02-09 15:50:37,551 [application-akka.actor.default] application[MembersDataAPI.scala:97] INFO: Recorded enterPaidDetails.show for 2******9
```

**Dynamo**
screen shot 2017-02-09 at 15 51 53

On payment completed (thank you page);
**PROD members-data-api log**
```
2017-02-09 16:00:15,751 [application-akka.actor.default] controllers.BehaviourController[BehaviourController.scala:49] INFO: removed enterPaidDetails.show for 2******9
```

**THEGULOCAL membership-frontend log**
```
2017-02-09 16:00:15,765 [application-akka.actor.default] application[MembersDataAPI.scala:107] INFO: Cleared behaviours for 2******9
```

**Dynamo**
screen shot 2017-02-09 at 16 01 28

frontend/app/controllers/Joiner.scala
JustinPinner: It's an acceptable thing to do: http://stackoverflow.com/questions/25843677/what-does-it-mean-to-have-an-empty-case-statement-in-scala Admittedly we're not returning anything here, but we're saying explicitly that we've considered what to do with a tier that isn't of the type we're looking for.

PR #1489, opened 2017-02-08
Deleting bean from comparisonTable.es6

svillafe:
## Why are you doing this?
We want to delete old js libraries, and this is the first step to do that. In this first PR we delete bean.js from `comparisonTable.es6`. The focus of this PR is to replace the library and not improve the performance of the functions.

In order to minimise the impact of the change, I have just replaced the behaviour of bean.js with built-in functions. In bean.js when a selector is passed to the function .on, it goes trough all the sub-elements of the root that fulfil that selector and attach an element with addEventListener. It is not the most performant thing to do but it was necessary in order to reuse the code of the handler and try to minimise the number of potential new bugs. 

## Changes
* Mofidied `comparisonTable.es6`

## Screenshots

N/A

### DIFF
--- a/frontend/assets/javascripts/src/modules/comparisonTable.es6
+++ b/frontend/assets/javascripts/src/modules/comparisonTable.es6
@@ -1,18 +1,17 @@
-import * as bean from 'bean'
 import $ from '$'
 
 'use strict';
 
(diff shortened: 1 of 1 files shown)


svillafe:
It is working on **PROD** on **IE10**

PR #1491, opened 2017-02-08
Removed the local events from the index page.

svillafe:
## Why are you doing this?
Local events are not being supported anymore by membership. As a result the index page of membership was broken. This PR fix the style of the page.

## Changes
* Change index.html
* Change grid.scss

### Before

picture 126

### After

picture 127


svillafe:
It is working on **PROD**

PR #1496, opened 2017-02-10
Checkout Accordion Design for PayPal

Ap0c:
## Why are you doing this?

The rollout of PayPal (and the new Stripe checkout) require a redesign of the checkout page, because the PayPal button was not playing nice with form validation. This changes the page to work more like the subscriptions checkout, so the payment section cannot be opened until the previous section has been validated (see screenshots below).

Also, I'm really, really sorry about the size of this PR, I'm aware that it's massive. I will try to explain all the bits to make it clearer what's going on. Conceptually not much has changed, it's mostly just structure and styling.

[**Trello Card**](https://trello.com/c/s1UepOq4/234-fix-paypal-popup-on-failed-form-validation)

## Changes

All of the changes revolve around the main PayPal checkout template: `paymentPayPal.scala.html`, which I have restructured.

### Duplicate Scala Templates

The template fragments that have been *created* are just duplicates of existing checkout fragments, I have mostly given them the same filename with the word `PayPal` appended to the end. This is done because I needed to make changes, and couldn't touch the existing files easily without breaking the current live checkout. However, when we go live with this new version the duplicates can go away.

I did manage to drop a template we weren't using any more, `submitPayPal.scala.html`, as there is no submit button in this new version of the checkout.

In summary:

- Modified `paymentPayPal.scala.html`
- Added `benefitsFieldsetPayPal.scala.html`
- Added `billingAddressPayPal.scala.html`
- Added `featureChoicePanelPayPal.scala.html`
- Added `nameDetailPayPal.scala.html`
- Added `paymentOptionsPayPal.scala.html`
- Removed `submitPayPal.scala.html`
- Added `termsPayPal.scala.html`
- Added `billingPeriodChoicePayPal.scala.html`

### Styling

I have added a single SASS file, `_payment-paypal.scss` and put all of the style code in there (and imported this in the root `style.scss`). These changes act as a skin, similar to `_payment-b.scss`, and rely on the `skin-checkout-paypal` class on the `main` element in `paymentPayPal.scala.html`. This is horrible, and will need to go away when we rebuild the checkout page. I have also added a new `lock.svg` icon that Ben provided me with.

In summary:

- Added `lock.svg`
- Added `_payment-paypal.scss`
- Modified `style.scss`

### JavaScript

[...]

PR #1500, opened 2017-02-14
Checkout Accordion Design Redux

Ap0c:
## Why are you doing this?

This is a re-deploy of #1496, as the previous deploy broke due to changes made in #1490, so was reverted in #1499. The `BillingPeriod`s were capitalised in membership-common, but were not in this branch in membership-frontend. This has now been fixed.


Ap0c:
Normal (old checkout) sign-ups are working ok, as are PayPal ones:

PR #1502, opened 2017-02-15
Simplify and remove many deprecated calls to the PromotionService

rtyley:
## Why are you doing this?

Getting rid of some deprecation warnings is nice (this leaves us with 4, rather than 8, deprecation-warnings related to `PromotionService`), but the main thing is deletion of some ugly lines of code (_44 additions and 69 deletions_) :smiley: 

I'm about to do some work in this area and wanted to tidy up before I started:

## Changes

We now pass around a `Option[ValidPromotion]` rather than a pair of `Option[PromoCode]`, reducing the number of arguments that need to be passed through, and getting rid of some fairly repeated code of the form `.filter(_.asTracking.isDefined).filter(_.asTracking.isDefined).flatMap(p => providedPromoCode)`

This will also require an update from membership-common: https://github.com/guardian/membership-common/pull/427

cc @Ap0c @AWare

### DIFF
--- a/frontend/app/controllers/Healthcheck.scala
+++ b/frontend/app/controllers/Healthcheck.scala
@@ -23,7 +23,7 @@ object Healthcheck extends Controller {
     new BoolTest("Events", () => GuardianLiveEventService.events.nonEmpty),
     new BoolTest("CloudWatch", () => CloudWatchHealth.hasPushedMetricSuccessfully),
     new BoolTest("ZuoraPing", () => zuoraSoapClient.lastPingTimeWithin(2.minutes)),
-    new BoolTest("Promotions", () => promos.all.nonEmpty),
+    new BoolTest("Promotions", () => promos.futureAll.value.exists(_.map(_.nonEmpty).getOrElse(false))),
     new BoolTest("Salesforce", () => salesforceService.isAuthenticated)
   )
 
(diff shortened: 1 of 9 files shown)


frontend/app/controllers/Healthcheck.scala
```diff
@@ -23,7 +23,7 @@ object Healthcheck extends Controller {
     new BoolTest("Events", () => GuardianLiveEventService.events.nonEmpty),
     new BoolTest("CloudWatch", () => CloudWatchHealth.hasPushedMetricSuccessfully),
     new BoolTest("ZuoraPing", () => zuoraSoapClient.lastPingTimeWithin(2.minutes)),
-    new BoolTest("Promotions", () => promos.all.nonEmpty),
+    new BoolTest("Promotions", () => promos.futureAll.value.exists(_.map(_.nonEmpty).getOrElse(false))),
```
rtyley: This may have caused the outage last night? Maybe `promos.futureAll` isn't suitable for the healthcheck?

jacobwinch: It looks we had a similar issue when using promos.futureAll for the subs healthcheck and had to revert the change:
https://github.com/guardian/subscriptions-frontend/pull/758/commits/5b5af1195f62af9fe41faecdca84c79bd9c007fa

PR #1504, opened 2017-02-15
Fix members-ticket-at-no-extra-cost usage-calculation

rtyley:
## Why are you doing this?

Partners & Patrons can get 6 members-tickets-at-no-extra-cost a year for Guardian Events. We use [Zuora Entitlements](https://knowledgecenter.zuora.com/CB_Billing/J_Billing_Operations/Usage) to keep track of how many tickets have been consumed this way, to ensure they don't go over their allowance.

A report passed on to us from Alice James in Guardian Events made us realise that we were incorrectly counting the number of tickets consumed: we should have been been counting from the start of the subscription *term* (the annual renewal of the subscription), rather than from the start of the subscription itself. This error meant that users who had been with us for more than a year were having their ticket consumption from the previous year counted against their allowance, incorrectly stopping them from using their allowance for the current year.

Slightly strangely, it looks like there has already been a PR to fix this issue: https://github.com/guardian/membership-frontend/pull/978 ...however, although the PR description implies the fix has been made, there's no sign of it in [in the diff](https://github.com/guardian/membership-frontend/pull/978/files#r101306237).

## Change
Use the _term start date_ rather than _subscription start date_ as the time from which to count allowance consumption. This should be fine so long as we only use annual terms...?

cc @paulbrown1982

### DIFF
--- a/frontend/app/services/MemberService.scala
+++ b/frontend/app/services/MemberService.scala
@@ -335,8 +335,12 @@ class MemberService(identityService: IdentityService,
       case plan: PaidSubscriptionPlan[_, _] => plan.features
       case _ => List.empty
     }
-    val startDate = subscription.startDate.toDateTimeAtStartOfDay(DateTimeZone.forID("America/Los_Angeles"))
+    val startDate = subscription.termStartDate.toDateTimeAtStartOfDay(DateTimeZone.forID("America/Los_Angeles"))
+
     zuoraService.getUsages(subscription.name, unitOfMeasure, startDate).map { usages =>
+      logger.info(s"getUsageCountWithinTerm: User ${subscription.accountId} has used ${usages.size} tickets since $startDate " +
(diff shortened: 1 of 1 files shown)


jacobwinch:
As far as I can tell we only use annual terms for Membership, so this change looks good to me.

👍 

(@paulbrown1982 is out this week)

rtyley:
As a brand-new test-user Partner, I can see the appropriate _"You can use one of your 6 allocated member tickets for this event."_ message banner on the event page:

https://membership.theguardian.com/event/brainwaves-beautiful-equations-31541125355
image

Clicking 'Book ticket' gets me the appropriate _"Member's ticket at no extra cost"_ ticket class unlocked:

image

I don't have any more-than-a-year-Partner credentials handy though, so we'll just have to see if complaints from existing Partners are reduced.

PR #1505, opened 2017-02-15
Don't chase abandoned carts if inappropriate

JustinPinner:
*See also: https://github.com/guardian/members-data-api/pull/162*

## Why are you doing this?
Not all carts are abandoned equally, so there are times when we can send out reminder emails and times we can't, or shouldn't. Currently, we only know when the checkout page was presented to a potential supporter, and that they didn't complete the process.

This PR adds simple notes to the behaviour table so that we can tell why a cart was abandoned and make the kind of decisions we need to about if / when to send out a reminder to complete the process.

## Screenshots
screen shot 2017-02-15 at 16 32 29

### DIFF
--- a/frontend/app/controllers/Joiner.scala
+++ b/frontend/app/controllers/Joiner.scala
@@ -125,7 +125,7 @@ object Joiner extends Controller with ActivityTracking
       identityUser <- identityService.getIdentityUserView(request.user, identityRequest)
     } yield {
       tier match {
-        case t: Tier.Supporter => MembersDataAPI.Service.addBehaviour(request, "enterPaidDetails.show")
+        case t: Tier.Supporter => MembersDataAPI.Service.upsertBehaviour(request, "enterPaidDetails.show", t.name)
         case _ =>
       }
       val plans = catalog.findPaid(tier)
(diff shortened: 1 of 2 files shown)


frontend/app/controllers/Joiner.scala
```diff
@@ -259,14 +259,17 @@ object Joiner extends Controller with ActivityTracking
...
           Forbidden(Json.toJson(error))
 
         case error: PaymentGatewayError =>
+          setBehaviourNote(tier.name, error.code)(request)
           handlePaymentGatewayError(error, request.user.id, tier.name, idRequest.trackingParameters, formData.deliveryAddress.countryName)
 
         case error =>
           salesforceService.metrics.putFailSignUp(tier)
           logger.error(s"User ${request.user.id} could not become ${tier.name} member: ${idRequest.trackingParameters}", error)
+          setBehaviourNote(tier.name, "card_error")(request)
```
Ap0c: This could also catch unknown PayPal errors when PayPal goes live. The name `card_error` might make people assume there's a problem with Stripe, so this may either need to be more generic, or we may need to split out the errors more when the time comes. It'll probably be trial and error catching specific PayPal errors because the docs are limited here... 😞

JustinPinner: You're right, but I don't know exactly what'll come out of here yet. The only intentional errors I've been able to create using the Stripe test card numbers generate a typical stack trace error message. I considered parsing the reason code out of that (we don't want to write it all to Dynamo or it'll cost us 💷 💷 💷) but I'd like to compile some real examples first. For now, this'll give me a prompt to dig into the logs to find out what really happened, and what we could potentially make use of.

PR #1509, opened 2017-02-16
[AppNexus fix] Setting a default case for Staff tier.

svillafe:
## Why are you doing this?
We didn't consider a default case for the thank you page of the Staff tier.

## Changes
* Converted the map to a match expression so the compiler gives a warning if we forgot to match a tier.
* Modified all the views in order to use the new methods.

## Screenshots

N/A

### DIFF
--- a/frontend/app/tracking/AppnexusPixel.scala
+++ b/frontend/app/tracking/AppnexusPixel.scala
@@ -8,9 +8,9 @@ import com.gu.salesforce.Tier.Friend
 
 object AppnexusPixel {
 
-  val thankYouPageIDs:Map[Tier, Int] = Map(Partner() -> 793015, Patron() -> 793021, Supporter() ->793016, Friend() -> 793017)
+  val thankYouPageIds:Map[Tier, Int] = Map(Partner() -> 793015, Patron() -> 793021, Supporter() ->793016, Friend() -> 793017)
 
-  val landingPageIDs:Map[Tier, Int] = Map(Supporter() -> 7269289)
+  val landingPageIds:Map[Tier, Int] = Map(Supporter() -> 7269289)
 
-  val checkoutPageIDs:Map[Tier, Int] = Map(Partner() -> 7269292, Patron() -> 7269296, Supporter() -> 7269291)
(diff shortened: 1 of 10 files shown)


frontend/app/tracking/AppnexusPixel.scala
```diff
@@ -8,9 +8,29 @@ import com.gu.salesforce.Tier.Friend
...
 
-  val landingPageids:Map[Tier, Int] = Map(Supporter() -> 7269289)
+  def getLandingPageID(tier: Tier) : Option[Int] = {
+    tier match {
+      case Supporter() => Some(7269289)
+      case _ => None
+    }
+  }
 
-  val checkoutPageids:Map[Tier, Int] = Map(Partner() -> 7269292, Patron() -> 7269296, Supporter() -> 7269291)
```
rtyley: I think you can keep all of `thankYouPageids`, `landingPageids` and `checkoutPageids` as Maps (which is more concise! Also, capitalise the the id as `Id` in the field names), but when you access them, use [`.get()`](http://www.scala-lang.org/api/2.12.1/scala/collection/Map.html#get(key:K):Option[V]) or similar, rather than invoking the `apply()` method, which does not return an option.

johnduffell: 😆

johnduffell:
yeah I'm happy either way - as said in my opinion it's a matter of style which one makes more sense, just better not to rely on the callers to remember to call .get so if you want to go that way just bake the `.get` into the function.

frontend/app/tracking/AppnexusPixel.scala
```diff
@@ -8,9 +8,9 @@ import com.gu.salesforce.Tier.Friend
 
 object AppnexusPixel {
 
-  val thankYouPageids:Map[Tier, Int] = Map(Partner() -> 793015, Patron() -> 793021, Supporter() ->793016, Friend() -> 793017)
+  val thankYouPageIDs:Map[Tier, Int] = Map(Partner() -> 793015, Patron() -> 793021, Supporter() ->793016, Friend() -> 793017)
```
rtyley: 'D' lower case!

PR #1511, opened 2017-02-16
Release New Stripe and Acceptance Test It

Ap0c:
## Why are you doing this?

We'd like to get the new checkout out with the new Stripe overlay ASAP, without waiting for PayPal. This means replacing the current checkout with the newly designed one, and simply hiding PayPal behind the AB-test flag introduced in #1485.

In order to do this, we also have to update the acceptance tests to a) work with the new checkout flow, and b) work with the new Stripe overlay. Much of the Selenium code needed to handle the overlay was taken from subscriptions-frontend, specifically the checkout [spec](https://github.com/guardian/subscriptions-frontend/blob/master/test/acceptance/CheckoutSpec.scala) and [helper](https://github.com/guardian/subscriptions-frontend/blob/master/test/acceptance/pages/Checkout.scala) files.

**Note**: A lot of duplicate fragments were created for the new checkout, which we hoped to merge back into the old fragments with this release. Unfortunately, a lot of these old fragments are still being used by the upgrade checkouts, so they have to stay where they are for the moment. Next job is to modernise the upgrade checkouts so we can merge down these fragments too.

[**Trello Card**](https://trello.com/c/a14R9BqI/339-let-s-get-new-stripe-out)

## Changes

- Stopped switching between the two versions of the checkout in `Joiner.scala`, there is now one checkout template and PayPal is accessed using an AB-test query param.
- Merged the old checkout template with the new one, picking up changes from both (`payment.scala.html` and `paymentPayPal.scala.html`).
- Renamed `addressDetailFields.scala.html`, `benefitsPanel.scala.html`, `billingAddressFields.scala.html`, `featureChoicePanel.scala.html`, `nameDetailFields.scala.html`, `paymentOptionsFields.scala.html` and `billingPeriodChoiceFields.scala.html` to indicate that they are no longer PayPal specific.
- Merged the two `terms.scala.html` fragments into one.
- Updated the Selenium tests in `JoinSupporterSpec.scala`, and the `EnterDetails.scala` and `Browser.scala` helper files.

## Screenshots

new-checkout-name-address

new-checkout-payment

new-checkout-stripe

More available on #1496.

### Sign up as
|           | Chrome | Firefox | Safari | IE10 |
|-----------|--------|---------|--------|------|
| Supporter |   ✅      |   ✅       |     ✅         | ✅              |
| Partner   |    ✅     |     ✅     |    ✅          |     ✅          |
| Patron    |    ✅        |     ✅        |      ✅        |     ✅          |


frontend/app/views/joiner/form/payment.scala.html
```diff
@@ -1,66 +1,75 @@
...
-    touchpointBackendResolution: services.TouchpointBackend.Resolution
+    touchpointBackendResolution: services.TouchpointBackend.Resolution,
+    paypalTestVariant: Option[String]
     )(implicit token: play.filters.csrf.CSRF.Token, request: actions.AuthRequest[_])
 
-@main(plans.tier.cta, pageInfo=pageInfo, simpleHeader=true, touchpointBackendResolutionOpt = Some(touchpointBackendResolution)) {
+@main(plans.tier.cta, pageInfo=pageInfo, simpleHeader=true, margins=false, touchpointBackendResolutionOpt = Some(touchpointBackendResolution)) {
 
-    <main class="page-content skin-checkout-b">
+    <main class="page-content skin-checkout-b skin-checkout-paypal">
```
Ap0c: Both these skins should really get merged down into the form sass files. However, these form files are still being used on the upgrade checkouts. Since we're planning to update these checkouts to the new design, this would probably best be done at the same time.

frontend/test/acceptance/pages/EnterDetails.scala
```diff
@@ -7,37 +7,41 @@ import org.scalatest.selenium.Page
...
+  def changeCountry(country: String) = { DeliveryAddress.selectCountryCode(country) }
 
-  def fillInCardDeclinedCvc() { CreditCard.fillInCardDeclinedCvc() }
+  def currencyHasChanged(): Boolean = hiddenElementHasText(currency, "US$")
 
-  def fillInCardDeclinedExpired() { CreditCard.fillInCardDeclinedExpired() }
+  def pay() = clickOn(className("js-stripe-checkout"))
 
-  def fillInCardDeclinedProcessError() { CreditCard.fillInCardDeclinedProcessError() }
+  def switchToStripe() = driver.switchTo().frame(driver.findElement(StripeCheckout.container.by))
```
mario-galic: Please do not use `driver.findElement` directly. Create another public method in `acceptance.util.Browser` trait for switching frames and use that:

```
  def switchFrame(q: Query) {
    if (pageHasElement(q))
      driver.switchTo().frame(q.webElement)
    else
      throw new MissingPageElementException(q)
  }
```

frontend/test/acceptance/JoinSupporterSpec.scala
```diff
@@ -68,12 +68,33 @@ class JoinSupporterSpec extends FeatureSpec with Browser
...
+      assert(enterDetails.stripeCheckoutHasExph())
+
+      Then("CVC field is present")
+      assert(enterDetails.stripeCheckoutHasCVC())
+
+      Then("submit button is present")
+      assert(enterDetails.stripeCheckoutHasSubmit())
+
+      And("they fill in credit card payment details,")
+      enterDetails.fillInCreditCardPaymentDetailsStripe()
```
mario-galic: This is doing two things: filling in details and clicking a button. I would separate out these two actions in respective methods.

frontend/test/acceptance/pages/EnterDetails.scala
```diff
@@ -47,20 +51,37 @@ case class EnterDetails(val testUser: TestUser) extends Page with Browser {
...
+    val cardCvc = xpath("//div[label/text() = \"CVC\"]/input")
+    val submitButton = xpath("//div[button]")
+
+    private def fillInHelper(cardNum: String) = {
+
+      setValueSlowly(cardNumber, cardNum)
+      setValueSlowly(cardExp, "1019")
+      setValueSlowly(cardCvc, "111")
+      continue()
+      Thread.sleep(5000)
```
mario-galic: I wield no hammer so this is just an opinion: If you need to use `Thread.sleep` to make the test work, you should be rethinking the test (but I know this is a hack taken from subs-fe)

frontend/test/acceptance/JoinSupporterSpec.scala
```diff
@@ -68,12 +68,33 @@ class JoinSupporterSpec extends FeatureSpec with Browser
...
       And("click 'Pay' button,")
       enterDetails.pay()
 
+      Then("the Stripe Checkout iframe should display")
+      assert(enterDetails.stripeCheckoutHasLoaded())
+
+      When("the checkout iframe is present")
+      enterDetails.switchToStripe()
+
+      Then("credit card field is present")
```
mario-galic: Write `Then, And, And, ...` instead of `Then, Then, Then, ...`

frontend/test/acceptance/JoinSupporterSpec.scala
```diff
@@ -68,12 +68,33 @@ class JoinSupporterSpec extends FeatureSpec with Browser
...
+      assert(enterDetails.stripeCheckoutHasCC())
+
+      Then("expiry date field is present")
+      assert(enterDetails.stripeCheckoutHasExph())
+
+      Then("CVC field is present")
+      assert(enterDetails.stripeCheckoutHasCVC())
+
+      Then("submit button is present")
+      assert(enterDetails.stripeCheckoutHasSubmit())
```
mario-galic: I guess you are asserting on the existence of CC, Exp, CVC and Submit button, so that you can safely go ahead and fill in the form. You should design your test such that `stripeCheckoutHasLoaded()` already gives you this confidence, so you do not need to make multiple extra assertions. If the submit button is there it is very likely strip checkout has loaded.

PR #1512, opened 2017-02-17
Change Remote Browser to Chrome on Windows 8.1.

Ap0c:
## Why are you doing this?

The acceptance tests don't work because they were designed against the Stripe DOM as it was observed in Chrome. However, they run remotely against Internet Explorer, for which Stripe appears to load a different DOM. Thus we're changing the remote browser to Chrome.

## Changes

- Update the driver to use Chrome.


Ap0c:
Well, the acceptance test is now passing ^, so I think that worked 🙂

PR #1515, opened 2017-02-20
Better Paypal Return And Cancel Errors

Ap0c:
## Why are you doing this?

The return and cancel URLs are never expected to be hit in production, so we would like to log any contact with them as errors. However, that doesn't mean the error pages can't be at least somewhat user-friendly...

[**Trello Card**](https://trello.com/c/QCR6Kgeh/239-monitor-return-and-cancel-urls-possibly-do-full-implementation)

## Changes

- Upgraded return URL to an error.
- Modified the PayPal error page to make it more user-friendly, with links back to the different tiers:
    + Supporter goes to Supporter landing page.
    + Partner goes to the Partner checkout page (because we don't have a landing page).
    + Patron goes to the Patron landing page.

## Screenshots

Yes, it's ugly, but we're expecting this page never to be reached...

paypal-new-errors

### DIFF
--- a/frontend/app/controllers/PayPal.scala
+++ b/frontend/app/controllers/PayPal.scala
@@ -51,7 +51,7 @@ object PayPal extends Controller with LazyLogging with PayPalServiceProvider {
   // redirected and needs to come back.
   def returnUrl = NoCacheAction {
 
-    logger.info("User hit the PayPal returnUrl.")
+    logger.error("User hit the PayPal returnUrl.")
     Ok(views.html.paypal.errorPage())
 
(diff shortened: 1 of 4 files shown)


rupertbates:
I think this is fine. If we find that we are getting users landing on these pages then we should obviously invest the time to improve them but until then 👍

Ap0c:
Just ran through the standard supporter flow and nothing seems to be broken. I don't want to hit the actual URLs, because I'd prefer if Sentry sent us a notification the first time they show up.

PR #1517, opened 2017-02-22
Paypal Acceptance Tests

Ap0c:
## Why are you doing this?

It would be good if we could have acceptance tests for PayPal ready before releasing it, to make sure we don't break it by accident.

These changes also involve adding the PayPal sandbox credentials to the membership config. In order to run the acceptance tests locally you will have to pull down the new DEV private config (instructions are in the **Download private keys** section of the membership-frontend [README](https://github.com/guardian/membership-frontend/blob/master/README.md)).

I've also used this PR as an opportunity to make some refactors suggested by @mario-galic in #1511.

[**Trello Card**](https://trello.com/c/WjyHHjYB/302-paypal-monitoring-testing)

## Changes

- Added a new scenario to the `JoinSupporterSpec` for PayPal.
- Modified `EnterDetails` page to include new methods for PayPal, including a new `PayPalCheckout` object.
- Pulled the `setValueSlowly` method out into the Browser module, as it's needed by both Stripe and PayPal.
- Added new config for the membership PayPal sandbox creds.
- Modified the `setValue` method to add a 'clear' option, because autofill was messing with the registration fields.
- Added methods for switching between windows and frames, and changing the URL, to the `Browser` module.

### DIFF
--- a/frontend/test/acceptance/JoinSupporterSpec.scala
+++ b/frontend/test/acceptance/JoinSupporterSpec.scala
@@ -38,7 +38,7 @@ class JoinSupporterSpec extends FeatureSpec with Browser
 
       Given("users click 'Become a Supporter' button on Membership homepage")
 
-      When("they land on 'Identity Frontend' page,")
+      When("They land on 'Identity Frontend' page,")
       val register = pages.Register(testUser)
       go.to(register)
       assert(register.pageHasLoaded)
@@ -60,7 +60,7 @@ class JoinSupporterSpec extends FeatureSpec with Browser
       And("should be logged in with their Identity account.")
       assert(enterDetails.userIsSignedIn)
 
-      When("users select USA delivery country,")
+      When("Users select USA delivery country,")
       enterDetails.changeCountry("US")
 
       Then("the currency should change.")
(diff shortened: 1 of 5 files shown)


svillafe:
Tested in my local environment and it ran correctly. In the scenario of having a chrome window open while the acceptance test is running, the second time I execute the test it failed.

frontend/test/acceptance/JoinSupporterSpec.scala
```diff
@@ -30,14 +30,15 @@ class JoinSupporterSpec extends FeatureSpec with Browser
   }
 
   feature("Become a Supporter in UK") {
+
     scenario("User joins as Supporter by clicking 'Become a Supporter' button on Membership homepage", Acceptance) {
```
svillafe: I would rename this scenario as: "User joins as Supporter using Stripe" so it is consistent to the new paypal scenario.

Ap0c: Good call, 66506b8.

frontend/test/acceptance/JoinSupporterSpec.scala
```diff
@@ -72,35 +73,115 @@ class JoinSupporterSpec extends FeatureSpec with Browser
...
+      }
+
+      And("should be logged in with their Identity account.")
+      assert(enterDetails.userIsSignedIn)
+
+      When("Users select USA delivery country,")
+      enterDetails.changeCountry("US")
+
+      Then("the currency should change.")
+      assert(enterDetails.currencyHasChanged)
```
JustinPinner: Is it possible to check it has changed in the right way? (We may do already - I haven't read that function yet)

JustinPinner: See - it does check for US$ - ignore me.

Ap0c: [We do](https://github.com/guardian/membership-frontend/pull/1517/files#diff-7f40b1cfbd6356bb7297b4ca0f47f200R20) 🙂

JustinPinner:
Nice! Ship it! 👍 🚢

frontend/test/acceptance/util/Browser.scala
```diff
@@ -55,6 +77,32 @@ trait Browser extends WebBrowser {
       throw new MissingPageElementException(q)
   }
 
+  // Switches to a new iframe specified by the Query, q.
+  def switchFrame(q: Query) {
```
mario-galic: Good!

frontend/test/acceptance/util/Browser.scala
```diff
@@ -55,6 +77,32 @@ trait Browser extends WebBrowser {
...
+      driver.switchTo().frame(q.webElement)
+    else
+      throw new MissingPageElementException(q)
+  }
+
+  /*
+   * Switches to the first window in the list of windows that doesn't match
+   * the parent window.
+   * */
+  def switchWindow() {
```
mario-galic: Do `switchWindow`, `switchToParentWindow`, and `changeUrl` fit better in `acceptance.util.Driver` instead of `Browser`?

frontend/test/acceptance/util/Browser.scala
```diff
@@ -34,13 +37,32 @@ trait Browser extends WebBrowser {
...
   }
 
+  /*
+    * Stripe wants you to pause between month and year and between each quartet in the cc
+    * This causes pain when you use Selenium. There are a few stack overflow posts- but nothing really useful.
+    * This pausing also seems to be necessary to make PayPal work properly,
+    * without this it starts to complain about invalid credentials after only
+    * a few characters.
+    * */
+  def setValueSlowly(q: Query, value: String): Unit = {
```
mario-galic: Interesting that the issue exists for both PayPal and Stripe. @Ap0c could you ask PayPal support why is this necessary, while @jacobwinch does the same for Stripe.

PR #1519, opened 2017-02-23
Switch on PayPal!

rupertbates:
## Why are you doing this?
To enable PayPal

### Notes 
* Currently only enabled via an A/B test query parameter so will only appear in supporter flow
* Whether a user sees the PayPal payment option is tracked in GA via the 'experience' custom dimension. The payment method they finally choose is tracked via the 'paymentMethod' custom dimension.


Ap0c:
These acceptance tests for this have failed twice, the first time because Zuora timed out 😒, the second because PayPal appeared to reject the sandbox login credentials... They were filled in fine by Selenium, PayPal just didn't accept them.

I'm going to run the tests again, and we'll see what happens.

PR #1520, opened 2017-02-23
Update Acceptance Test Config For PayPal Sandbox

Ap0c:
## Why are you doing this?

So the acceptance tests failed on #1517 because Saucelabs doesn't have access to the private config where they are stored. The solution to this is to make them available as Travis environment variables for the Saucelabs tests, and fall back to the private config when testing locally.

## Changes

- Updated acceptance test config to use optional environment variables, and fall back to those present in the private config (which it imports).


Ap0c:
These changes have worked, as demonstrated by the lovely green tick from prout shown above 🙂.

PR #1522, opened 2017-02-23
Update to AWS SDK v1.11.95

rtyley:
## Why are you doing this?
In a separate PR, I was getting a `ClassNotFoundException` until I did this AWS SDK update.

```
Caused by: java.lang.ClassNotFoundException: com.amazonaws.transform.JsonErrorUnmarshallerV2
```

## Changes

Requires updates from https://github.com/guardian/membership-common/pull/433

Note that `services.S3PrivateKeyService` was replaced by [`com.gu.memsub.auth.common.S3PrivateKeyService`](https://github.com/guardian/memsub-common-play-auth/blob/ff74974f/src/main/scala/com/gu/memsub/auth/common/S3PrivateKeyService.scala) some time ago and should have been deleted then!

cc @svillafe @AWare


rtyley:
I can still get to https://membership.theguardian.com/test-users (which uses AWS S3 to get a secret key), custom CloudWatch metrics are still being reported, and post-deploy passed, so I'd say this change looks fine.

PR #1524, opened 2017-02-24
Improve Logging For Payment Methods

Ap0c:
## Why are you doing this?

There are two main aims for this PR:

1. Start logging the payment method using CloudWatch metrics, so that we can improve monitoring of the platform health (and allow @rtyley to make pretty graphs).
2. Start logging the success or failure of NVP requests to PayPal, so we have a better idea of why payments fail.

[**Trello Card**](https://trello.com/c/nRzemYIN/360-record-stripe-vs-paypal-sign-ups-in-aws)

## Changes

- Added new method for recording payment methods in `MemberMetrics`, recording the method itself as a dimension.
- Called this new metric logging method once sign-up is complete in `MemberService`.
- Restructured the handling of NVP response from PayPal in `PayPalService`. Instead of passing the `Response` object around, we now decode it immediately and passed the decoded `QueryString` around instead. This means that we can extract the status of the response for logging straight away, and then extract further parameters further down the line.

### DIFF
--- a/frontend/app/monitoring/MemberMetrics.scala
+++ b/frontend/app/monitoring/MemberMetrics.scala
@@ -1,6 +1,8 @@
 package monitoring
 
 import com.gu.salesforce.Tier
+import com.gu.zuora.soap.models.Commands.PaymentMethod
+import com.amazonaws.services.cloudwatch.model.Dimension
 
 class MemberMetrics(val backendEnv: String) extends TouchpointBackendMetrics {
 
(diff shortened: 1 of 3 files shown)


frontend/app/monitoring/MemberMetrics.scala
```diff
@@ -26,6 +28,20 @@ class MemberMetrics(val backendEnv: String) extends TouchpointBackendMetrics {
     put(s"failed-sign-up-${tier.name}")
   }
 
+  def putCreationOfPaidSubscription(paymentMethod: Option[PaymentMethod]) = {
+
+    val methodName = paymentMethod match {
+      case Some(method) => method.getClass.getSimpleName
+      case None => "UnknownPaymentMethod"
```
rtyley: Sigh. `paymentMethod` will probably never be `None` for a paid subscription - but will be `None` when a Friend signs up. `createPaymentMethod()` should possibly return just a `PaymentMethod`, not an `Option[PaymentMethod]`, as it doesn't seem to be used in the Free flow?

https://github.com/guardian/membership-frontend/blob/13cf75ae3e7/frontend/app/services/MemberService.scala#L459

Anyway, not a blocker for this change, but a bit sad...

Ap0c: Updated to ignore the `None` case 279687e.

Ap0c:
Just ran through the flow, everything looks like it's working ok. Will monitor the logs to see that stuff is appearing as expected.

Ap0c:
The metrics appear to be tallying ok with the Supporter sign-ups, and successful NVP requests are appearing in the logs.

PR #1528, opened 2017-02-27
Set State/Province label correctly for US/Canadian users

rupertbates:
## Why are you doing this?
Currently US and Canadian users are seeing the label 'County (optional)' for their State or Province select box. This is incorrect both because the terminology is wrong and also because this field is not optional for them.

This PR fixes both of these issues

### Before Canada:
screen shot 2017-02-27 at 11 47 19
### Before US:
screen shot 2017-02-27 at 11 47 02

### After Canada:
screen shot 2017-02-27 at 11 48 04
### After US:
screen shot 2017-02-27 at 11 47 53


JustinPinner:
Looks sensible 👍

PR #1534, opened 2017-03-06
Add the 'Backend' dimension to all Touchpoint metrics.

rupertbates:
## Why are you doing this?
Currently we only add the Backend dimension to CloudWatch metrics if the value is different to the Stage eg. Test users use the UAT backend in Prod.
It would be useful for the purposes of filtering events to record the Backend whatever the value, this change does that.


JustinPinner:
LGTM 👍 
Out of interest, why did we limit the capture of these events? Is it because of the 💷💷💷s it costs to capture them?

rupertbates:
Not really sure, we asked Roberto who wrote the code and he said that it was something to do with Librato not supporting dimensions at the time (he couldn't really remember), but it does now hence the change.

PR #1537, opened 2017-03-06
Added polyfill for Promise.

svillafe:
## Why are you doing this?
Webpack 2 needs a polyfill for Promise in order to be able to run in IE11.

## Screenshots

After adding the polyfill:

picture 198


svillafe:
This configuration, does not make sense with Wepack in the current status.

PR #1539, opened 2017-03-07
Allow unauthenticated users to view checkout (not for general public!)

rtyley:
## Why are you doing this?

We're testing whether Merging Registration into the Supporter checkout will improve conversion.

Note that it is still not possible for an unauthenticated user to complete checkout - this just lets them get as far as the checkout page: https://membership.theguardian.com/join/supporter/enter-details - further steps **will fail**.

## Changes

Regular users will still authenticate, but it's possible to enable the unauthenticated behaviour by opting-in to the 'MergedRegistration feature (I've added a very small amount of scaffolding for opt-in
features):

https://membership.theguardian.com/feature/MergedRegistration
https://membership.theguardian.com/feature/MergedRegistration/Off

The code that respects this is in `OptionallyAuthenticatedNonMemberAction`
and `authenticatedIfNotMergingRegistration()`.

#### with https://mem.thegulocal.com/feature/MergedRegistration/On : 

image

#### with https://mem.thegulocal.com/feature/MergedRegistration/Off : 

...user is forced to sign-in, and then:

image

cc @Ap0c  @dominickendrick

### DIFF
--- a/frontend/app/views/fragments/form/nameDetailFields.scala.html
+++ b/frontend/app/views/fragments/form/nameDetailFields.scala.html
@@ -1,9 +1,4 @@
-@(firstName: Option[String], lastName: Option[String])(implicit request: RequestHeader)
-@import utils.TestUsers.testUsers
-
-@if(firstName.exists(testUsers.isValid)) {
-    <!-- VALID TEST USER -->
-}
+@(privateFields: Option[com.gu.identity.play.PrivateFields])(implicit request: RequestHeader)
 
 <fieldset class="fieldset">
 
@@ -21,7 +16,7 @@
                    * same size limit is imposed here.
                    *@
                    maxlength="30"
-                   value="@firstName"
+                   value="@privateFields.flatMap(_.firstName)"
                    class="input-text js-name-first"
                    required
                    aria-required="true"/>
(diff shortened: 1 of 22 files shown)


frontend/app/views/fragments/form/nameDetailFields.scala.html
```diff
@@ -21,7 +16,7 @@
                    * same size limit is imposed here.
                    *@
                    maxlength="30"
-                   value="@firstName"
+                   value="@privateFields.flatMap(_.firstName)"
```
paulbrown1982: Do you need .mkString here?

rtyley: Strictly speaking I don't need `.mkString` here - though I realise I've been inconsistent and used it in `stripeCheckout.scala.html`! Twirl renders `None` options as the empty string, and `Some[String]` objects with just the string content. The code I'm replacing with this commit was already taking advantage of this in fact.

Here's an example of this running OK in dev:

image

You can see there's no `Some(...)` wrapper in the page source:

image

However, I did use `.mkString` superfluously in `stripeCheckout.scala.html`, so I'll remove that.

paulbrown1982: 👍

svillafe:
The normal flow is working on IE and Chrome.

rtyley:
First name and Last name are correctly pre-populated on the standard flow:

image

...and the email is successfully passed through:

image


...turning on MergedRegistration, I can reach https://membership.theguardian.com/join/supporter/enter-details without signing in:

image

...but as anticipated, I can not yet complete payment.

PR #1544, opened 2017-03-15
Paid comments AB test

JustinPinner:
## Why are you doing this?
We're testing paid commenting and have opened a PR in frontend to introduce a 0% AB test. This is the back-end of that test.

## Changes
-Add a new bundle variant
-Drop the 'give' section from the landing page
-Allow for more breakpoints than just `mem-full` (this still needs some effort)
-Drop a cookie when the user selects one of the paid options
-Changes the copy a little, to better reflect what the user expects

## Screenshots
See the [frontend PR](https://github.com/guardian/frontend/pull/16149) (there's a big GIF there!)

cc @Ap0c @svillafe

PR #1547, opened 2017-03-16
Default test cookie to theguardian.com domain

JustinPinner:
## Why are you doing this?
Follows on from https://github.com/guardian/membership-frontend/pull/1544. Testing on CODE showed that the domain name generated for the cookie wasn't good enough when the bundles test is completed by the user.

As this was only written to allow testing on thegulocal.com, we can just use theguardian.com now (and will have to override it manually when testing locally in future).

cc @rupertbates @Ap0c @svillafe

### DIFF
--- a/frontend/assets/javascripts/src/modules/landingBundles.es6
+++ b/frontend/assets/javascripts/src/modules/landingBundles.es6
@@ -20,10 +20,7 @@ const PRINT_CTA = document.querySelector(PRINT_CTA_SELECTOR);
 const COMMIT_CTA = document.querySelector(COMMIT_BUTTON_SELECTOR);
 const COMMIT_COOKIE_NAME = 'GU_PDCOMCTA';
 
-const cookieDomain = () => {
-    return document.location.host.substr(document.location.host.indexOf('.') ? document.location.host.indexOf('.') + 1 : 0);
-};
-const setCookie = (name, value, days = 7, path = '/', domain = cookieDomain()) => {
+const setCookie = (name, value, days = 7, path = '/', domain = 'theguardian.com') => {
     const expires = new Date(Date.now() + days * 864e5).toGMTString();
     document.cookie = name + `=${encodeURIComponent(value)}; expires=${expires}; path=${path}; domain=${domain}`;
 };


rupertbates:
What was the problem with the existing code?

JustinPinner:
Actually, this is still going to be a problem. We're testing frontend on dev-theguardian.com, not dev.theguardian.com, and membership-frontend only has a guardian.com instance (I think?), so the cookie won't drop properly.

Yet another case that's harder than it should be to test 😞 

Of course, it'll all work fine in PROD, but that doesn't help now. I'll have to re-think.

JustinPinner:
In case that last comment didn't make the problem clearer, the issue is that out frontend UI on CODE runs as a sub domain of `dev-theguardian.com`, while our membership frontend UI runs under a subdomain of `theguardian.com`. We can't read a cookie dropped for `theguardian.com` from `dev-theguardian.com` when we get back to the original article and comment.

That's why it'd be fine in PROD (everything is on `theguardian.com` or a subdomain thereof).

Unless someone knows how this might be possible for the CODE env, I think we're going to have to manually edit the cookie to demo it. Which is just unpleasant.

rupertbates:
I see what you mean, could you pass a querystring parameter and then set the cookie on membership according to that?

frontend/assets/javascripts/src/modules/landingBundles.es6
```diff
@@ -20,10 +20,7 @@ const PRINT_CTA = document.querySelector(PRINT_CTA_SELECTOR);
 const COMMIT_CTA = document.querySelector(COMMIT_BUTTON_SELECTOR);
 const COMMIT_COOKIE_NAME = 'GU_PDCOMCTA';
 
-const cookieDomain = () => {
-    return document.location.host.substr(document.location.host.indexOf('.') ? document.location.host.indexOf('.') + 1 : 0);
-};
-const setCookie = (name, value, days = 7, path = '/', domain = cookieDomain()) => {
+const setCookie = (name, value, days = 7, path = '/', domain = 'theguardian.com') => {
```
rtyley: You can use `Config.guardianShortDomain` for this!

JustinPinner:
Closing this as I ended up doing a PROD deploy to test it.

frontend/assets/javascripts/src/modules/landingBundles.es6
JustinPinner: I'll use that next time I'm working on this feature. Thanks!

PR #1550, opened 2017-03-17
Fix To Hide PayPal In Apps

Ap0c:
## Why are you doing this?

PayPal does not work in the app web views, so we need to hide the button for users in the apps. Previously we were doing this by checking for the presence of the iOS or Android campaign codes in the INTCMP query param. However, the apps are now passing through a much wider range of campaign codes, depending on where the user has come from. Therefore we need to switch to using a `platform` query param to figure out whether we are in the apps or not.

The iOS app has not gone into production with the INTCMP changes, so all we need to do there is check for `platform=ios`, which has been implemented in [this PR](https://github.com/guardian/ios-live/pull/3095). However, the Android app went into production with these changes yesterday, so until we can get the fix deployed next week, we need to temporarily check for the most common of the INTCMP codes it is now sending through to us.

[**Trello Card**](https://trello.com/c/ipFJgiI1/402-fix-paypal-in-apps-check-for-platform)

## Changes

- Check for platform query param, and don't show PayPal if it matches 'ios' or 'android'.
- Check for three most common android INTCMPs, and don't show PayPal if they are present.

## Screenshots

**Platform: iOS**

paypal-ios

**Android Campaign Code**

paypal-android

**Anything Else**

paypal-normal

cc @alfavata @maxspencer

### DIFF
--- a/frontend/assets/javascripts/src/modules/form/helper/formUtil.js
+++ b/frontend/assets/javascripts/src/modules/form/helper/formUtil.js
@@ -34,11 +34,20 @@ define([
     };
 
     function hasPaypal () {
-        var regex = /^APP_.*_MEMBERSHIP_PAYMENT_SCREEN$/; // matches Android (APP_ANDROID_MEMBERSHIP_PAYMENT), iOS (APP_IOS_MEMBERSHIP_PAYMENT) & Windows app (if they implement it one day)
-        var code = url.getQueryParameterByName('INTCMP');
 
-        var fromApps = regex.exec(code);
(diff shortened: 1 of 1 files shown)


maxspencer:
Looks good, we'll add the `platform=android` as soon as possible next week. Sorry for the oversight on that.

Ap0c:
Just ran through:

- IE10, UK Supporter, Stripe
- IE10, US Supporter, PayPal
- IE10, UK Friend -> Supporter, Stripe

All worked fine. Also, the platform and Android INTCMPs appear to be hiding the button correctly.

PR #1556, opened 2017-03-22
Replacing caret with tilde for webpack 2

svillafe:
## Why are you doing this?
This is to avoid npm bump our version of webpack, only the patches.

## Changes
* Update package.json

cc: @Ap0c


svillafe:
Stripe and Paypal supporter flow working on IE/UK

PR #1567, opened 2017-03-30
Send thankyou email

rupertbates:
## Why are you doing this?
When users sign up to our monthly contribution product we want to send them a thank you email. For v1 of this product we want to send the same email as is sent to one off contributors so I am piggy backing on their email mechanism which is triggered by putting a message onto an SQS queue.

### DIFF
--- a/frontend/app/utils/AwsAsync.scala
+++ b/frontend/app/utils/AwsAsync.scala
@@ -1,0 +1,29 @@
+package utils
+
+import com.amazonaws.AmazonWebServiceRequest
+import com.amazonaws.handlers.AsyncHandler
+import java.util.concurrent.{Future => JFuture}
+
+import scala.concurrent.{Future, Promise}
+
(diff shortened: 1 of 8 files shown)


JustinPinner:
Looks ok to me 👍

frontend/app/utils/AwsAsync.scala
```diff
@@ -0,0 +1,29 @@
+package utils
+
+import com.amazonaws.AmazonWebServiceRequest
+import com.amazonaws.handlers.AsyncHandler
+import java.util.concurrent.{Future => JFuture}
+
+import scala.concurrent.{Future, Promise}
+
+class AwsAsyncHandler[Request <: AmazonWebServiceRequest, Response] extends AsyncHandler[Request, Response] {
```
rupertbates: Yup that's correct. I guess we could refactor it out into membership-common but that doesn't  really feel like a good place for it either. I think this should all sit more naturally in the new site hopefully.

PR #1569, opened 2017-03-31
Remove duplication by payment-methods in `createContributor()`

rtyley:
## Why are you doing this?

This was prompted by the duplication in `createContributor()` increasing in size slightly with https://github.com/guardian/membership-frontend/pull/1567#discussion_r108980121

By abstracting the payment method a little earlier, we can remove a lot of the duplicated creation steps (they're really important steps, so it's good not to duplicate them!), and reduce the number of arguments passed to methods.

This is a pure refactor, and should not affect functionality!

### DIFF
--- a/frontend/app/services/MemberService.scala
+++ b/frontend/app/services/MemberService.scala
@@ -17,7 +17,7 @@ import com.gu.memsub.subsv2.reads.SubPlanReads._
 import com.gu.memsub.subsv2.services.{SubIds, _}
 import com.gu.memsub.subsv2.{SubscriptionPlan, _}
 import com.gu.memsub.util.Timing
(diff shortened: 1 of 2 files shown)


frontend/app/services/MemberService.scala
```diff
@@ -539,7 +535,7 @@ class MemberService(identityService: IdentityService,
         Account.payPal(contactId, currency, autopay = true)
     }
 
-  private def createPaymentMethod(contactId: ContactId, email: Option[String], paymentForm: CommonPaymentForm, stripeCustomer: Option[Customer]) =
```
rtyley: Note that `contactId: ContactId` was not really a necessary parameter on this method :smile:

rupertbates:
👍 Nice stuff

PR #1570, opened 2017-03-31
Change the link destinations and the nav size

rupertbates:
## Why are you doing this?
Alex B via Ben has asked for some minor changes to the text size of the navigation. 

## Changes
* Logos in the nav bar now link to the Guardian main site
* Reduce the size of the text in the nav bar for tablet and above


JustinPinner:
Looks ok to me. So we're never going larger than `mobileLandscape` in the font size? 👍

rupertbates:
@JustinPinner that's right

PR #1574, opened 2017-03-31
Update Acceptance Tests To Include US

Ap0c:
## Why are you doing this?

We want to make sure we don't break the US flow by accident. Now that the acceptance tests do two runs, one for Stripe and another for PayPal, we can switch one to the US flow and keep one on the UK version.

[**Trello card**](https://trello.com/c/MwH3sX1n/364-we-should-update-our-acceptance-tests-to-capture-us-checkout-flow)

**Note**: The PayPal acceptance test is flaky locally due to the PayPal overlay doing stuff. If you want to run these locally, they pass fine using @mario-galic's sauce labs tunnel 🚂.

## Changes

- Added new method for filling out US address details.
- Switched Stripe flow over to checking the US checkout.

### DIFF
--- a/frontend/test/acceptance/pages/EnterDetails.scala
+++ b/frontend/test/acceptance/pages/EnterDetails.scala
@@ -88,9 +88,7 @@ case class EnterDetails(val testUser: TestUser) extends Page with Browser {
       setSingleSelectionValue(country, "US")
       setValue(addressLine1, "222 Broadway")
       setValue(town, "New York")
-
-      val state = id("state-deliveryAddress")
-      setSingleSelectionValue(state, "New York")
+      setSingleSelectionValue(id("state-deliveryAddress"), "New York")
       setValue(postCode, "10038")
     }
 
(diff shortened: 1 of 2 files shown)


frontend/test/acceptance/pages/EnterDetails.scala
```diff
@@ -75,13 +77,23 @@ case class EnterDetails(val testUser: TestUser) extends Page with Browser {
...
       setValue(town, "London")
       setValue(postCode, "N1 9GU")
     }
 
+    def fillInUS() = {
+      setSingleSelectionValue(country, "US")
+      setValue(addressLine1, "222 Broadway")
+      setValue(town, "New York")
+
+      val state = id("state-deliveryAddress")
```
mario-galic: You could inline this.

Ap0c: f273914

Ap0c:
Sigh, sign-ups are ok, but the PayPal sandbox is down, which has nuked the acceptance tests. Will re-run them when it comes back up 😤.

Ap0c:
It's back up, things look good ^.

PR #1575, opened 2017-03-31
Client side validation of the email for the merge checkout flow

svillafe:
## Why are you doing this?

Under the context of the 'merge checkout flow' test, we would like to give feedback to the user as soon as possible about the fact if the proposed email is already used in the database or not. Therefore in this PR, we are introducing a new route and its correspond method in order to check if the user already exists in the database (this method is currently hardcoded to return always true). From the client-side point of view, since we are making an asynchronous validation and the current validation system (validity.js) does not support this type of validation, we implement the email validation outside the current validation framework. This decision was made after a discussion with the team, mainly based on the fact that we are going to replace this code in the near future with the new platform.

## Tested in
| Browser | Supporter Stripe checkout | Supporter Paypal checkout | Stripe Monthly Contribution |  Paypal Monthly Contribution| Upgrade from Friend to Supporter  |
|---------|---------------------------|---------------------------|-----------------------------|-----------------------------|-----------------------------------|
| IE      |                           |                           |                             |                             |                                   |
| Chrome  |                           |                           |                             |                             |                                   |

## Screenshots

picture 239

### DIFF
--- a/frontend/app/services/IdentityService.scala
+++ b/frontend/app/services/IdentityService.scala
@@ -18,12 +18,17 @@ import play.api.Play.current
 import play.api.libs.concurrent.Execution.Implicits.defaultContext
 import play.api.libs.json.Json.toJson
 import play.api.libs.json._
-import play.api.libs.ws.{WS, WSResponse}
+import play.api.libs.ws.{WS, WSRequest, WSResponse}
 import services.IdentityService.DisregardResponseContent
 import views.support.IdentityUser
 
 import scala.concurrent.Future
 import scala.concurrent.duration._
+import scala.util.{Failure, Success}
+import dispatch._
+import Defaults.timer
+import play.api.http.Status
+import play.api.mvc.Results
 
 case class IdentityServiceError(s: String) extends Throwable {
   override def getMessage: String = s
@@ -86,15 +91,29 @@ case class IdentityService(identityApi: IdentityApi) {
         IdentityUser(fullUser, doesPasswordExist)
       }
 
+  def doesUserExist(email: String)(implicit idReq: IdentityRequest): Future[Boolean] = (identityApi.get("user",
+    idReq.headers,
+    idReq.trackingParameters :+ ("emailAddress" -> email),
+    "get-user-by-email") { resp =>
+    Right(resp.status == Status.OK)
+  }).valueOr(_ => false)
+
   def getFullUserDetails(user: IdMinimalUser)(implicit identityRequest: IdentityRequest): Future[IdUser] =
     retry.Backoff(max = 3, delay = 2.seconds, base = 2){ () =>
-      identityApi.get(s"user/${user.id}", identityRequest.headers, identityRequest.trackingParameters)
+      getUser(user).value
     }.map(_.getOrElse{
       val guCookieExists = identityRequest.headers.exists(_._1 == "X-GU-ID-FOWARDED-SC-GU-U")
       val guTokenExists = identityRequest.headers.exists(_._1 == "Authorization")
       val errContext = s"SC_GU_U=$guCookieExists GU-IdentityToken=$guTokenExists trackingParamters=${identityRequest.trackingParameters.toString}"
       throw IdentityServiceError(s"Couldn't get user's ${user.id} full details. $errContext")
     })
+
+  def getUser(user: IdMinimalUser)(implicit idReq: IdentityRequest): EitherT[Future, String, IdUser] = identityApi.get(s"user/${user.id}",
+    idReq.headers,
+    idReq.trackingParameters,
+    "get-user") {
+      _.json.validate[IdUser].asEither.leftMap(_.mkString(","))
+    }
 
   def doesUserPasswordExist(identityRequest: IdentityRequest): Future[Boolean] =
     identityApi.getUserPasswordExists(identityRequest.headers, identityRequest.trackingParameters)
@@ -149,25 +168,13 @@ trait IdentityApi {
     }
   }
 
-  def get(endpoint: String, headers:List[(String, String)], parameters: List[(String, String)]) : Future[Option[IdUser]] = {
-    Timing.record(IdentityApiMetrics, "get-user") {
-      val url = s"${Config.idApiUrl}/$endpoint"
-      WS.url(url).withHeaders(headers: _*).withQueryString(parameters: _*).withRequestTimeout(1000).get()
-        .recover { case e =>
-          Logger.error("Failure trying to retrieve user data", e)
-          throw e
-        }
-        .map { response =>
-          recordAndLogResponse(response.status, "GET user", endpoint)
-          val jsResult = (response.json \ "user").validate[IdUser]
-          if (jsResult.isError) Logger.error(s"Id Api response on $url : ${response.json.toString}")
-          jsResult.asOpt
-        }
-        .recover { case e =>
-          Logger.error("Failure trying to deserialise user data", e)
-          None
-        }
-    }
+  def get[A](
+    endpoint: String,
+    headers: List[(String, String)],
+    parameters: List[(String, String)],
+    metricName: String)(func: WSResponse => Either[String, A]): EitherT[Future, String, A] = {
+    execute(endpoint, metricName, func,
+      WS.url(s"${Config.idApiUrl}/$endpoint").withHeaders(headers: _*).withQueryString(parameters: _*).withRequestTimeout(1000).withMethod("GET"))
   }
 
   def post[A](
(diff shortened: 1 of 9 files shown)


frontend/assets/javascripts/src/modules/form/validation/existEmail.es6
```diff
@@ -0,0 +1,32 @@
+import ajax from 'ajax';
+import debounce from 'lodash/debounce'
+
+export function init() {
+    const emailField = document.getElementById('email');
+    const isUseErrorMessage = document.getElementById('form-field__error-message-email-checker');
+    const checkerURL = '/user/check-existing-email';
+
+    const validateEmail = debounce((event) => {
+        debugger;
```
Ap0c: ?

frontend/assets/javascripts/src/modules/form/validation/existEmail.es6
```diff
@@ -0,0 +1,32 @@
...
+        let prospectiveEmail = emailField.value;
+
+        if(!prospectiveEmail.includes('@') || !prospectiveEmail.includes('.')) {
```
Ap0c: What does the `.` tell you?

frontend/assets/javascripts/src/modules/form/validation/existEmail.es6
```diff
@@ -0,0 +1,32 @@
...
+            return;
+        }
+
+        ajax({
+            url: checkerURL + '?email=' + emailField.value,
```
Ap0c: Is there a danger from not url encoding this?

frontend/assets/javascripts/src/modules/form/validation/existEmail.es6
```diff
@@ -0,0 +1,32 @@
...
+            method: 'get',
+            success: ({emailInUse}) => {
+                if(emailInUse){
+                    isUseErrorMessage.classList.remove('is-hidden')
```
Ap0c: `;`

frontend/assets/javascripts/src/modules/form/validation/existEmail.es6
```diff
@@ -0,0 +1,32 @@
...
+                }else {
+                    isUseErrorMessage.classList.add('is-hidden')
```
Ap0c: `;`

frontend/assets/javascripts/src/modules/form/validation/existEmail.es6
```diff
@@ -0,0 +1,32 @@
...
+                }
+            },
+        });
+
+    }, 250);
+
+    emailField.addEventListener('keyup', validateEmail)
```
Ap0c: `;`

frontend/assets/stylesheets/components/_payment-paypal.scss
```diff
@@ -221,6 +221,7 @@
         background-color: #fff;
         top: 2px;
         margin-left: -12px;
+        border: 1px solid $c-neutral4;
```
Ap0c: Ok so this is technically the same as `guss-colour(neutral-4)`, but tbh I have no idea what our convention is at this point, we have a mix throughout the codebase. Do you prefer to use the guss functions, or the defined colours as you've done here?

frontend/app/views/joiner/form/payment.scala.html
```diff
@@ -54,6 +56,12 @@ <h1 class="checkout-headline">@fragments.form.formCta(plans.tier)</h1>
 
                     <div class="form-panel__heading">
                         <h2 class="form-panel__headline">Name and address</h2>
+
+                          <div class="text-note {is-hidden}">
```
Ap0c: What do the curly braces do here?

frontend/assets/javascripts/src/modules/form/helper/formUtil.js
```diff
@@ -60,6 +60,10 @@ define([
         return !!document.querySelector('.js-continue-name-address');
     }
 
+    function hasEmailInput () {
+        return !!document.getElementById('email') && (document.getElementById('email') instanceof HTMLInputElement)
```
Ap0c: `;`

frontend/app/services/IdentityService.scala
```diff
@@ -10,19 +10,85 @@ import monitoring.IdentityApiMetrics
...
+
+  def post(endpoint: String, data: Option[JsObject], headers: List[(String, String)], parameters: List[(String, String)], metricName: String): Future[Int] = {
+    Timing.record(IdentityApiMetrics, metricName) {
+      val requestHolder = WS.url(s"${Config.idApiUrl}/$endpoint").withHeaders(headers: _*).withQueryString(parameters: _*).withRequestTimeout(5000)
+      val response = requestHolder.post(data.getOrElse(JsNull))
+      response.foreach(r => recordAndLogResponse(r.status, s"POST $metricName", endpoint ))
+      response.map(_.status)
+        .andThen {
+          case Success(status) =>
+            if ((status / 100) != 2) // non 2xx code
```
davidfurey: I know you've just moved this from elsewhere in the file, but we should be able to write this line so that it is simple enough to not need a comment.

frontend/app/services/IdentityService.scala
```diff
@@ -10,19 +10,85 @@ import monitoring.IdentityApiMetrics
...
 
+trait IdentityApi {
+
+  def getUserPasswordExists(headers:List[(String, String)], parameters: List[(String, String)]) : Future[Boolean] = {
+    val endpoint = "user/password-exists"
+    val url = s"${Config.idApiUrl}/$endpoint"
+    Timing.record(IdentityApiMetrics, "get-user-password-exists") {
+      WS.url(url).withHeaders(headers: _*).withQueryString(parameters: _*).withRequestTimeout(1000).get().map { response =>
+        recordAndLogResponse(response.status, "GET user-password-exists", endpoint)
+        (response.json \ "passwordExists").asOpt[Boolean].getOrElse(throw new IdentityApiError(s"$url did not return a boolean"))
```
davidfurey: Should the failure cause be logged? A similar error is on line 75.

frontend/app/services/IdentityService.scala
```diff
@@ -10,19 +10,85 @@ import monitoring.IdentityApiMetrics
...
 
 case class IdentityServiceError(s: String) extends Throwable {
   override def getMessage: String = s
 }
 
+trait IdentityApi {
+
+  def getUserPasswordExists(headers:List[(String, String)], parameters: List[(String, String)]) : Future[Boolean] = {
+    val endpoint = "user/password-exists"
+    val url = s"${Config.idApiUrl}/$endpoint"
```
davidfurey: I would move these two vals inside the timing (for consistency with `def get` below)

frontend/assets/javascripts/src/modules/form/validation/existEmail.es6
svillafe: @Ap0c The idea is to avoid an unnecessary request to the API. Therefore, I check that the email has at least a `@` and a `.` for the top level domain.

frontend/assets/javascripts/src/modules/form/validation/existEmail.es6
svillafe: @Ap0c Currently the email checker service is hard-coded and the response does not depend on the input. Once the service is implemented, we will be able to test if we need to URL-enconded or not this parameter. This PR should include that service logic, I will talk with @rtyley about this logic.

frontend/assets/javascripts/src/modules/form/validation/existEmail.es6
Ap0c: Ok, makes sense. I was just thinking that lots of email addresses have dots earlier on, e.g. `first name.lastname@theguardian.com`. If you're specifically checking for the TLD `.`, is it worth doing something like:

```
!prospectiveEmail.split('@')[1].includes('.')
```

I realise this is just a quick check, so don't bother if you think it's not worth it =).

frontend/assets/javascripts/src/modules/form/validation/existEmail.es6
Ap0c: I was thinking more about the fact that special characters in the email address with mess with the request URL. For example, an email address with an `&` in it will screw up the query params.

frontend/assets/javascripts/src/modules/form/validation/existEmail.es6
```diff
@@ -0,0 +1,31 @@
...
+
+export function init() {
+    const emailField = document.getElementById('email');
+    const isUseErrorMessage = document.getElementById('form-field__error-message-email-checker');
+    const checkerURL = '/user/check-existing-email';
+
+    const validateEmail = debounce(() => {
+        let prospectiveEmail = emailField.value;
+
+        if(!prospectiveEmail.includes('@') || !prospectiveEmail.includes('.')) {
```
davidfurey: Pedantic maybe, but it is possible to have a valid email address that doesn't have any `.`s.  A number of country tlds have MX records (e.g. `.ai`, so `test@ai` may be a valid address)

frontend/assets/javascripts/src/modules/form/validation/existEmail.es6
svillafe: I totally forgot the semicolon in that line. Nevertheless, a little note for the next project, we should probably consider remove `;` from js files. In addition, we should add a JS code validator.

frontend/assets/javascripts/src/modules/form/validation/existEmail.es6
```diff
@@ -14,7 +14,7 @@ export function init() {
         }
 
         ajax({
-            url: checkerURL + '?email=' + emailField.value,
+            url: checkerURL + '?email=' + encodeURIComponent(emailField.value),
```
Ap0c: Nice!

rtyley:
I've updated server-side endpoint for checking if an email address is already in use, I think this PR is now ready to go out, @svillafe would you agree?

rtyley:
@mr-mr the email-in-use warning we show with this change is _"This email is already used in our database, please use another email"_:

image

I think the use of the word 'database' is off-putting, and the most likely case is that the user actually has a Guardian Identity account with that email and would probably prefer to use that address. 

[**subscribe.theguardian.com**](https://subscribe.theguardian.com/checkout) has a better message, which is: _"You already have a Guardian account. Please sign in or use another email address."_

image

I'd prefer to switch to that, is that ok with you?

rtyley:
> subscribe.theguardian.com has a better message, which is: "You already have a Guardian account. Please sign in or use another email address."

I've now updated this branch with ec8b31b to use this improved message - screenshot:

image

PR #1580, opened 2017-04-04
Bundles Landing Price Tweaks

Ap0c:
## Why are you doing this?

The prices for some of the bundles are wrong. The error message isn't visible enough.

## Changes

- Update the bundles prices.
- Make the error message more visible.

## Screenshots

updated-prices


JustinPinner:
Hmm, I don't know... Maybe we should just SHIPIT! 👍 🚢

PR #1564, opened 2017-03-29
Bundles Landing Page

Ap0c:
## Why are you doing this?

This is the landing page for the new simple and coherent bundles.

[**Trello Card**](https://trello.com/c/7P8Cl1FV/332-build-new-bundle-landing-page-links-out-to-old-pages)

## Changes

- Added landing page to bundles controller.
- Added bundles landing template.
- Added svgs.
- Added JavaScript to allow the user to select contribution type, amount and print type.
- Added 'Guardian Titlepiece Web' font.
- Added a new stylesheet for the landing page.
- Added the new page to `/uk` route for the moment.
- Added the page to `robots.txt` until we're ready to launch.

## Screenshots

**Bundles Landing Desktop**

bundles-landing-desktop

**Bundles Landing Tablet**

bundles-landing-tablet

**Bundles Landing Mobile Landscape**

bundles-landing-mob-landscape

**Bundles Landing Mobile**

bundles-landing-mobile

PR #1583, opened 2017-04-04
Make Bundles Landing Page Available To Web Crawlers (Remove From robots.txt)

Ap0c:
## Why are you doing this?

We want to drive traffic to the bundles landing page from Google. Also, adding `/uk` to robots.txt has been messing with Google traffic to `/uk/supporter`.

## Changes

- Removed `/uk` from robots.txt (it was added in #1564).


svillafe:
🐻

PR #1588, opened 2017-04-10
Add attempt/success metrics for monthly contributions

davidfurey:
## Why are you doing this?

So that we can monitor the number of contributor attempts/successes to detect faults in the system.

## Changes
* Add metric filter for new monthly contributor attempt
* Add metric filter for successful new monthly contributor


JustinPinner:
LGTM 👍

PR #1294, opened 2016-08-02
Remove Giraffe and redirect to contribute.theguardian.com

rtyley:
Contributions is now at:

* https://contribute.theguardian.com/ <- will become the destination of https://membership.theguardian.com/contribute
* https://contribute.theguardian.com/uk <- will become the destination of https://membership.theguardian.com/uk/contribute
...etc

I'd lik to release this by 4pm London time today, otherwise it's gonna be tomorrow...

cc @AWare

- [x] https://contribute.theguardian.com/ geo-location works correctly - done with https://app.fastly.com/#/configure/service/3eAImrcMhkiYNBpDzEIpbx/diff/2,3 - which passes the `X-GU-GeoIP-Country-Code` header through
- [x] Pass on `INT` `CMP` and `mcopy` campaign codes

PR #1589, opened 2017-04-13
Allow users who don't have an Identity account to complete checkout

rtyley:
## Why are you doing this?

This work follows on from https://github.com/guardian/membership-frontend/pull/1539, which allowed unauthenticated users to view the checkout page - this change actually makes checkout work! When the user hits the buy button (eg Credit Card or PayPal), an Identity account is created for the user as a first step, before the rest of the Salesforce/Zuora processing occurs.

The feature is hidden behind the `MergedRegistration` feature switch:

Once that cookie is dropped, you can go to the checkout page without already being signed in:

https://membership.theguardian.com/join/supporter/enter-details

## Changes
The classes under the new `services.checkout.identitystrategy` Scala package handle the Identity-related tasks around checkout:

* `NewUser` : Creates a new Identity account
* `ExistingSignedInUser` : Gets & *updates* the user's existing Id account

This also simplifies the implementation code for `MemberService.createMember()`.

This change uses the new Create-And-Authenticate-User Identity API call added with https://github.com/guardian/identity/pull/621, as well as the handy authentication-cookie classes added with https://github.com/guardian/identity/pull/622. There was also supporting code in https://github.com/guardian/identity-play-auth/commit/0babd70f02f5b8c089af8ef734f01daf8c6e288b

I've also introduced Cats for the the Identity API client, I realise we still have Scalaz all over the place... I apologise for having increased the number of FP libraries to 2.

There's still another PR to check if the email address is already in use: https://github.com/guardian/membership-frontend/pull/1575

cc @dominickendrick

### DIFF
--- a/frontend/app/controllers/Joiner.scala
+++ b/frontend/app/controllers/Joiner.scala
@@ -215,16 +215,6 @@ object Joiner extends Controller with ActivityTracking with PaymentGatewayErrorH
         s"There has been an error in updating your email. You may already have an Identity account with $googleEmail. Please try signing in with that email."),
       _ => Redirect(routes.Joiner.enterStaffDetails()).flashing("success" ->
         s"Your email address has been changed to $googleEmail"))
-//      yield {
-//        responseCode match {
-//          case 200 => Redirect(routes.Joiner.enterStaffDetails())
-//            .flashing("success" ->
-//              s"Your email address has been changed to $googleEmail")
-//          case _ => Redirect(routes.Joiner.staff())
-//            .flashing("error" ->
-//              s"There has been an error in updating your email. You may already have an Identity account with $googleEmail. Please try signing in with that email.")
-//        }
-//      }
   }
 
   def unsupportedBrowser = CachedAction(Ok(views.html.joiner.unsupportedBrowser()))
--- a/frontend/app/controllers/Joiner.scala
+++ b/frontend/app/controllers/Joiner.scala
@@ -8,8 +8,7 @@ import com.gu.contentapi.client.model.v1.{MembershipTier => ContentAccess}
 import com.gu.i18n.CountryGroup
 import com.gu.i18n.CountryGroup.UK
 import com.gu.i18n.Currency.GBP
-import com.gu.memsub.BillingPeriod
-import com.gu.memsub.util.Timing
+import com.gu.identity.play.{AuthenticationService => _, _}
 import com.gu.salesforce._
 import com.gu.stripe.Stripe
 import com.gu.stripe.Stripe.Serializer._
@@ -17,7 +16,7 @@ import com.gu.zuora.soap.models.errors._
 import com.netaporter.uri.dsl._
 import com.typesafe.scalalogging.LazyLogging
 import configuration.{Config, CopyConfig}
-import forms.MemberForm._
+import forms.MemberForm.{paidMemberJoinForm, _}
 import model._
 import play.api.Play.current
 import play.api.i18n.Messages.Implicits._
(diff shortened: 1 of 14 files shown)


frontend/app/controllers/Joiner.scala
```diff
@@ -210,53 +209,66 @@ object Joiner extends Controller with ActivityTracking with PaymentGatewayErrorH
...
-              s"There has been an error in updating your email. You may already have an Identity account with $googleEmail. Please try signing in with that email.")
-        }
-      }
+      emailUpdateResult <- identityService.updateEmail(request.identityUser, googleEmail)(IdentityRequest(request)).value
+    } yield emailUpdateResult.fold(
+      _ => Redirect(routes.Joiner.staff()).flashing("error" ->
+        s"There has been an error in updating your email. You may already have an Identity account with $googleEmail. Please try signing in with that email."),
+      _ => Redirect(routes.Joiner.enterStaffDetails()).flashing("success" ->
+        s"Your email address has been changed to $googleEmail"))
+//      yield {
```
svillafe: Is it possible to delete these comments?

frontend/app/services/IdentityService.scala
```diff
@@ -44,80 +99,40 @@ case class IdentityService(identityApi: IdentityApi) {
...
+    identityApi.post("/user/password", Some(Json.obj("newPassword" -> password)), idReq.headers, idReq.trackingParameters, "update-user-password")(DisregardResponseContent)
   }
 
-  def updateUserPassword(password: String, identityRequest: IdentityRequest, userId: String) {
-    val json = Json.obj("newPassword" -> password)
-    identityApi.post("/user/password", Some(json), identityRequest.headers, identityRequest.trackingParameters, "update-user-password")
-  }
-
-  def updateUserMarketingPreferences(req: IdentityRequest, user: IdMinimalUser, allowMarketing: Boolean) =
-    postFields(Json.obj("statusFields.receiveGnmMarketing" -> allowMarketing), user.id, req)
```
paulbrown1982: Where did this setting go in the refactored flow?

rtyley: Good point! I will address.

rtyley: So `updateUserMarketingPreferences()` was actually unused anywhere in membership-frontend - it's been unused since https://github.com/guardian/membership-frontend/pull/1294 was merged last year. However, it's worth noting that I haven't yet added a _"Keep me up to date with offers from the Guardian"_ checkbox, and so I am _not_ setting `receiveGnmMarketing` to true when I create the user.

See the checkbox here on https://profile.theguardian.com/register :

image
 

That checkbox will probably need to be added to the checkout on Membership Frontend before going fully live with this flow - it's still hid behind a feature flag tho', so I feel happy to leave that change to later.

svillafe:
I have tested the Supporter and monthly contributor flow using Stripe and are working. Nevertheless, when I tried to run the PayPal checkout flow, it failed. See image below:

picture 246

rtyley:
I'm merging this code because I'd like to give it time to bed in - I'll be monitoring it over the next few hours and will revert it if bad things show up. @svillafe and I have reviewed the code, run through the important flows and found a couple of issues, but none block this code going out:

* The acceptance tests are currently failing in Dev for the PayPal run - *but they do this for `master` too*
* When unauthenticated users try to complete with PayPal they get the error reported by @svillafe above - but this is only for users with the `MergedRegistration` feature switch turned on - and it will remain turned off for users by default - the general public are not going to be encountering this issue yet.

svillafe:
Tested **monthly-contribution** on **PROD** and is working (both **PayPal** and **Stripe**).

rtyley:
As a test-user, I've manually been able to sign up as a Supporter using the conventional flow (authenticating before going to the checkout page), succesfully paying with both Stripe and PayPal:

image
image

PR #1594, opened 2017-04-18
upgrade membership common version to 0.390 

lmath:
Upgrade to version 0.390 of membership-common. PR [453](https://github.com/guardian/membership-common/pull/453) in membership common means that now is the time to fetch email from Zuora rather than salesforce. 

## Why are you doing this?

Keep up to date with membership common. 

## Changes
* Use a newer version of membership common
* This means that we need to get email from Zuora rather than from Salesforce.

### DIFF
--- a/frontend/app/services/MemberService.scala
+++ b/frontend/app/services/MemberService.scala
@@ -8,7 +8,7 @@ import com.gu.i18n.Currency.GBP
 import com.gu.i18n.{Country, CountryGroup, Currency}
 import com.gu.identity.play.{IdMinimalUser, IdUser}
 import com.gu.memsub.Subscriber.{FreeMember, PaidMember}
-import com.gu.memsub.Subscription.{Feature, ProductRatePlanId, RatePlanId}
+import com.gu.memsub.Subscription.{Feature, ProductRatePlanId, RatePlanId, AccountId}
 import com.gu.memsub.promo.PromotionApplicator._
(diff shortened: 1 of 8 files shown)


lmath:
Currently seeking any suggestions, but particularly including the best way to avoid doing this 4 times:
```
val emailFromZuora = zuoraRestService.getAccount(request.subscriber.subscription.accountId) map { email =>
      email match {
        case \/-((accountSummary: AccountSummary)) => accountSummary.billToContact.email
        case -\/(_) => None
      }
    }
````

File: frontend/app/services/MemberService.scala [hunk 1]
```diff
@@ -585,7 +587,8 @@ class MemberService(identityService: IdentityService,
                                         payPalBaid: String): Future[UpdateResult] =
     for {
       sub <- subscriptionService.current[SubscriptionPlan.Member](contact).map(_.head)
-      result <- zuoraService.createPayPalPaymentMethod(sub.accountId, payPalBaid, contact.email.getOrElse(""))
+      maybeEmail <- zuoraRestService.getAccount(sub.accountId) map (_.toOption.flatMap(_.billToContact.email))
```
johnduffell: it's a shame we have to make an extra call here now, I wonder what `subscriptionService.current` does that we can't get from zuora.  Now that we have the identity id in zuora it might be doable @jacobwinch

PR #1595, opened 2017-04-18
Ensure we set the full name country in Salesforce

paulbrown1982:
## Why are you doing this?
Currently Membership sets the country code in Salesforce to a 2-letter ISO code. This is different to subscriptions-frontend and Identity, so this change mitigates the risk of breaking international paper deliveries if a subscriber goes on to become a member.

## Changes
- Use the full name for countries in Salesforce to make Membership consistent with subs.

## Screenshots
picture 78

Also see: https://github.com/guardian/subscriptions-frontend/pull/915

cc @jacobwinch @rtyley @Ap0c @rupertbates @ajosephides

### DIFF
--- a/frontend/app/services/SalesforceService.scala
+++ b/frontend/app/services/SalesforceService.scala
@@ -1,11 +1,12 @@
 package services
 
-import com.gu.i18n.Country._
 import com.gu.identity.play.{IdMinimalUser, IdUser}
 import com.gu.salesforce.ContactDeserializer.Keys
 import com.gu.salesforce._
 import com.gu.stripe.Stripe.Customer
-import forms.MemberForm.{MonthlyContributorForm, CommonForm, JoinForm}
+import dispatch.Defaults.timer
+import dispatch._
+import forms.MemberForm.{CommonForm, JoinForm, MonthlyContributorForm}
 import model.GenericSFContact
 import monitoring.MemberMetrics
 import play.api.Play.current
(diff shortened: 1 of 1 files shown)


jacobwinch:
Sounds sensible to me 👍 

However, I'm not completely sure where else the 2-letter country code might be used - could this impact on reporting, for example?

rtyley:
Consistency is good, but I think ideally we would be using 2-letter ISO code everywhere - country _names_ have mutable spellings (https://github.com/guardian/membership-common/commit/2d1d929009ce38c0f552aa090d2f7d7ad7d13216) for just one example - so using ISO codes as our source of truth seems like a good idea to me? If there are ouputs that require the text of a country name, it feels to me that it would be better to have that as a final translation step?

paulbrown1982:
I agree in principle, however Salesforce stores the country as a plain text field. As such we already have 4 sets of string data representing countries.

- Migration from QSS has put full name there, with the exception of United States being USA
- Subscriptions puts the full name string from CountryGroup.scala
- Membership puts the 2-letter country code from CountryGroup.scala
- Manual data entry via the call centre puts whatever the agent types in.

In terms of output translation:

- For subscriptions manual fulfilment within Salesforce there is no translation applied to any field, and so our suppliers are spending hours normalising the 2-letter codes to be appropriate for printing on address labels.
- For Membership plus pack, the supplier has in the past managed to build something to automate that normalisation from 2-letter to printable package label.
- When retrieving a contact in Zuora's API, it only ever outputs the full country name.

Therefore I'd rather just render the field untrustworthy and store the country in Salesforce as a human readable/post package printable text rather than a computer-understandable enum.

paulbrown1982:
Confirmation from the membership gift supplier: "For confirmation, changing to the full name is no issue for us and, will in fact, help us with the country sortation."

PR #1596, opened 2017-04-19
Update Sentry Raven library to v8.0.2

rtyley:
## Why are you doing this?

To get rid of these warnings we've been seeing:

"This event was reported with an old version of the java SDK."

E.g. on https://sentry.io/the-guardian/membership/issues/254948502/

Note we may also want to update `raven-js` - we're using `3.0.4` and the latest is `3.14.2`: https://github.com/getsentry/raven-js/releases


jacobwinch:
👍

PR #1597, opened 2017-04-20
Fix PayPal payments on the not-yet-Identity-authenticated flow

rtyley:
## Why are you doing this?

This follows on from #1589, where @svillafe [discovered](https://github.com/guardian/membership-frontend/pull/1589#issuecomment-293890744) we couldn't get PayPal payments to complete.

## Changes
This change removes the requirement that the user be authenticated to hit these endpoints:

* `POST /paypal/setup-payment`
* `POST /paypal/create-agreement`

This means they can hit the PayPal button before we've created them an Identity account. The main bit of functionality this affects is Test-Users, as the logged-in user was used to determine which PayPal backend to use. With this change, the `pre-signin-test-user` cookie (dropped by visiting https://membership.theguardian.com/test-users) is now permitted as an alternative way of selecting the Test backend.

There's also some small tidy-ups here, using Play's [explicit body parsers](https://www.playframework.com/documentation/2.5.x/ScalaBodyParsers#Choosing-an-explicit-body-parser) to avoid having to do the Json parsing/validation ourselves :sparkles:

image

### DIFF
--- a/frontend/app/controllers/PayPal.scala
+++ b/frontend/app/controllers/PayPal.scala
@@ -1,9 +1,11 @@
 package controllers
 
-import actions.AuthRequest
 import com.typesafe.scalalogging.LazyLogging
-import play.api.libs.json.{JsError, JsSuccess, Json, Reads}
-import play.api.mvc.{AnyContent, Controller}
+import play.api.libs.json.Json.toJson
+import play.api.libs.json._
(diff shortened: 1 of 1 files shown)


rtyley:
Using conventional flow, able to sign-up using PayPal:

image

...and using the new merged flow, still good:

image

PR #1601, opened 2017-04-24
Avoiding the hasEmailInput returns true in other screen.

svillafe:
## Why are you doing this?
We have seen some sentry logs ([here](https://sentry.io/the-guardian/membership-client-side/issues/258894041/)) returning undefined regarding the EmailChecker's code. This took place because `formUtil ` was using the id field as a way of checking if it has to call the init function of this module. Unfortunately, that id was being also used in the feedback form. Therefore, in this PR I am changing that condition, to check whether the paragraph `element form-field__error-message-email-checker` exists or not.

### DIFF
--- a/frontend/assets/javascripts/src/modules/form/helper/formUtil.js
+++ b/frontend/assets/javascripts/src/modules/form/helper/formUtil.js
@@ -61,7 +61,7 @@ define([
     }
 
     function hasEmailInput () {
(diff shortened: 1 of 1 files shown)


svillafe:
Tested the Supporter flow with Paypal in IE and with Stripe in Chrome and are working.

svillafe:
It is working on PRO. Tested the Supporter Stripe checkout with IE in prod and is working. In addition, tested the upgrade flow and is also working.

PR #1603, opened 2017-04-25
Shut off non-HTTPS access to mem.thegulocal.com

svillafe:
## Why are you doing this?
We should run the site under https. Therefore, we are updating the membership.conf file which contains the nginx configuration for DEV.


svillafe:
Riff-Raff had two failed deploys. In both cases, it struggled in creating the new instances.

PR #1604, opened 2017-04-27
Add some documentation on reading logs

rtyley:
## Why are you doing this?

@jacobwinch shared the excellent https://github.com/jorgebastida/awslogs tool with me, after I shared the Fastly logs information with him, and it seemed like a good idea to document this knowledge!

### DIFF
--- a/docs/logs.md
+++ b/docs/logs.md
@@ -1,0 +1,36 @@
+Fastly logs
+===========
+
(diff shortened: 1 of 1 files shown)


jacobwinch:
Good idea! 👍

Ap0c:
For future reference, we think the acceptance tests failed here because of issues with the deploy, hence the re-run.

PR #1609, opened 2017-05-03
Add receive gnm marketing checkbox

rtyley:
## Why are you doing this?

Spoke to Evie Wilson & Robert Chilvers on the Data Privacy team regarding the 'Receive Guardian Marketing' checkbox. They are happy for us to go ahead with the current wording as used on
https://profile.theguardian.com/register, ie:

    "Keep me up to date with offers from the Guardian"

Evie said it would be best to put the checkbox near where we're asking for the email, to make it clear what it was in relation to (so near the start of the flow, rather than the end, in our case). The checkbox must also default to 'off' for new users, this is a requirement of GDPR.

They pointed out that General Data Protection Regulation (GDPR) is coming into effect in 2018, at which point we will be required to return an audit trail, with each change being accompanied by:

* DateTime
* Application where change was made (eg Identity or Membership)
* What wording was used when the user accepted the mailing

...however, that is going to be a larger project which we do not need to implement now.

## Changes
*  Adding the "Keep me up to date with offers from the Guardian" checkbox

## Screenshots
image

### DIFF
--- a/frontend/app/views/fragments/form/marketingChoices.scala.html
+++ b/frontend/app/views/fragments/form/marketingChoices.scala.html
@@ -1,12 +1,29 @@
-@(marketingChoices: Option[com.gu.identity.play.StatusFields])
+@(marketingChoicesOpt: Option[com.gu.identity.play.StatusFields])
+@marketingChoicesOpt match {
+    case Some(marketingChoices) => {
+        <fieldset class="u-h">
+            <input type="hidden"
+            name="marketingChoices.gnnMarketing"
+            value="@marketingChoices.receiveGnmMarketing" />
 
-<fieldset class="u-h">
+            <input type="hidden"
+            name="marketingChoices.thirdParty"
+            value="@marketingChoices.receive3rdPartyMarketing" />
+        </fieldset>
+    }
+    case None => {
+        <div class="fieldset__fields">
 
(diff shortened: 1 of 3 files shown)


frontend/app/views/fragments/form/marketingChoices.scala.html
```diff
@@ -1,12 +1,29 @@
...
+            value="@marketingChoices.receive3rdPartyMarketing" />
+        </fieldset>
+    }
+    case None => {
+        <div class="fieldset__fields">
+
+            <div class="form-field js-billing-input-container">
+
+                <label class="label" for="use-delivery-address">
+                    <input type="checkbox" name="marketingChoices.gnnMarketing" id="receive-gnn-marketing" value="true"/>
```
rtyley: An HTML forms refresher (for me) - `value="true"` does not mean that the checkbox will default to checked, it just means that if it _is_ checked by the user, the value that will be sent encoded in the form data will be `true`.

https://www.w3.org/TR/html401/interact/forms.html#adef-value-INPUT

svillafe: Example of usage: https://www.w3schools.com/jsref/prop_checkbox_value.asp

PR #1616, opened 2017-05-04
Roll out new style supporter page to all countries except US

joelochlann:
## Why are you doing this?
There's evidence to suggest that the landing page design for the UK caused a strong uplift in conversions, estimated at 1.73% to 4.01% based on data here: https://docs.google.com/spreadsheets/d/1QzU656ip-TrF2PJTDmfsdOhK6FhYgxr0B-KtA4GuGYI/edit#gid=522088769

## Changes
* Point all countries except for the US at the new-style page
* A lot of code is now redundant so I've deleted it
* The US apparently will need specific copy which we don't have yet so we've left that one for now

## Australia
picture 535
## Canada
picture 536
## International
picture 537
## Europe
picture 538
## UK
picture 540
## US
picture 539


joelochlann:
cc @guardian/contributions

PR #1618, opened 2017-05-05
Change eventbrite.api.url & update membership-common to 4.02

michaelwmcnamara:
## Why are you doing this?
 We are seeing timeout issues with Eventbrite in sentry logs. We are trying to log it better and reduce the incidence of timeouts.

## Trello card: [Here](https://trello.com)
None

## Changes
1) Changed Eventbrite API url to go through fastly shield node - as we've seen issues with response times from cloudfront.
2) New logging in membership-common to track okhttp socket-timeout exceptions

## Tested in
Tested in events and masterclasses and signed up patron flow.

### DIFF
--- a/frontend/conf/application.conf
+++ b/frontend/conf/application.conf
@@ -16,7 +16,7 @@ event.discountMultiplier=0.8
 
 eventbrite.url="http://www.eventbrite.co.uk"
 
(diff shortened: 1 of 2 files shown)


rtyley:
We should also think about turning on `stale-while-revalidate` so that requests are _always_ served from the cache - even if new content is being fetched:

https://docs.fastly.com/guides/performance-tuning/serving-stale-content#usage

PR #1619, opened 2017-05-05
Increase timeout on PayPal requests from 10 to 20 seconds

rtyley:
## Why are you doing this?

We've seen quite a few failures in dev, because the PayPal sandbox is so slow - also in our acceptance tests [using test-users on PROD](https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logEventViewer:group=membership-frontend-PROD;stream=membership-frontend-PROD-application-i-02547072286a9a83d;refid=33314970187306741362686890102563553358816392836826202112;reftime=1493894930000).

## Changes
* Timeout increased on PayPal requests from 10 to 20 seconds
* Change requests to PayPal use Futures rather than blocking calls - prompted because the `configurableFutureRunner(20.seconds)` method gives a `Future`-based response.

Acceptance tests pass in dev (and hopefully slightly more reliably!)

### DIFF
--- a/frontend/app/controllers/PayPal.scala
+++ b/frontend/app/controllers/PayPal.scala
@@ -7,6 +7,9 @@ import play.api.mvc.{Controller, Request}
 import services.{PayPalService, TouchpointBackend}
 import utils.TestUsers.PreSigninTestCookie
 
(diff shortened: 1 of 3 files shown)


jacobwinch:
👍 👍 👍

PR #1621, opened 2017-05-08
Add 'become a supporter' link to main nav

joelochlann:
## Why are you doing this?
Based on the 84% bounce rate of people arriving on the supporter page, it's possible that people are heading to other parts of the site to understand what becoming a supporter means.

But we don't provide an easy means to get back to the supporter page. Since patrons only comprise 0.32% of our supporter equivalents (even accounting for the fact that a patron is worth 12 times a supporter) and they make it into the nav bar, it seems fair "become a supporter" should be there too

## Changes
* Add "become a supporter" to the top nav
* Put a campaign code on the link so we know if this work had any effect

## Screenshots
picture 549
picture 550

### DIFF
--- a/frontend/app/model/Nav.scala
+++ b/frontend/app/model/Nav.scala
@@ -19,6 +19,7 @@ object Nav {
   }
 
   val primaryNavigation = List(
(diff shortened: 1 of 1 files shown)


joelochlann:
cc @guardian/contributions

PR #1625, opened 2017-05-09
updating membership common to 4.02 in its own pr

michaelwmcnamara:
This is a partial rerun of #1618 - this change only updates the membership common version.

Why are you doing this?

We are seeing timeout issues with Eventbrite in sentry logs. We are trying to log it better 

Changes:
Update membership common version to 0.402 to pick up new logging which will track okhttp socket-timeout exceptions

Tested in:

Tested in events and masterclasses and signed up patron flow.


jacobwinch:
👍

michaelwmcnamara:
DEploy problem occurred - will revert.

PR #1630, opened 2017-05-11
Use global execution context in TouchpointBackend

jacobwinch:
## Why are you doing this?
We've had a lot of issues on app start-up since approx 21/04/2017. This is restricting our ability to deploy quickly - often deploys are taking 10+ minutes due to boxes cycling on start-up, and other deploys fail completely (see the riff-raff history for this project).

I've done a bit of digging, and suspect that the change in execution context might be linked to the problem. Locally this simple import change shows a big improvement, so I'd like to try it out in production before investigating this track too much more.

## Trello card: 
N/A

## Changes
* Use global execution context instead of Play's default execution context

cc @paulbrown1982 @michaelwmcnamara @johnduffell @rtyley @rupertbates @davidfurey

### DIFF
--- a/frontend/app/services/TouchpointBackend.scala
+++ b/frontend/app/services/TouchpointBackend.scala
@@ -26,7 +26,7 @@ import configuration.Config.Implicits.akkaSystem
 import model.FeatureChoice
 import monitoring.TouchpointBackendMetrics
 import org.joda.time.LocalDate
-import play.api.libs.concurrent.Execution.Implicits.defaultContext
+import scala.concurrent.ExecutionContext.Implicits.global
 import tracking._
 import utils.TestUsers.{TestUserCredentialType, isTestUser}
 import play.api.mvc.RequestHeader


jacobwinch:
@paulbrown1982 - I'm holding off on that until I see it work...

jacobwinch:
The deploy succeeded without cycling boxes, but we're still throwing errors on start-up...
image

johnduffell:
thanks Jacob, interesting, so are you thinking it's thread starvation (all threads waiting) rather than deadlock? or still not sure?  I think the thread pool we switched to doesn't expand when you enter a "blocking {...}" block which I was slightly concerned about, so maybe it's that. No threads left to do any actual work.

jacobwinch:
Not 100% sure but that's the possibility I was exploring...

Yes, it was reading this which started me off down this path:
“The number of concurrently blocking computations can exceed the parallelism level only if each blocking call is wrapped inside a blocking call (more on that below).”
http://docs.scala-lang.org/overviews/core/futures.html#the-global-execution-context

PR #1634, opened 2017-05-12
upgrading to membership common 0.402 - some logging improvements

michaelwmcnamara:
## Why are you doing this?
There is a PR to upgrade to membership common V0.407
Before that goes out I want to upgrade to V0.402 - this was experiencing startup issues last time I attempted it (see https://github.com/guardian/membership-frontend/pull/1625), but testing has shown that a recent change: https://github.com/guardian/membership-frontend/pull/1630 has fixed this in test.
That said I want to push it out in isolation to be sure it isnt still causing problems, before it gets wrapped up in later versions.

## Changes
Upgrades membershipCommon library to 0.402


johnduffell:
sounds sensible, thanks for doing that! 👍

PR #1605, opened 2017-04-28
Digital bundle price test 1 (AB)

JustinPinner:
## Why are you doing this?
We're about to test different price points for the digital pack subscription. This PR uses the original /bundles landing page, and adjusts it to work (functionally, at least) down to tablet breakpoints.

We also drop a cookie when someone clicks one of the intent-to-buy buttons, whether it's to follow up on the Digi Pack or to make a contribution.

The dotcom portion of this is (currently) on CODE, but that may need to be redeployed from time to time. 

## Changes
Should only affect the old /bundles landing page, so I'm not anticipating problems with any of the production features we're running.

### Dotcom thrasher
screen shot 2017-04-28 at 10 40 50

### Bundles landing page - wide
screen shot 2017-05-03 at 17 01 00 - bundles-1-wide

### Bundles landing page - desktop
screen shot 2017-05-03 at 17 00 00 - bundles-1-desktop

### Bundles landing page - tablet
screen shot 2017-05-03 at 16 58 00 - bundles-1-tablet

### Thank you page (tablet version shown)
screen shot 2017-05-03 at 17 04 00 - bundles-1-thankyou-tablet

PR #1631, opened 2017-05-12
Add AB Testing framework and Merged Registration AB test

rtyley:
## Why are you doing this?

Adds Merged Registration as an A/B test, along with a general purpose a/b testing framework for Membership, influenced by https://github.com/guardian/contributions-frontend/blob/fc51a3d7/app/abtests/Test.scala, with a few tweaks to how variants are expressed.

### Cookies

1. The 'audience id' or 'MVT' cookie, called `GU_MEM_mvt_id`, which identifies where the user is in the audience range (from 0 to 10000).
2. The AB-test specific cookies, eg `gu.membership.ab.merged-registration`, which specify which variant a user is in for a given test.

Example cookies:
```
Set-Cookie: GU_MEM_mvt_id=3672; Max-Age=315360000; Expires=Mon, 10 May 2027 10:15:18 GMT; Path=/; Secure
Set-Cookie: gu.membership.ab.merged-registration=control; Max-Age=31536000; Expires=Sat, 12 May 2018 10:15:18 GMT; Path=/
```

These cookies are dropped by the `AddAbTestingCookiesToResponse` Action, which is part of `NoCacheAction` and so dropped on any uncached page view.

### Allocating a variant to a user

The variant allocated to user is chosen using the first of these 3 methods that applies:

1. Forcing a variant with a query parameter, eg: https://mem.thegulocal.com/join/supporter/enter-details?ab.merged-registration=merged
2. Reading the AB-test-specific cookie, eg `gu.membership.ab.merged-registration`
3. Deriving a variant from the MVT id

### Client-Side

A `guardian.abTests` variable is created in `javaScriptFirstSteps` (though you do have to populate `PageInfo` to make this happen, so it's not on every page yet):

```
    var guardian = {
        ...
        abTests: {"merged-registration":"merged"},
        ...
    };
```

### Analytics

Google Analytics gets the `experience` property (custom dimension `dimension16`) set:

image

PR #1635, opened 2017-05-12
Switched to Using Array.prototype.forEach.call instead of NodeList.forEach In Bundles 404 Page

Ap0c:
## Why are you doing this?

#1605 introduced a [usage](https://github.com/guardian/membership-frontend/pull/1605/files#diff-fb2260c75d55c07e84cdf15072712903R42) of `NodeList.forEach`, which is unfortunately only supported in [modern browsers](https://developer.mozilla.org/en-US/docs/Web/API/NodeList/forEach#Browser_Compatibility). This resulted in this vague error in Safari 8:

foreach-safari

and this surprisingly more helpful error in IE10:

foreach-ie

This PR switches to using `[].forEach.call`.

Fortunately, because the `landingBundles.es6` script loads fairly far down in the initialisation order, the checkout flows don't seem to have been affected, so the impact is fairly minimal. I only noticed because the bundles landing page javascript stopped working (`bundlesLanding.es6`), because it happens to load further down. This may explain why we didn't notice this for a while.

[**Trello Card**](https://trello.com/c/P6DneiPJ/563-mysterious-foreach-error-in-ie-on-membership)

## Changes

- Switch use of `NodeList.forEach` to `Array.prototype.forEach.call`.

### DIFF
--- a/frontend/assets/javascripts/src/modules/landingBundles.es6
+++ b/frontend/assets/javascripts/src/modules/landingBundles.es6
@@ -40,7 +40,7 @@ export function init() {
 }
 
 function bindCommitButtonEvents() {
-    COMMIT_CTAS.forEach(function(el) {
+    [].forEach.call(COMMIT_CTAS, function(el) {
         el.addEventListener('click', function(evt){
             setCookie(COMMIT_COOKIE_NAME, 1, COMMIT_COOKIE_DAYS);
         });


JustinPinner:
Thanks @Ap0c. Looks reasonable to me but happy for others to comment prior to merge. Otherwise 👍

Ap0c:
The errors have gone in IE10 and Safari 8, and I was able to sign up supporters in the UK using Stripe and the US using PayPal. Also got to see the shiny new checkout flow from #1631!

ie-working

PR #1636, opened 2017-05-12
add cloudwatch metrics to track membership signup attempts and third-party payment failures

michaelwmcnamara:
## Why are you doing this?
The aim of this change is to add monitoring to the payments process so that we can track the ratio of payment attempts to success/failures and potentially alert if the ratio exceeds a given threshold

## Changes
Add cloudwatch metric for attempted signup to any paid membership tier
Add cloudwatch metric for payment failure returned from stripe during signup for a paid membership tier
Add cloudwatch metric for payment failure returned from paypal during signup for a paid membership tier

## Testing
Simulated payments in dev and confirmed that payment attempt was being tracked in cloudwatch
Ran all tests including integration tests - confirmed tests passed.

## Screenshots
image


jacobwinch:
Looks good to me (once conflicts are resolved) 👍

PR #1640, opened 2017-05-16
Refresh events archive on an hourly basis

jacobwinch:
## Why are you doing this?
We're seeing 1000s of timeouts from Eventbrite every day. There is some Fastly tuning in progress which should improve matters, but (in my opinion) a major part of the problem is simply that we're making 1000s of unnecessary requests.

If I understand it correctly, the archivedEvents tasks are used to populate /events/archive. Presumably this only actually changes a few times per day, when the previous day's events are marked as ended (perhaps this only actually happens once per day?).

The archiveEvents tasks retrieve all of the events with a status of 'ended' by looping through pages from Eventbrite. This is currently 34 GET requests for Guardian Live Events and 10 GET requests for Masterclasses (so 44 requests combined). We run these tasks every minute (https://github.com/guardian/membership-frontend/blob/master/frontend/conf/application.conf#L22); that's 63K requests per day per instance.

I've made a config change in Fastly which means that we allow a higher max-age for urls which lookup ended events (and I've also configured stale-while-revalidate for these requests). As the chance of this task timing out is now significantly lower, I think we can safely schedule this task on an hourly basis.

## Trello card:
N/A

## Changes
* Run archived events tasks once per hour, instead of once per minute. This means we will make 1056 requests per instance, per day (which is probably still more than we need to, but a lot better than 63K).

## Testing
Confirmed that the archive still loads.

## Screenshots
N/A

### DIFF
--- a/frontend/app/services/EventbriteService.scala
+++ b/frontend/app/services/EventbriteService.scala
@@ -65,7 +65,7 @@ trait EventbriteService extends WebServiceHelper[EBObject, EBError] {
   // goal here is to keep the downstream cache warm (AWS CloudFront - ttl 61s).
   lazy val eventsTask = eventsTaskFor("live", 1.second, Config.eventbriteRefreshTime.seconds)
 
-  lazy val archivedEventsTask = eventsTaskFor("ended", 29.seconds, Config.eventbriteRefreshTime.seconds)
+  lazy val archivedEventsTask = eventsTaskFor("ended", 29.seconds, 3600.seconds)
 
   lazy val draftEventsTask =  eventsTaskFor("draft", 59.seconds, Config.eventbriteRefreshTime.seconds)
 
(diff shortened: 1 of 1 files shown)


jacobwinch:
cc @paulbrown1982 @michaelwmcnamara @rtyley

paulbrown1982:
👍 providing this change and the Fastly change doesn't affect the freshness or presence of Live Events and Masterclasses.

jacobwinch:
@paulbrown1982 it shouldn't - we use separate tasks for Live, Draft and Ended events.

paulbrown1982:
Fab, I was getting at whether the 2hr caching in Fastly only covered the archive listings. I've just seen the ```req.url ~ "status=ended"``` VCL condition and am satisfied 😄 👍

PR #1644, opened 2017-05-17
tell zuora subscribe command to use None title

johnduffell:
for subscriptions, to fulfil from zuora, we need to save the user's title.  To do this I need to amend the zuora subscribe command writer to optionally write the title.  However in membership, we don't ask for a title at all.

To allow this difference I could create a new type which has the title too for subs only, and subclass appropriately, but I thought for now since it's only one field I can just None it out in membership.  If it became a second field I'd start to use another approach.

See https://github.com/guardian/membership-common/pull/483 for the change in membership common.

@guardian/membership-and-subscriptions opinions welcome

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -9,7 +9,7 @@ object Dependencies {
   val sentryRavenLogback = "com.getsentry.raven" % "raven-logback" % "8.0.2"
   val scalaUri = "com.netaporter" %% "scala-uri" % "0.4.16"
   val memsubCommonPlayAuth = "com.gu" %% "memsub-common-play-auth" % "0.9" // v0.9 is the latest version published for Play 2.4...
-  val membershipCommon = "com.gu" %% "membership-common" % "0.408"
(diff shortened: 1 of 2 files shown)


project/Dependencies.scala
```diff
@@ -9,7 +9,7 @@ object Dependencies {
   val sentryRavenLogback = "com.getsentry.raven" % "raven-logback" % "8.0.2"
   val scalaUri = "com.netaporter" %% "scala-uri" % "0.4.16"
   val memsubCommonPlayAuth = "com.gu" %% "memsub-common-play-auth" % "0.9" // v0.9 is the latest version published for Play 2.4...
-  val membershipCommon = "com.gu" %% "membership-common" % "0.408"
+  val membershipCommon = "com.gu" %% "membership-common" % "0.411-SNAPSHOT"
```
johnduffell: https://github.com/guardian/membership-common/pull/483

jacobwinch:
Sounds fine to me 👍 (pending membership-common release)

PR #1646, opened 2017-05-18
Make metric namespacing consistent

jacobwinch:
## Why are you doing this?
CloudWatch metrics for this project are going to the namespace: "membership". Except for metrics related to Salesforce, which go to "Frontend"... I'm a pedant and this annoys me.

cc @michaelwmcnamara 

## Trello card: 
N/A

## Changes
* Fix namespace for Salesforce CloudWatch metrics


michaelwmcnamara:
Recommend checking there is not existing reporting that is reliant on these metrics.
Otherwise approved :+

jacobwinch:
Good point Mike - I checked it out and there were a few alarms firing off metrics configured in the Frontend namespace. I'll recreate the useful ones after merging this change (I made an executive decision and deleted some which haven't received any data since 2014!)

PR #1649, opened 2017-05-18
Commit with an old base to test Sentry Releases

rtyley:
## Why are you doing this?

https://github.com/guardian/prout/issues/50 - trying again, this time with the correct version of Prout deployed.


rtyley:
Failure was due to the PayPal test getting shunted into the old 'creepy' version:

0112screenshot

jacobwinch:
That's a very interesting catch... @Ap0c could this be related to the PayPal problem you were investigating?

Ap0c:
> @Ap0c could this be related to the PayPal problem you were investigating?

It's possible that it could coincide. However, this is a known bug that doesn't technically break anything, but does cause poor UX 😥.

PR #1650, opened 2017-05-19
Use latest membership common to pull in improved Scheduled Task

jacobwinch:
## Why are you doing this?
In order to pull in the latest membership-common (https://github.com/guardian/membership-common/pull/485), which reduces the number of blocking calls that we make on start-up.

Although deploys have stabilised recently, we're still seeing spikes in Sentry on most deploys:
https://sentry.io/the-guardian/membership/issues/193188674/

I'm hoping this change will help with that.

## Changes
* Use latest version of membership-common
* Remove timeouts as ScheduledTask no longer requires them


jacobwinch:
cc @johnduffell

johnduffell:
excellent! 🚀 👍

PR #1652, opened 2017-05-22
Pass through Referer Url and Referer Pageview Id

jranks123:
## Why are you doing this?

Currently, for contributions, we pass through the referer pageview id and referer url through to the datalake and use it to anaylse the content that drives acquisition. By only having this view on contributions, our current analysis is skewed towards contributions and doesn't tell the whole story.

This PR is the first step towards being able to see this data in the datalake. 

Would love to get @rtyley 's opinion :)


frontend/app/controllers/Info.scala
```diff
@@ -21,6 +21,9 @@ import scala.concurrent.Future
 trait Info extends Controller {
   def supporterRedirect(countryGroup: Option[CountryGroup]) = NoCacheAction { implicit request =>
     val determinedCountryGroup = (countryGroup orElse request.getFastlyCountryCode).getOrElse(CountryGroup.RestOfTheWorld)
+    val referrerUrl = request.headers.get("referer")
```
dominickendrick: This is not used

frontend/app/controllers/Info.scala
```diff
@@ -21,6 +21,9 @@ import scala.concurrent.Future
...
+    println("the referrerUrl: " + referrerUrl)
```
dominickendrick: Please remove

frontend/assets/javascripts/src/modules/analytics/campaignCode.es6
```diff
@@ -15,7 +15,7 @@ function recordCampaign () {
 	// Don't want JS to break if there is a problem reading query params.
 	// Just don't set the cookie if this happens.
 	try {
-
+gz
```
dominickendrick: Please remove this

frontend/app/controllers/TierController.scala
```diff
@@ -199,7 +199,7 @@ object TierController extends Controller with ActivityTracking
       }
 
       def doUpgrade(): Future[Result] = {
-        val upgrade = memberService.upgradePaidSubscription(paidMember, target, form, CampaignCode.fromRequest)
+        val upgrade = memberService.upgradePaidSubscription(paidMember, target, form, CampaignCode.fromRequest, RefererUrl.fromRequest, RefererPageviewId.fromRequest)
```
rtyley: I'd like to avoid adding multiple parameters that have to be handed down in all call invocations - could these fields be encapsulated in a single content case class? Should probably have all 3 of these in it:

* CampaignCode.fromRequest
* RefererUrl.fromRequest
* RefererPageviewId.fromRequest

frontend/app/controllers/Info.scala
```diff
@@ -142,6 +145,7 @@ trait Info extends Controller {
         description = Some(CopyConfig.copyDescriptionSupporters)
       ),
       detailImageOrientated))
+      .withCookies(Cookie("gu_mem_ref_url", refUrl.getOrElse("")), Cookie("gu_refpvid", refPvid.getOrElse("")))
```
rtyley: Could you put `"gu_mem_ref_url"` and `"gu_refpvid" in string constants, as they're used elsewhere?

Separate thing- these two bits of code are equivalent, and I think the second one is a bit prettier, so I often use it in preference:

* `refUrl.getOrElse("")`
* `refUrl.mkString`

frontend/app/utils/TrackingDataVariables.scala
```diff
@@ -0,0 +1,37 @@
...
+object RefererUrl {
+  def fromRequest(implicit request: RequestHeader): Option[RefererUrl] =
+    for {
+      referer <- request.cookies.get("gu_mem_ref_url")
+    } yield {
+      RefererUrl(referer.value)
+    }
+}
+
+case class RefererPageviewId(get: String)
```
rtyley: Replace these 3 case classes (`CampaignCode`, `RefererUrl`, `RefererPageviewId`) with a single `Referral` one!

frontend/app/utils/ReferralData.scala
```diff
@@ -0,0 +1,30 @@
...
+case class ReferralData(campaignCode: Option[String], url: Option[String], pageviewId: Option[String])
+
+object ReferralData {
+
+  val CampaignCodeKey = "mem_campaign_code"
+  val UrlKey = "gu_mem_ref_url"
+  val PageviewIdKey = "gu_refpvid"
+
+  def makeCookies(implicit request: RequestHeader): Seq[Cookie] = {
+    val refUrl = request.headers.get("referer").map(Cookie(ReferralData.UrlKey, _))
```
johnduffell: this is causing the 500 error pagerduty alerts as not everything that can be in a referer header is valid in a cookie value

johnduffell: also according to https://github.com/guardian/membership-frontend/blob/master/frontend/app/controllers/Info.scala#L93 this is a Cached [for 60 seconds](https://github.com/guardian/membership-frontend/blob/master/frontend/app/controllers/Cached.scala#L15) method, and we can't set a cookie based on the referrer in a cached method

PR #1651, opened 2017-05-19
Restyle australia supporter page

joelochlann:
## Why are you doing this?
Same reasoning as #1616 (which erroneously says "except US" - should be "except US & Australia", see referenced PRs for the full story)

## Changes
- Adds an A/B test to test two variants of the Australian supporter landing page
- This means I had to make the Australian supporter page uncached. @rtyley I'd appreciate advice on any adverse consequences this may have!
- New variant is based on the UK/Canada/EU/ROW design, with a few tweaks which came from the US office (top video holding image changed, second video replaced by an image, bottom section is now quote from the editor)

## Old mobile
screencapture-membership-theguardian-au-supporter-1495209775678

## New mobile
screencapture-localhost-9100-au-supporter-1495210187551

## Old tablet
screencapture-membership-theguardian-au-supporter-1495209812198

## New tablet
screencapture-localhost-9100-au-supporter-1495210214341

## Old desktop
screencapture-membership-theguardian-au-supporter-1495209823401

## New desktop
screencapture-localhost-9100-au-supporter-1495210225231

cc @guardian/contributions

PR #1653, opened 2017-05-23
Include country in signup logging

joelochlann:
In #1651 I forgot to log the country on successful signup. Since I'm trying to measure the effect of the new Australia supporter page design on supporter signup, I need to know which ones are actually Australian signups! Otherwise I'll be counting signups in other countries who were placed in a test bucket but probably never saw the new page

I also changed my impression logging so that I know which impressions occurred after this change

@rtyley


joelochlann:
Following discussion with @rtyley, am ditching in favour of #1656 which may do the trick

PR #1655, opened 2017-05-23
some updates to cloudwatch metrics

michaelwmcnamara:
## Why are you doing this?
Moving attempted signup metric closer to start of make-member function so that we are tracking as much of this process as possible
Adding extra cloudwatch metrics
A) because we want to be able to track the cause of signup failures
B) in the case of the tracking added to the "thank-you" function, there are a lot of operations in this function and current tracking would not detect if signup failed at this stage

## Changes
Moved attempted signup metric closer to start of make-member function;
added tracking for signup failures due to Gateway errors; 
added tracking to show when user has reached the end of the "thank-you" funtion;

### DIFF
--- a/frontend/app/controllers/Joiner.scala
+++ b/frontend/app/controllers/Joiner.scala
@@ -292,8 +292,7 @@ object Joiner extends Controller with ActivityTracking with PaymentGatewayErrorH
         case t: Tier if !upgrade => salesforceService.metrics.putThankYou(tier)
         case _ =>
       }
-      Ok({salesforceService.metrics.putThankYou(tier)
-        views.html.joiner.thankyou(
+      Ok({views.html.joiner.thankyou(
         request.subscriber,
         paymentSummary,
         paymentMethod,
--- a/frontend/app/controllers/Joiner.scala
+++ b/frontend/app/controllers/Joiner.scala
@@ -123,7 +123,6 @@ object Joiner extends Controller with ActivityTracking with PaymentGatewayErrorH
   def enterPaidDetails(
     tier: PaidTier,
     countryGroup: CountryGroup) = OptionallyAuthenticatedNonMemberAction(tier).async { implicit request =>
-
     val userOpt = authenticatedIdUserProvider(request)
     implicit val resolution: TouchpointBackend.Resolution =
       TouchpointBackend.forRequest(PreSigninTestCookie, request.cookies)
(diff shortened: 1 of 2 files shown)


frontend/app/monitoring/MemberMetrics.scala
```diff
@@ -32,12 +36,12 @@ class MemberMetrics(val backendEnv: String) extends TouchpointBackendMetrics {
     put(s"failed-sign-up-${tier.name}")
   }
 
-  def putFailSignUpStripe(tier: Tier) {
-    put(s"failed-sign-up-stripe-${tier.name}")
+  def putFailSignUpGatewayError(tier: Tier) {
+    put(s"failed-sign-up-paypal-${tier.name}")
```
jacobwinch: Need to change this metric name to match the new method name

frontend/app/controllers/Joiner.scala
```diff
@@ -293,15 +290,16 @@ object Joiner extends Controller with ActivityTracking with PaymentGatewayErrorH
         case t: Tier.Supporter if !upgrade => MembersDataAPI.Service.removeBehaviour(request.user)
         case _ =>
       }
-      Ok(views.html.joiner.thankyou(
+      Ok({salesforceService.metrics.putThankYou(tier)
```
jacobwinch: Have just realised that upgrades will also call this method (https://github.com/guardian/membership-frontend/blob/master/frontend/app/controllers/TierController.scala#L235), so in this form we are going to be counting new acquisitions and upgrades under a single metric.

jacobwinch: Maybe we can capture what we want by extending the pattern match above...

frontend/app/controllers/Joiner.scala
```diff
@@ -290,18 +287,21 @@ object Joiner extends Controller with ActivityTracking with PaymentGatewayErrorH
...
     } yield {
       tier match {
-        case t: Tier.Supporter if !upgrade => MembersDataAPI.Service.removeBehaviour(request.user)
+        case t: Tier.Supporter if !upgrade => {salesforceService.metrics.putThankYou(tier)
+          MembersDataAPI.Service.removeBehaviour(request.user)}
+        case t: Tier if !upgrade => salesforceService.metrics.putThankYou(tier)
         case _ =>
       }
-      Ok(views.html.joiner.thankyou(
+      Ok({salesforceService.metrics.putThankYou(tier)
```
jacobwinch: Need to remove this one now or we'll be double-counting!

PR #1656, opened 2017-05-23
Add canRun condition to supporter landing page test

joelochlann:
When I log a successful signup with A/B test participation data, I want to know that this person actually saw the thing relevant to the A/B test.

It seemed the best way to do this was to use the `canRun` function so the cookie only gets dropped if the user visits (say) the Australian supporter landing page.

But this messes with the behaviour of `describeParticipation` which we use for logging - when logging on the signup page it will assume they aren't in the test because `canRun` isn't true on that page.

So I added `describeParticipationFromCookie` which takes the cookie as the source of truth for test participation. This function would not be suitable for use on a request where a test cookie may or may not be dropped, but is it suitable for requests which we know are not relevant for A/B tests - and the post-join request certainly seems like one of these.

What do you think @rtyley?

### DIFF
--- a/frontend/app/abtests/ABTest.scala
+++ b/frontend/app/abtests/ABTest.scala
@@ -25,8 +25,12 @@ abstract class ABTest(val slug: String, val audience: AudienceRange, canRun: Req
   def cookieFor(variant: BaseVariant) =
     Cookie(cookieName, variant.slug, maxAge = Some(ABTest.CookieAge), httpOnly = false)
 
-  def describeParticipation(implicit request: RequestHeader): String =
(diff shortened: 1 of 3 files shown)


joelochlann:
Yeah, I did think about that but I feel like it could actually be more confusing - at least this way `describeParticipation` really does give you the participation the browser has or will have by the end of the request, no ifs or buts. While `describeParticipationFromCookie` is the participation as determined by the cookie.

Whereas it was really hard for me to find function names that encapsulated the difference between `allocate` with `canRun` and without, which seemed to indicate that it wasn't a good distinction to make

PR #1592, opened 2017-04-13
Add thank-you page survey asking users what prompted their decision to join

rtyley:
## Why are you doing this?

This comes at the request of Amanda on the Acquisition team, who says they've seen great results after doing this on Contributions: https://github.com/guardian/contributions-frontend/pull/248 - they've gathered some really great stories.

## Changes

Adds a _"Before you go, we would love to know a little bit more about your decision to support us. We have a few questions that should take only a couple of minutes to complete."_ message with a CTA leading to https://www.surveymonkey.co.uk/r/6PM3PZ3 on the thank-you page.

Text and design comes from Mario Andrade.

## Screenshots

image

cc @dominickendrick

PR #1662, opened 2017-06-07
Reinstate testimonials survey

joelochlann:
Added  in #1592 
Removed in #1657 

Now it's back again! With a new survey link. This is to collect potentially election-related testimonials over the next few days

@rtyley


joelochlann:
Perhaps we should create an admin UI for turning this thing on & off...

PR #1664, opened 2017-06-08
Fix "Already have a Guardian account? Sign in" message on checkout for mobile-app users

rtyley:
## Why are you doing this?

This changes the approach used to populate username in the _"You're signed in as Jess Blogs"_ message from client-side to server-side, as the checkout page (https://membership.theguardian.com/join/supporter/enter-details) already does an Identity API request to get all user details before rendering.

The old client-side approach depended on the `GU_U` cookie to get the display-name, which has so far not been available to mobile-apps. The apps mainly just use an access token, not cookies, but have been able to exchange that access token for a short-lived `SC_GU_U` cookie (*not* the `GU_U` cookie which contains the display-name) since https://github.com/guardian/identity/pull/553. Because the `GU_U` cookie wasn't present, the javascript incorrectly displayed the _"Already have a Guardian account? Sign in"_ message, even though the user _was_ signed in.

In the past day or so there was work done with https://github.com/guardian/identity/pull/663 to also give a `GU_U` cookie in return for the token, but that was reverted with https://github.com/guardian/identity/pull/664 when we realised doing the fix server-side in membership-frontend was a better approach.

#### Before

https://membership.theguardian.com/join/supporter/enter-details - logged in, but with no `GU_U` cookie:
image
...note that the page is asking you sign in, even though you're already signed in.

#### After
https://mem.thegulocal.com/join/supporter/enter-details - logged in, but with no `GU_U` cookie:
image
...correctly the _"You're signed in as sdfsdfsdf sdfsdfsdf. Sign out"_ message is displayed.

Other pages after this fix:

image
image
image

cc @fgauchet @Leyths @markjamesbutler


rtyley:
All looks good on PROD (signed-in with no `GU_U` cookie):

image
image
image
image

PR #1666, opened 2017-06-08
add logging for thank you success

michaelwmcnamara:
## Why are you doing this?
nvestigating why cloudwatch metrics show a gap between successful member creation and sending of thank you page.
Extra logging needed to target these events so I can be sure I am looking at the correct examples

## Changes
* Add logging to Joiner.scala to track when thank you method has been successful for new memberships

### DIFF
--- a/frontend/app/controllers/Joiner.scala
+++ b/frontend/app/controllers/Joiner.scala
@@ -294,8 +294,10 @@ object Joiner extends Controller with ActivityTracking with PaymentGatewayErrorH
     } yield {
       tier match {
         case t: Tier.Supporter if !upgrade => {salesforceService.metrics.putThankYou(tier)
+          logger.info(s"thank you page displayed ${tier.name}")
           MembersDataAPI.Service.removeBehaviour(request.user)}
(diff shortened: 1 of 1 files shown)


frontend/app/controllers/Joiner.scala
```diff
@@ -294,8 +294,10 @@ object Joiner extends Controller with ActivityTracking with PaymentGatewayErrorH
     } yield {
       tier match {
         case t: Tier.Supporter if !upgrade => {salesforceService.metrics.putThankYou(tier)
+          logger.info(s"thank you page displayed ${tier.name}")
           MembersDataAPI.Service.removeBehaviour(request.user)}
-        case t: Tier if !upgrade => salesforceService.metrics.putThankYou(tier)
+        case t: Tier if !upgrade => {salesforceService.metrics.putThankYou(tier)
+          logger.info(s"thank you page displayed ${tier.name}")}
         case _ =>
       }
```
jacobwinch: Can we move the logging outside the pattern match to avoid repetition? It's also worth logging the identityId so we can identify specific users affected by this issue.

PR #1667, opened 2017-06-09
Updating the maintenance schedule to cover the Salesforce Summer '17 release

paulbrown1982:
Updating the maintenance schedule to cover the Salesforce Summer '17 release this evening: https://status.salesforce.com/status/maintenances/22710
https://releasenotes.docs.salesforce.com/en-us/summer17/release-notes/salesforce_release_notes.htm

## Why are you doing this?
Better experience for our readers rather than a failing checkout flow.

## Changes
* Updating the maintenance schedule to cover the Salesforce Summer '17 release

## Screenshot
picture 114
 
cc @rtyley @dominickendrick @johnduffell @jacobwinch @michaelwmcnamara

### DIFF
--- a/frontend/app/controllers/Outages.scala
+++ b/frontend/app/controllers/Outages.scala
@@ -3,7 +3,7 @@ package controllers
 import play.api.mvc.Controller
 
 object Outages extends Controller {
-  def maintenanceMessage = CachedAction {
+  def maintenanceMessage = NoCacheAction {
     Ok(views.html.info.maintenanceMessage())
   }
 
--- a/frontend/app/utils/PlannedOutage.scala
+++ b/frontend/app/utils/PlannedOutage.scala
@@ -19,9 +19,19 @@ object PlannedOutage {
     new DateTime(2016,11, 6, 4,45, UTC)
   )
 
+  /*
+  Salesforce is performing their Summer '17 Major Release on Friday 9 June 2017 at 10pm BST.
+  The maintenance window is expected to last 5 minutes. Typical releases like this usually result in one minute's downtime.
+   */
+  val salesforceUpgrade = new Interval(
+    new DateTime(2017, 6, 9, 21, 0, UTC),
+    new DateTime(2017, 6, 9, 21, 5, UTC)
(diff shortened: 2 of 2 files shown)


jacobwinch:
Sounds sensible to me 👍

johnduffell:
good spot 👍 
as a future enhancement maybe we can use the zuora and salesforce health check to display the message

frontend/app/controllers/Outages.scala
```diff
@@ -3,7 +3,7 @@ package controllers
 import play.api.mvc.Controller
 
 object Outages extends Controller {
-  def maintenanceMessage = CachedAction {
+  def maintenanceMessage = NoCacheAction {
```
johnduffell: actually why is this changing to no cache?  the actual url is a [temporary redirect](https://github.com/guardian/membership-frontend/blob/395d9fdcbe54cdf670edac075ac0a06a833e0570/frontend/app/actions/Fallbacks.scala#L14) and it's that that shouldn't be cached.

paulbrown1982: Ah good point, this page contains static content and is on a static public URL. The dynamic-ness comes from the redirect. I'll remove that change.

frontend/app/utils/PlannedOutage.scala
```diff
@@ -19,9 +19,19 @@ object PlannedOutage {
...
   )
 
+  /*
+  Salesforce is performing their Summer '17 Major Release on Friday 9 June 2017 at 10pm BST.
+  The maintenance window is expected to last 5 minutes. Typical releases like this usually result in one minute's downtime.
+   */
+  val salesforceUpgrade = new Interval(
+    new DateTime(2017, 6, 9, 21, 0, UTC),
+    new DateTime(2017, 6, 9, 21, 5, UTC)
+  )
```
rtyley: Minor thing, but Salesforce are giving their outage time in BST, so you could consider configuring using the same timezone?

```
val BST = DateTimeZone.forID("Europe/London")
```

PR #1668, opened 2017-06-09
Fix acceptance test broken with 

rtyley:
## Why are you doing this?

I removed the `js-user-displayname` css class in PR #1664 because I thought it was only used by the display-or-hide-username JS, but it turns out it was used for the QA tests too :)

https://travis-ci.org/guardian/membership-frontend/builds/241203618#L282


jacobwinch:
👍

PR #1675, opened 2017-06-14
Record CMP for supporters

jranks123:
## Why are you doing this?

Currently, any campaign code that is passed through as CMP instead of INTCMP, doesn't get recorded. This is often used for email campaigns. This PR makes sure that we check for CMP codes when there isn't an INTCMP provided. 

## Changes
* Change 1
* Change 2

### DIFF
--- a/frontend/assets/javascripts/src/modules/analytics/campaignCode.es6
+++ b/frontend/assets/javascripts/src/modules/analytics/campaignCode.es6
@@ -25,7 +25,7 @@ function recordCampaign () {
 	} catch (ex) {
 		console.log('Failed to retrieve campaign code:', ex.message);
 	} finally {
-        var campaignCode = intcmp || cmp;
+        let campaignCode = intcmp || cmp;
 		if (campaignCode) {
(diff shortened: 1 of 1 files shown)


frontend/assets/javascripts/src/modules/analytics/campaignCode.es6
```diff
@@ -10,19 +10,22 @@ import URLSearchParams from 'URLSearchParams';
...
-		campaignCode = urlParams.get('INTCMP');
+		intcmp = urlParams.get('INTCMP');
+        cmp = urlParams.get('CMP');
+
 
 	} catch (ex) {
 		console.log('Failed to retrieve campaign code:', ex.message);
 	} finally {
-
+        var campaignCode = intcmp || cmp;
```
Ap0c: `const` or `let`

johnduffell:
:skull_and_crossbones: close or ship?

jranks123:
Ah, I missed this as I was on holiday. It's for merge!

PR #1679, opened 2017-06-20
Fix uknown typo and sneakily bumps membership-common again

AWare:
## Why are you doing this?
Fixed a typo in membership-common, so fixing it everywhere.
## Changes
* uknown => unknown
* membership-common => https://github.com/guardian/membership-common/pull/495  https://github.com/guardian/membership-common/pull/500

PR #1684, opened 2017-06-27
Fix Stripe Error Option Type

Ap0c:
## Why are you doing this?

[Changes](https://github.com/guardian/membership-common/pull/493) in membership-common converted Stripe errors from a `String` to an `Option[String]`. The code in `Joiner.scala` is expecting a `String`, and thus bumping membership-common resulted in failing builds, as in #1679.

This changes `setBehaviourNote` to expect an `Option[String]` instead, and tweaks the calls in `Joiner.scala` to reflect that.

**Note:** I've bumped membership-common to the latest version here. Running through the checkout flow seems to work ok, but if anyone can think of any recent changes that might cause issues please let me know (@rupertbates @AWare @paulbrown1982 @alexduf).

## Changes

- Bump membership-common to 0.425.
- Switch to using an `Option[String]` instead of a `String` for error messages.

### DIFF
--- a/frontend/app/controllers/Joiner.scala
+++ b/frontend/app/controllers/Joiner.scala
@@ -265,13 +265,13 @@ object Joiner extends Controller with ActivityTracking with PaymentGatewayErrorH
 
         case error: PaymentGatewayError =>
           salesforceService.metrics.putFailSignUpGatewayError(tier)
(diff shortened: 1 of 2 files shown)


AWare:
🍾

Ap0c:
Done the standard check, Supporter sign-ups seem ok.

PR #1687, opened 2017-06-30
Add social-signin buttons to merged-checkout

rtyley:
## Why are you doing this?

The initial Merged-Registration test [indicated that merging registration damaged conversion](https://github.com/guardian/membership-frontend/pull/1654), and we've made various improvements in the hope that we can get a positive result instead! The final change before turning on a new A/B test is to add the social-sign-in buttons which were missing from the previous merged-registration experience - we found that about a third of users who come to our normal checkout page have used social-sign-in to login.

These buttons won't be visible to the general public yet, as there's no active AB test running, but you'll be able to review them by turning on the merged-registration feature:

The oauth links are hard-coded to production, as I haven't had time to get the [Identity OAuth application](https://github.com/guardian/identity-federation-api) running in dev- we'll test them in prod!

## Screenshots

image
image

If anyone thinks they can do a better layout for the social-sign-in buttons, please go ahead, this was the best I could do with my limited css layout abilities.

### DIFF
--- a/frontend/app/configuration/Config.scala
+++ b/frontend/app/configuration/Config.scala
@@ -44,6 +44,12 @@ object Config {
 
   private val idSkipConfirmation: (String, String) = "skipConfirmation" -> "true"
 
+  def googleSigninUrl = oauthWebAppSigninUrl("google")(_)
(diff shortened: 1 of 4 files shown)


rtyley:
Ok, will merge and we can modify design if necessary in another PR.

PR #1688, opened 2017-07-04
Make Monthly Contributions More Yellow

Ap0c:
## Why are you doing this?

There is a concern that the current checkout page looks a bit broken, which could cause drop-offs where payments are concerned. This makes the whole monthly contributions checkout page yellow, instead of just the box in the middle. It brings it more into line with the equivalent page in support frontend [here](https://github.com/guardian/support-frontend/pull/83), which will eventually supersede the page in membership. It will also keep @superfrank happy 🐧.

## Changes

- Made the area surrounding the checkout form yellow.
- Added a hacky after pseudo-element, to make the previously white section further down the page also appear yellow.

## Screenshots

**Before:**

old-checkout

**After:**

new-checkout

### DIFF
--- a/frontend/assets/stylesheets/components/_payment-monthly-contribution.scss
+++ b/frontend/assets/stylesheets/components/_payment-monthly-contribution.scss
@@ -1,5 +1,13 @@
 
 .skin-monthly-contribution {
+    background-color: guss-colour(multimedia-main-2);
+
+    &:after {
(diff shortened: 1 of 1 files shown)


superfrank:
chris

Ap0c:
Thanks guys...

Ap0c:
Looks ok.

PR #1694, opened 2017-07-11
Kibana

AWare:
## Why are you doing this?
So membership-frontend logs end up in Kibana. :+1:

## Changes
* Membership Frontend Logs in Kibana
* Membership Frontend Permissions to Use Kinesis

@johnduffell

PR #1696, opened 2017-07-12
Kibana Attempt 2

AWare:
Reverts guardian/membership-frontend#1695

Attempting to make this right.

**Changes since Reverts guardian/membership-frontend#1694**
Config file on server already has kibana details in, so no need to use user data.

### DIFF
--- a/cloud-formation/membership-app.cf.yaml
+++ b/cloud-formation/membership-app.cf.yaml
@@ -50,6 +50,12 @@ Parameters:
     Description: Security group that grants access to the account's Vulnerability
       Scanner
     Type: AWS::EC2::SecurityGroup::Id
+  LoggingKinesisStream:
+    Description: Kinesis stream id to send logging to e.g. elk-stack-ElkKinesisStream-12345
+    Type: String
+  LoggingPolicy:
+    Description: Policy needed to access the kinesis stream
+    Type: String
 Mappings:
   StageVariables:
     PROD:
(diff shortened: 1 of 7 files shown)


cloud-formation/membership-app.cf.yaml
```diff
@@ -107,6 +113,13 @@ Resources:
             aws --region ${AWS::Region} s3 cp s3://membership-dist/${Stack}/${Stage}/frontend/frontend_1.0-SNAPSHOT_all.deb /tmp
             mkdir /etc/gu
             aws --region ${AWS::Region} s3 cp s3://membership-private/${Stage}/membership.private.conf /etc/gu
+             cat <<EOF >>/etc/gu/subscriptions-frontend.private.conf
```
jacobwinch: Should this be membership.private.conf?

jacobwinch:
What's the difference between this PR and the original attempt?

johnduffell:
@jacobwinch the changes to the config are no longer happening in the userdata in the cloudformation courtesy of @davidfurey suggestion this morning.  They are in the S3 config.

johnduffell:
@pvighi has created a code environment so it would be worth running it there first.

jacobwinch:
Thanks @johnduffell - this happened after I asked the original question, which is why I was confused!

pvighi:
@johnduffell I think this is the one project I didn't create a CODE env for since the template was already in yaml and updating with amigo so I didn't need to change anything.

PR #1697, opened 2017-07-14
Drop requirement for user to be a member for Masterclass purchases

rtyley:
## Why are you doing this?

Further to https://github.com/guardian/membership-frontend/pull/1680 ...Kirsty Buck has been reporting further problems around SEO - users being directed to our pages, either tho' they force membership as a purchase requirement (which masterclasses don't want). This change drops the requirement for users _to be a member_ for Masterclass purchases - it won't force them to log in, or be a member when they click on the event's buy button - the Member will only get a discount if they are logged in as a member.

### DIFF
--- a/frontend/app/controllers/Event.scala
+++ b/frontend/app/controllers/Event.scala
@@ -1,5 +1,6 @@
 package controllers
 
+import _root_.services.{EventbriteService, GuardianLiveEventService, MasterclassEventService}
 import actions.ActionRefiners._
 import actions.Fallbacks._
 import actions.{Subscriber, SubscriptionRequest}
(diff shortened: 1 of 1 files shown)


Ap0c:
To clarify, non-members on Masterclasses get sent to Eventbrite for standard price tickets. Members on Masterclasses will get sent through with a discount code in the query param?

Live events are unaffected, they still require membership?

rtyley:
> To clarify, non-members on Masterclasses get sent to Eventbrite for standard price tickets.

yep- users who are not signed in as a member will get the standard price tickets.

> Members on Masterclasses will get sent through with a discount code in the query param?

yep- users signed in as members will get the appropriate discount code


> Live events are unaffected, they still require membership?

yep- no change to Live events, the user must still be signed in as a member to get through to the eventbrite page, and the user will be redirected to a become-a-member page if they aren't signed-in as a
a member.

rtyley:
> Testing in PROD failed!

The failure occurred while facing the PayPal frame ([screencast](https://saucelabs.com/beta/tests/c94e16b2b5324b199feafa35ca8e0822/watch#536)), showing the 'Agree and Pay' screen:

image

...the test had already passed for Stripe payment. The 'Agree and Pay' screen visually looks very similar to how it was in [the previous test run](https://saucelabs.com/beta/tests/5f914f77ee7141aaa33e705668160d17/watch#573), but the test run reported that it could not find a require element id on the page.

Having re-run the travis build, the build is now passing: https://travis-ci.org/guardian/membership-frontend/builds/254795702 - so I'm going to chalk this up to the flakey PayPal experience :(

The changes I introduced certainly shouldn't have impacted the PayPal experience, being restricted to just impacting events.

PR #1699, opened 2017-07-21
JVM metrics in librato (plus CODE env)

johnduffell:
# Why are you doing this?

## problem

We have some problems with EC2 instances recycling regularlyish or dropping out of health temporarily.  This I suspect is due to garbage collection as we're using the GC that has a large heap and stops the world when it is cleaning it all up.  This is the same problem I saw on dotcom and changing to G1 collector solved that issue.

I want to gather some GC stats to prove that is the problem before I actually make the change.
## original apprach
I originally stole some code from dotcom to put the metrics into cloudwatch on a timer by calling the JMX api.  Needs a bit of final tidy up but I pushed it here https://github.com/guardian/membership-frontend/compare/jvm-cloudwatch

## liberato
it was suggested to use liberato as it has a standard collector that can collect all sorts of info about the boxes  without hassle, so we don't have to worry about coding it up.
Unfortunately it turned out to be a bit of a nightmare as their architecture has the JVM exposing a JMX interface over tcp, the collectd engine (separate install) needs to be running on the instance and grabbing data from the JVM (and other processes) and pushing that into librato over https.  The logging for collectd is very poor so if it doesn't work you have to guess, and I'm struggling to get librato to show the data nicely.  I managed to use the "hostname" field in collectd to send tags to librato so I can tell which instance is pushing data, but it seems like it doesn't support more tags so there will be difficulty distinguishing between ASGs (in cloudwatch dimensions are similar to tags)

## or use kibana?
Another simple alternative is to push JSON to kibana which can also draw graphs and have things tagged, and the data is correlated with the logging.  But I feel like we should get the kibana server properly set up before we start getting too far with that.

## code env is available separately
I can separate out the PR (I should have done it anyway!) to have a CODE env without librato monitoring.

## 👋 
@paulbrown1982 @jacobwinch @michaelwmcnamara


cloud-formation/membership-app.cf.yaml
```diff
@@ -59,6 +59,18 @@ Mappings:
       DynamoDBTables:
```
johnduffell: all these changes in cloudwatch are purely for supporting a CODE env

cloud-formation/membership-app.cf.yaml
```diff
@@ -106,14 +124,12 @@ Resources:
...
-            dpkg -i /tmp/frontend_1.0-SNAPSHOT_all.deb
-            chown frontend /etc/gu/membership.private.conf
-            chmod 0600 /etc/gu/membership.private.conf
-            /opt/cloudwatch-logs/configure-logs application ${Stack} ${Stage} ${App} /var/log/frontend/frontend.log
+            export stack='${Stack}'
+            export app='${App}'
+            export stage='${Stage}'
+            export region='${AWS::Region}'
+            aws s3 cp --region ${AWS::Region} s3://membership-dist/${Stack}/${Stage}/frontend/init-instance.sh .
+            . init-instance.sh
```
johnduffell: all the stuff is moved to a script to prevent having to cloudform when we want to change the startup of instances

frontend/build-tc.sh
```diff
@@ -1,2 +1,5 @@
 #!/bin/bash
 echo ${BUILD_NUMBER:-DEV} > conf/build.txt
+# put the script needed on the machine into the package so it ends up on the instance via riffraff
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+echo "##teamcity[publishArtifacts '$DIR/instance => .']"
```
johnduffell: this magic println pushes the instance dir into the build artifacts dir, which riffraff eventually pushes onto the box so we can access it at startup

cloud-formation/membership-app.cf.yaml
jacobwinch: I know it's not been introduced by this PR, but I don't think membership-frontend actually needs any information about Promotions these days...

cloud-formation/membership-app.cf.yaml
jacobwinch: 👍

johnduffell:
closed in favour of #1702 but I'll reopen as a code env PR

PR #1700, opened 2017-07-21
update membership common to better report sf lookup errors

johnduffell:
## Why are you doing this?
We would like to have a better idea when things go wrong with salesforce lookup.  At the moment, regardless of what happens, memberhship common will return a None for lookup found no user or lookup had an error.  This was hindering our ability to find out why users were not getting their membership fulfilled by members-data-api.

As a result, I have updated membership-common to return an Either where the left is an error, and the right is an Option of whether the user existed.

This PR pulls those changes into membership.

@pvighi check it out!


johnduffell:
I'll merge this one first thing in the morning

PR #1692, opened 2017-07-07
Reinstate AB test for Merged Registration

rtyley:
## Why are you doing this?

This is a redo of the AB test created in PR #1631, but this time with social-sign-in icons as refined in PR #1687.

## Changes
I've tweaked the `canRun` for the AB test so that it only activates for non-signed-in users (because the experience does not vary for signed-in users). The AB test cookie will only get dropped for non-signed-in cookies, and will persist for those users. So this test is measuring the conversion of people who have either:

* *control* variant: always been forced to profile.theguardian.com/register
* *merged* variant : seen the merged-registration page at membership.theguardian.com/join/supporter/enter-details?ab.merged-social-registration=merged at least once

PR #1701, opened 2017-07-21
Remove Merged 'Social' Registration AB test

rtyley:
## Why are you doing this?

This test (introduced with PR #1692) has completed. The results are still in favour of the control variant: ie using a separate page for Identity registration:

https://docs.google.com/spreadsheets/d/1p6KknBtlbkAuJ3cqMOSr-i5e_TVPGy7OB0P7wo0oWWY/edit#gid=0


johnduffell:
it's always nice when an AB test proves a theory wrong!
although it does seem barmy, I assume you've consulted with some UX minds to work out if we're violating any other UX principles?
👍

rupertbates:
Really interesting, thanks for seeing this through Roberto. 💯

PR #1705, opened 2017-07-25
CODE environment for membership

johnduffell:
## Why are you doing this?
At the moment there is no way to test cloudformation/instance related changes.  This is find for most app development, but when you want to change something risky that means you have to go to prod before you can know if it worked, which leaves you with some messy cleanup if  it goes wrong, and can cause a prod outage easily.

To remove a lot of the risk, it makes sense to have a separate env where we can occasionally cloudform, similar to prod, even if the instances don't really do much apart from start and pass the healthcheck.

This PR introduces a basic code env which will have one instance running, we can deploy with riffraff and cloudform as we like.

I originally did this as part of getting librato/collectd up and running, but it turned out that was way more complicated than actually doing it in scala from the jvm.


frontend/build-tc.sh
```diff
@@ -1,2 +1,7 @@
 #!/bin/bash
 echo ${BUILD_NUMBER:-DEV} > conf/build.txt
+
+# put the script needed on the machine into the package so it ends up on the instance via riffraff
+# the special ##teamcity command causes teamcity to write it to the build output
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+echo "##teamcity[publishArtifacts '$DIR/instance => .']"
```
johnduffell: @lmath I wasn't sure the best update for the comment... maybe you can suggest a replacement?

lmath: This is good!

paulbrown1982: Wonderfully magic!

cloud-formation/membership-app.cf.yaml
```diff
@@ -87,6 +101,10 @@ Resources:
         Value:
           Ref: Stage
         PropagateAtLaunch: 'true'
+      - Key: Name
+        Value:
+          Fn::FindInMap: [ StageVariables, { Ref: Stage }, InstanceName ]
+        PropagateAtLaunch: 'true'
```
johnduffell: this is just so the instances get a name so we can more easily tell which is which

johnduffell:
so since the non cloudformation changes are just adding new scripts to the package, the new code should work with the old cloudformation, then once the files are there, we can change to the new cloudformation.  This reduces the risk of the deploy as it means rolling back involves less manual steps if anything goes wrong.
1. deploy code -> if fails roll back the code
1. deploy cloudformation -> if fails roll back the cloudformation only
1. add one instance to ASG -> if fails roll back the cloudformation and kill the specific instance
1. kill instance, then redeploy/recycle ASG -> if fails roll back cloudformation and kill new instances

frontend/build-tc.sh
```diff
@@ -1,2 +1,7 @@
 #!/bin/bash
 echo ${BUILD_NUMBER:-DEV} > conf/build.txt
+
+# put the script needed on the machine into the package so it ends up on the instance via riffraff
```
davidfurey: Maybe:
`put the init-instance.sh script into the riff-raff package so that it ends up in S3 ready to be fetched by the instance on boot`

frontend/build-tc.sh
davidfurey: You can achieve the same result with less teamcity magic by configuring the sbt-riffraff-artifact plugin to include it:

https://github.com/guardian/support-frontend/blob/master/build.sbt#L82

frontend/build-tc.sh
johnduffell: https://github.com/guardian/membership-frontend/pull/1706

frontend/build-tc.sh
johnduffell: thanks!

PR #1707, opened 2017-07-26
don't try to reuse the snowplow tracker and subject

johnduffell:
## Why are you doing this?
Since https://github.com/guardian/membership-frontend/pull/1704 the snowplow events have slowed to glacial pace (i.e. only a few today)
This is probably due to the snowplow java library not being pure functional in many ways, therefore making it hard to refactor our use of.

Since the documentation isn't brill, this is kind of an educated stab in the dark swinging closer back to what we had originally.  I.e. only reusing the emitter which is the thing that creates the async client, but not reusing the subject or tracker.

If that fails, it seems like not many people know much about this library, so we'll have to find a third way.  I'm very much reluctant to prioritise analytics over actual site stability for the core user need of signing up, even though it's not a complete breakage.

But fingers crossed this will work and everyone will be happy again!

@jacobwinch @dominickendrick

### DIFF
--- a/frontend/app/tracking/ActivityTracking.scala
+++ b/frontend/app/tracking/ActivityTracking.scala
@@ -298,7 +298,9 @@ trait ActivityTracking {
 
   private def executeTracking(data: TrackerData) {
     try {
-      val tracker = ActivityTracking.getTracker
+      val emitter = ActivityTracking.getEmitter
+      val subject = new Subject
+      val tracker = new Tracker(emitter, subject, "membership", "membership-frontend")
       val dataMap = data.toMap
       tracker.trackUnstructuredEvent(dataMap)
     } catch {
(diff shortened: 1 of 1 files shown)


jacobwinch:
👍

dominickendrick:
👍

johnduffell:
think it's back
image
@jranks123  is thinking about how we need to deal with missing data

johnduffell:
spoke too soon, there's a concurrency issue in the version of snowplow client we're using, it uses an arraylist in the emit call for the emitter.  so not thread safe.  However every time we create one, we leak a thread pool.  so basically the library is impossible to use
I notice they fixed this in a more recent version by synchronising that method (love java...)
https://github.com/snowplow/snowplow-java-tracker/blob/master/src/main/java/com/snowplowanalytics/snowplow/tracker/emitter/BatchEmitter.java#L86
old version was
https://github.com/snowplow/snowplow-java-tracker/blob/java-0.5.2/snowplow-java-tracker-core/src/main/java/com/snowplowanalytics/snowplow/tracker/core/emitter/Emitter.java#L142

not sure how feasible it is to upgrade to the non broken version as we seem to be on a snapshot of 0.5.2

davidfurey:
It might be easier to wrap a single instance of the client in an Actor than to upgrade the client?

johnduffell:
ah thanks didn't see that @davidfurey that's a good idea, for a stop gap I've just done [this](https://github.com/guardian/membership-frontend/pull/1708)

PR #1708, opened 2017-07-27
synchronise calls to tracker between threads

johnduffell:
## Why are you doing this?
So as per the problem reported in https://github.com/guardian/membership-frontend/pull/1707#issuecomment-318320866
I couldn't update the client due to some [weird java generics trick](https://github.com/snowplow/snowplow-java-tracker/blob/7b2eb085acc47288222bd58df042c3ffd6d4b407/src/main/java/com/snowplowanalytics/snowplow/tracker/emitter/BatchEmitter.java#L66) that scalac can't handle.
Anyway in the end since these figures are quite important for populating the data lake, I've decided to go for the next dirty hack short of rolling back.

## Changes
Wrap the track call (that uses an ArrayList) in a synchronised block.

## Screenshots (of my face)
image

@dominickendrick @jranks123 @paulbrown1982

### DIFF
--- a/frontend/app/tracking/ActivityTracking.scala
+++ b/frontend/app/tracking/ActivityTracking.scala
@@ -302,7 +302,9 @@ trait ActivityTracking {
       val subject = new Subject
       val tracker = new Tracker(emitter, subject, "membership", "membership-frontend")
       val dataMap = data.toMap
-      tracker.trackUnstructuredEvent(dataMap)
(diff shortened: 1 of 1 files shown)


paulbrown1982:
👍 if there's no other practicable way.

johnduffell:
no-one's a winner from this PR 😣

PR #1710, opened 2017-07-27
log the actual exception

johnduffell:
## Why are you doing this?
Bad logging... we're only logging the message (i.e. "null") rather than the whole exception.  This makes it hard to know what really went wrong.

@paulbrown1982 @dominickendrick @davidfurey @jranks123


paulbrown1982:
👍

PR #1711, opened 2017-07-27
synchronise on the right object!

johnduffell:
## Why are you doing this?
So the snowplow logging keeps breaking.  The solution [was to synchronise on the track call](https://github.com/guardian/membership-frontend/pull/1708), but I synchronised on the wrong object.  This fixes that.

In related news, along my journey through the snowplow code, I found that the snowplow client "async" mode basically [blocks the current thread while it's doing an async request](https://github.com/snowplow/snowplow-java-tracker/blob/java-0.5.2/snowplow-java-tracker-core/src/main/java/com/snowplowanalytics/snowplow/tracker/core/emitter/Emitter.java#L212), meaning the thread pools we/it were leaking were all useless (especially given the code that uses the pool is not itself threadsafe!)  So a whole pool was being created just so that the main thread could block waiting for the single possible concurrent request.  seriously!

I would recommend against updating the latest client as it's not much better, just don't use it!  As is the plan...

@jranks123 @dominickendrick @paulbrown1982 @davidfurey

### DIFF
--- a/frontend/app/tracking/ActivityTracking.scala
+++ b/frontend/app/tracking/ActivityTracking.scala
@@ -302,7 +302,7 @@ trait ActivityTracking {
       val subject = new Subject
       val tracker = new Tracker(emitter, subject, "membership", "membership-frontend")
       val dataMap = data.toMap
-      tracker.synchronized {
+      emitter.synchronized {
         tracker.trackUnstructuredEvent(dataMap)
       }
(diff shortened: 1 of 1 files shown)


johnduffell:
@paulbrown1982 re our chat actually the synchronised block is still needed because it's protecting the ArrayList, not the thread pool.

PR #1713, opened 2017-07-31
Reduce log noise from health monitoring task

jacobwinch:
## Why are you doing this?
* The health monitoring task currently logs every time it completes successfully (at the moment this is every 10 seconds). This means there is a lot of noise in the logs, which is making it harder to debug problems in production.

## Changes
* Log success case at debug level so it does not appear in CloudWatch logs.

cc @johnduffell

### DIFF
--- a/frontend/app/monitoring/HealthMonitoringTask.scala
+++ b/frontend/app/monitoring/HealthMonitoringTask.scala
@@ -72,8 +72,11 @@ object Scheduler extends StrictLogging {
   )(implicit system: ActorSystem, executionContext: ExecutionContext) = {
     logger.info(s"Starting $task.name scheduled task with an initial delay of: $initialDelay. This task will refresh every: $interval")
     system.scheduler.schedule(initialDelay, interval) {
-      task.task().onFailure {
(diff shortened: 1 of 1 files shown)


johnduffell:
that's the whole point of kibana 😆
https://membership-logs.gutools.co.uk/app/kibana#/discover?_g=()&_a=(columns:!(_source),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'logstash-*',key:app,negate:!f,type:phrase,value:membership-frontend),query:(match:(app:(query:membership-frontend,type:phrase)))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'logstash-*',key:logger_name,negate:!t,type:phrase,value:'monitoring.Scheduler$'),query:(match:(logger_name:(query:'monitoring.Scheduler$',type:phrase))))),index:'logstash-*',interval:auto,query:(query_string:(analyze_wildcard:!t,query:'*')),sort:!('@timestamp',desc))

johnduffell:
I'm not overly possessive about this log line though, but we should make sure the logging system is fit for purpose. my recommendation would be change to debug, and we can make sure kibana only shows info and above by default. 👍

johnduffell:
👍 🎉 🌳

PR #1714, opened 2017-08-01
Revert "Reduce log noise from health monitoring task"

jacobwinch:
Reverts guardian/membership-frontend#1713

No idea how/why... but merging guardian/membership-frontend#1713 seems to have been preventing the healthcheck from passing. I'd like to revert this while I investigate...


jranks123:
👍

PR #1715, opened 2017-08-01
Revert "Revert "Reduce log noise from health monitoring task""

jacobwinch:
The issues with deployment seem to have been related to a bad AMI. We noticed the following error in /var/log/cloud-init-output.log:

```
# install and start play app and create frontend user
dpkg -i /dist/frontend_1.0-SNAPSHOT_all.deb
dpkg: error: dpkg status database is locked by another process
```

After successfully re-deploying the AMI which was baked on 31/07, we re-baked in Amigo and deployed again without issue.

As the problem was unrelated to guardian/membership-frontend#1713 I'm going to try shipping it again.

Thanks for the help @johnduffell


johnduffell:
👍 approved in person earlier

PR #1716, opened 2017-08-01
Check that dpkg lock is free before installing app

jacobwinch:
## Why are you doing this?
We had some issues with deploys this morning (see https://github.com/guardian/membership-frontend/pull/1715 and related PRs).

After some investigation with @johnduffell we traced this back to the fact that an automated update process (https://help.ubuntu.com/lts/serverguide/automatic-updates.html) had a lock on the dpkg status database.

This led to the following error when attempting to install and startup the app:

```
# install and start play app and create frontend user
dpkg -i /dist/frontend_1.0-SNAPSHOT_all.deb
dpkg: error: dpkg status database is locked by another process
```

...which obviously meant that the healthcheck was never going to pass!

For some reason the following AMI seems to significantly increase the chances of falling foul of this race condition:
https://amigo.gutools.co.uk/recipes/xenial-membership/bakes/306

However, it's quite possible that we have been hitting this issue on individual instances before without seeing a major impact. (It's fairly tricky to debug this problem because it requires detaching a dying instance from the ASG before AWS terminates it).

This change means that we wait for the dpkg lock to be free before attempting to install/start the app.

## Trello card: 
N/A

## Changes
* Add a check to ensure that dpkg lock is free before attempting to install the app
* Tweak bash options and user data to stop the script from being abandoned when lsof exits with return code of 1.

## Testing
* Deployed to code using the troublesome AMI and the existing scripts, in order to confirm that boxes consistently failed to come into service.
* Cloudformed the CODE stack to confirm that the script changes successfully avoid this race condition and allow instances to come into service.

cc @rupertbates @paulbrown1982 @davidfurey

### DIFF
--- a/frontend/instance/frontend/init-instance.sh
+++ b/frontend/instance/frontend/init-instance.sh
@@ -1,4 +1,4 @@
-#!/bin/bash -ev
+#!/bin/bash -v
 # this script is called by all instances when they first start.
 # you can test it standalone by exporting the stack/stage/app/region variables and running it manually
 
--- a/frontend/instance/frontend/init-instance.sh
+++ b/frontend/instance/frontend/init-instance.sh
@@ -15,11 +15,11 @@ lsof /var/lib/dpkg/lock
 rc=$?
 echo "Return code is $rc"
 until [ "$rc" -eq "1" ]; do
-echo "Waiting for dpkg lock to become free"
-sleep 1s;
-lsof /var/lib/dpkg/lock
-rc=$?
-echo "Return code is $rc"
+  echo "Waiting for dpkg lock to become free"
+  sleep 1s;
+  lsof /var/lib/dpkg/lock
+  rc=$?
+  echo "Return code is $rc"
 done
 echo "dpkg lock is free, ready to install application"
 
--- a/frontend/instance/frontend/init-instance.sh
+++ b/frontend/instance/frontend/init-instance.sh
@@ -10,6 +10,19 @@ aws --region $region s3 cp --recursive s3://membership-dist/${stack}/${stage}/fr
 mkdir /etc/gu
 aws --region $region s3 cp s3://membership-private/${stage}/membership.private.conf /etc/gu
 
+# wait for dpkg lock to become free
+lsof /var/lib/dpkg/lock
+rc=$?
+echo "Return code is $rc"
+until [ "$rc" -eq "1" ]; do
+echo "Waiting for dpkg lock to become free"
+sleep 1s;
+lsof /var/lib/dpkg/lock
+rc=$?
+echo "Return code is $rc"
+done
+echo "dpkg lock is free, ready to install application"
+
 # install and start play app and create frontend user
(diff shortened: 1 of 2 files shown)


jacobwinch:
If we decide that we want to take this approach we should also update the userData scripts on our other apps, as they could also be exposed to the same issue.

johnduffell:
thanks @jacobwinch , it's a shame our bash scripting skills are so poor but even so I think it's pretty clear. Agree we should update the other apps, although on reflection, I'm not sure why we use dpkg to install our app in the first place?

jacobwinch:
It looks like we started using dpkg here: https://github.com/guardian/membership-frontend/pull/1376

paulbrown1982:
👍 we can quickly effect this change on other apps if we start seeing the same failures.

jacobwinch:
We're going to:

1. Cloudform with the template from this branch
2. Merge this PR which will start a deploy with the new init.instance.sh script

jacobwinch:
Looking at the tests... PROD sign-ups look fine.

jacobwinch:
It looks like we timed out waiting for Stripe Checkout:
https://saucelabs.com/beta/tests/4eaae8fefd374304bd1a01bff4dc5589/commands#212
https://saucelabs.com/beta/tests/4eaae8fefd374304bd1a01bff4dc5589/commands#263

I'm confident this isn't related to my change so I'm going to re-run the tests...

frontend/instance/frontend/init-instance.sh
```diff
@@ -10,6 +10,19 @@ aws --region $region s3 cp --recursive s3://membership-dist/${stack}/${stage}/fr
...
+  echo "Waiting for dpkg lock to become free"
+  sleep 1s;
+  lsof /var/lib/dpkg/lock
+  rc=$?
+  echo "Return code is $rc"
+done
+echo "dpkg lock is free, ready to install application"
+
 # install and start play app and create frontend user
 dpkg -i /dist/frontend_1.0-SNAPSHOT_all.deb
```
mchv: Obviously it is only reducing the chance to have a contention issue, as a lock could be taken between your last lock check and the call to `dpkg`, but the probability is very low.

Through this [stackoverflow answer](https://askubuntu.com/a/231859) I discovered  `aptdcon` which is another way to solve this problem, but requires it to be installed on the base image.

mchv:
@jacobwinch @johnduffell good investigation work guys 💯

frontend/instance/frontend/init-instance.sh
johnduffell: agreed thanks for the comments, we checked the lock status for the first 30 seconds from when the userdata started, second 0 was held by one process, second 1-9ish was held by the unattended-update process, and the remaining time it was free all the time.  Another approach would have been to keep retrying dpkg inside the loop, instead of lsof, but it seemed neater to just try once.

aptdcon is an interesting one, although in our case we have to wait for the package to install before we can do some chown with the user that's from the package, so we'd have to block until it was installed.  But that's not to say any of the stuff we're doing is best practice!

PR #1725, opened 2017-09-20
Support for Australian Supporters to use our Australia Stripe account

paulbrown1982:
## Why are you doing this?

Added support for Australian customers to use our Australia Stripe account so that they don't incur banking charges.

## Changes
- Added a second StripeService to MemberService basically.
- Renamed 'giraffe' nomenclature where I could.
- Removed the contributions stuff so I don't have to configure the payment gateway, and I don't have to regression test those flows. I'll remove the client-side code and the bundles pages in a subsequent PR @rupertbates 
- These changes also include the [membership-common](https://github.com/guardian/membership-common/pull/530) upgrade to subscriptionService.either.
- Also pulls in [membership-common](https://github.com/guardian/membership-common/pull/532) changes to support multiple Stripe API versions.

## Screenshots
cc @johnduffell @desbo @dominickendrick @davidfurey @rupertbates


frontend/app/services/MemberService.scala
```diff
@@ -456,21 +458,20 @@ class MemberService(identityService: IdentityService,
...
       val ratePlanChargeId = catalog.contributor.charges.chargeId.get
-
       val plan = RatePlan(productRatePlanId = planId.get,
                           chargeOverride = Some(ChargeOverride(productRatePlanChargeId = ratePlanChargeId,
                                                 price = Some(joinData.payment.amount))))
       val currency = GBP
       val today = DateTime.now.toLocalDate
-
-      Contribute(account = createAccount(contactId, identityId, currency, joinData.payment),
+      val transactingCountry = Country.UK // TODO get from GEO-IP
```
paulbrown1982: This method may be deletable if there are no and not to be any Recurring Contributions going through membership-frontend.

paulbrown1982: This method may be deletable if there are no and not to be any Recurring Contributions going through membership-frontend.

frontend/app/actions/ActionRefiners.scala
```diff
@@ -70,7 +70,7 @@ object ActionRefiners extends LazyLogging {
       authRequest = new AuthenticatedRequest(user, request)
       tp = authRequest.touchpointBackend
       member <- OptionEither(authRequest.forMemberOpt)
-      subscription <- OptionEither.liftEither(tp.subscriptionService.either[SubscriptionPlan.FreeMember, SubscriptionPlan.PaidMember](member))
+      subscription <- OptionEither.liftEither(tp.subscriptionService.either[SubscriptionPlan.FreeMember, SubscriptionPlan.PaidMember](member).map(_.toOption.flatten))
```
johnduffell: oh, we're already doing a liftEither, so no need to flatten the either out.  Just needed to remove the '.liftEither'

johnduffell: that assumes we want to keep the error if it failed, rather than replacing with None.  So the user will get a 500 error on line 88 instead of just logged out.

johnduffell: as discussed we will keep the old behaviour rather than surfacing the error that would have been treated as not found

frontend/app/services/paymentmethods/PaymentMethodInitialiser.scala
```diff
@@ -23,21 +23,27 @@ trait PaymentMethodInitialiser[PMCommand <: zuora.soap.models.Commands.PaymentMe
...
+    } yield {
+      val card = stripeCustomer.card
+      CreditCardReferenceTransaction(card.id, stripeCustomer.id, card.last4, CountryGroup.countryByCode(card.country), card.exp_month, card.exp_year, card.`type`)
     }
-  } yield {
-    val card = stripeCustomer.card
-    CreditCardReferenceTransaction(card.id, stripeCustomer.id, card.last4, CountryGroup.countryByCode(card.country), card.exp_month, card.exp_year, card.`type`)
   }
+
+  def appliesToCountry(country: Country): Boolean = countryWhitelist.contains(country) || !countryBlacklist.contains(country)
```
johnduffell: ok (but possibly could be a boolean!)

johnduffell:
yep good work! 👍

AWare:
:+1:

PR #1729, opened 2017-10-20
DO NOT MERGE: Optimise for new members-data-api approach

jacobwinch:
## Why are you doing this?
members-data-api now uses Zuora as it's primary data source. 

The members-data-api changes that @lmath has been working on mean that updating tier information in Salesforce (in order to trigger a webhook which was used to populate Dynamo) is now unnecessary. 

In fact, having Salesforce amending membership data in a Dynamo table which is now used more like a cache is a bad thing - from now on we only want members-data-api to read or write from/to this DB.

## Changes
* Remove membership/members-data-api comparison code. Since both now use Zuora as the data source, this check feels redundant.
* Remove any code which writes tier information into Salesforce.
* Remove some whitespace from tests

## Checklist
- [x] Disable outbound message in Salesforce
- [x] Tidy up outbound message receiver from members-data-api
- [ ] Confirm that there are no other SF dependencies on these API calls (need to do this before merging)

### DIFF
--- a/frontend/app/services/MemberService.scala
+++ b/frontend/app/services/MemberService.scala
@@ -128,11 +128,6 @@ class MemberService(identityService: IdentityService,
         case Failure(e) => logger.error(s"Could not create paid Zuora subscription: ID=${user.id}", e)
       }
 
-    def updateSalesforceContactWithMembership(stripeCustomer: Option[Customer]): Future[ContactId] =
-      salesforceService.updateMemberStatus(user.minimal, tier, stripeCustomer).andThen { case Failure(e) =>
-        logger.error(s"Could not update Salesforce contact with membership status for user ${user.id}", e)
-      }
-
     formData match {
       case payingForm: PaidMemberJoinForm =>
         val transactingCountry = payingForm.billingAddress.flatMap(_.country) orElse payingForm.deliveryAddress.country orElse ipCountry getOrElse UK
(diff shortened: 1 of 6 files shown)


jacobwinch:
cc @mario-galic - we can finally remove your FIXMEs!

frontend/app/services/MemberService.scala
```diff
@@ -128,25 +128,18 @@ class MemberService(identityService: IdentityService,
         case Failure(e) => logger.error(s"Could not create paid Zuora subscription: ID=${user.id}", e)
       }
 
-    def updateSalesforceContactWithMembership(stripeCustomer: Option[Customer]): Future[ContactId] =
-      salesforceService.updateMemberStatus(user.minimal, tier, stripeCustomer).andThen { case Failure(e) =>
```
jacobwinch: Ah, it looks like this is how the Stripe Customer makes it onto the contact. @paulbrown1982 - I think you wanted to remove this anyway?

paulbrown1982: We need to check if there are any SF triggers requiring this to take place during the upgrade flow.

jacobwinch: Yes, @MichaelOjoGIT raised a few concerns with this, so I'll do some further testing and then discuss with @davidpepper223 before merging...

PR #1734, opened 2017-10-26
Client-side Dependencies Upgrade - Part 3 - Raven-js

Ap0c:
## Why are you doing this?

Part of the ongoing effort to bring our client-side dependencies up-to-date. This time: raven-js.

This is based on the `update-babel` branch, so it'll pick up those changes in the diff until that gets merged into master.

**Note:** For anyone wondering why I'm doing these upgrades in such small chunks, it's because in most cases these are fairly big version jumps, so there's a reasonable chance of stuff breaking. I'm trying to keep each change as isolated as possible to make it easier to test, deploy to CODE and so on, and then potentially roll back individually in case of any problems.

[**Trello Card**](https://trello.com/c/LBGmydEm/1047-update-ravenjs)

## Changes

- Bumped `raven-js` to the latest version.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -55,7 +55,7 @@
     "mkdirp": "^0.5.1",
     "phantomjs-prebuilt": "^2.1.13",
     "qwery": "~4.0.0",
-    "raven-js": "^3.0.4",
+    "raven-js": "^3.19.1",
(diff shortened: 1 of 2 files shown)


Ap0c:
Can't see any errors in PROD.

PR #1737, opened 2017-10-30
Remove HTTP access from the security groups

rupertbates:
## Why are you doing this?
As part of our GDPR work I've been checking the security group access on our apps to make sure we aren't allowing any more than we need to. 
We now use Fastly to force a redirect from HTTP to HTTPS for all traffic so the port 80 ingress points should never be used.

## Trello card: [Here](https://trello.com/c/QVxm5eYS/1014-check-that-security-groups-for-support-frontend-and-membership-frontend-expose-the-minimum-entry-and-exit-ports)


johnduffell:
:gdpr: 🔒 👍 🚀

PR #1739, opened 2017-11-01
Remove shared ssh key access

rupertbates:
## Why are you doing this?
More GDPR work, removing access to running instances using the aws-membership key pair as all
ssh access for team members should be managed by github teams via [github-keys-to-s3-lambda](https://github.com/guardian/github-keys-to-s3-lambda)

## Trello card: [Here](https://trello.com/c/WLCRICJv/1017-check-that-support-frontend-and-membership-frontend-is-not-accessible-via-a-shared-ssh-key)


rupertbates:
@davidfurey it is and I've tested it on CODE

PR #1746, opened 2017-11-06
Reverted cloud formation changes

svillafe:
## Why are you doing this?
Revert cloud formation PR

### DIFF
--- a/cloud-formation/membership-app.cf.yaml
+++ b/cloud-formation/membership-app.cf.yaml
@@ -261,11 +261,6 @@ Resources:
         ToPort: '9000'
         SourceSecurityGroupId:
           Ref: LoadBalancerSecurityGroup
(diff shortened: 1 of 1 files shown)


jacobwinch:
Due to spikes in the following errors (which we believe have been caused by this change):
https://sentry.io/the-guardian/membership/issues/276458487/
https://sentry.io/the-guardian/membership/issues/399956804/
https://sentry.io/the-guardian/membership/issues/58960782/

paulbrown1982:
Also what about NTP/LDAP/SSH/SCP? Would they be blocked only letting 443 outbound?

svillafe:
In the following PR we fix: 
https://sentry.io/the-guardian/membership/issues/276458487/
https://sentry.io/the-guardian/membership/issues/399956804/

PR #1748, opened 2017-11-06
Https for content api

svillafe:
## Why are you doing this?
To make CAPI requests under `https`


tomrf1:
👍

tomrf1:
@svillafe @davidfurey I've just released v11.39 of content-api-client, which includes the switch to https

PR #1752, opened 2017-11-13
Send acquisition events

desbo:
## Why are you doing this?
To track membership signups in the new Acquisitions table. 

Note that I have moved Ophan to a `package.json` dependency, this means we can drop the old module that served as a wrapper for the `curl` request (and we don't need to make the request at all). 

TODO: see if it's necessary to attach `pageviewId` to the upgrade forms


frontend/app/model/MembershipAcquisitionData.scala
```diff
@@ -0,0 +1,80 @@
...
+    case BillingPeriod.Quarter => PaymentFrequency.Quarterly
+    case BillingPeriod.SixMonthsRecurring => PaymentFrequency.SixMonthly
+    case BillingPeriod.Year => PaymentFrequency.Annually
+    case _: OneOffPeriod => PaymentFrequency.OneOff
+  }
+
+  private def paymentMethodToProvider(method: PaymentMethod): PaymentProvider = method match {
+    case _: PaymentCard => PaymentProvider.Stripe
+    case _: PayPalMethod => PaymentProvider.Paypal
+    case _: GoCardless => PaymentProvider.Gocardless
```
jranks123: Do we need to do exhaustive pattern matches here?

desbo: It is exhaustive - these are the only possible types: https://github.com/guardian/membership-common/blob/master/src/main/scala/com/gu/memsub/PaymentMethod.scala

PR #1753, opened 2017-11-15
Client-side Dependencies Upgrade - Part 10 - Bower

Ap0c:
## Why are you doing this?

Not technically a client-side library, but it pulls in content that gets used on the client-side (guss/pasteup), so worth bringing up to date.

[**Trello Card**](https://trello.com/c/XC8ppyYj/1095-update-bower-on-membership)

## Changes

- Bumped bower to latest version.


Ap0c:
Looks ok.

PR #1754, opened 2017-11-16
Skip decoding session data with play-json for now

desbo:
This is a temporary workaround that I'll need to deploy now until I can
fix it properly tomorrow.

## Why are you doing this?
A part of `acquisition-event-producer` uses a json feature from Play 2.4 that doesn't exist in 2.5.
This causes a fatal exception when we try to decode the data on line 34. Thankfully, this will not affect PROD at the moment as there's an issue where the session data isn't passed along to this method, so it never gets as far as the `fromJson` call.

This PR is just to be on the safe side so that the dangerous code isn't included at all until I fix it properly tomorrow.


johnduffell:
👍

PR #1755, opened 2017-11-20
Use latest acquisition-event-producer and add test

desbo:
This is a proper fix for the workaround I added here: https://github.com/guardian/membership-frontend/pull/1754

`acquisition-event-producer` is now published for Play 2.5, so the JSON decoding works.


frontend/app/actions/CommonActions.scala
```diff
@@ -116,6 +118,10 @@ trait CommonActions extends LazyLogging {
   def ChangeToPaidAction(targetTier: PaidTier) = SubscriptionAction andThen checkTierChangeTo(targetTier)
 }
 
+object CommonActions {
+  val acquisitionDataSessionKey: String = "acquisitionData"
```
rupertbates: :1st_place_medal:

PR #1757, opened 2017-11-22
[WIP] Removing snowplow 

dominickendrick:
## Why are you doing this?
This removes the calls to the `event-stream-analytics` tracking pipeline AKA `snowplow`.
This system is being deprecated in favour of using the acquisitions pipeline in Ophan.

image

This should only be merged once we have confirmed all new tracking works correctly

@guardian/contributions


project/Dependencies.scala
```diff
@@ -17,7 +17,6 @@ object Dependencies {
   val playCache = PlayImport.cache
   val awsSimpleEmail = "com.amazonaws" % "aws-java-sdk-ses" % awsClientVersion
   val sqs = "com.amazonaws" % "aws-java-sdk-sqs" % awsClientVersion
-  val snowPlow = "com.snowplowanalytics" % "snowplow-java-tracker" % "0.5.2-SNAPSHOT"
```
rtyley: Now that this dependency is being removed you can also remove the 'snapshots' resolver! :confetti_ball: 

It was originally added here... https://github.com/guardian/membership-frontend/pull/191/files#r152640638

project/Dependencies.scala
```diff
@@ -17,7 +17,6 @@ object Dependencies {
...
   val bCrypt = "com.github.t3hnar" %% "scala-bcrypt" % "3.1"
```
rtyley: Do we still need `bCrypt`?

dominickendrick: No, I'll remove it : )

project/Dependencies.scala
dominickendrick: Done!

Mullefa:
As discussed, Bolt uses a data source dependant on Snowplow to report on supporters. Therefore, we should check it is ok to stop reporting on supporters in Bolt before merging this.

johnduffell:
is this going anywhere? there's a nice set of conflicts here!

davidfurey:
@johnduffell - I hope so, this this will allow us to block non-https traffic for membership-frontend, ticking another thing off CyberEssentials.

johnduffell:
🎉

PR #1763, opened 2017-12-13
Tweak Header And Brandbar

Ap0c:
## Why are you doing this?

Tweaking some scaling and padding in the header and the brandbar. Visible locally only at the moment.

## Changes

- Tweaked scaling and padding in header.
- Tweaked scaling and padding in brandbar.
- Switched to using a map rather than an array for grunt assets.


Ap0c:
The ASG is a bit out of whack, fixing now ^.

PR #1765, opened 2018-01-09
Marketing consent

rupertbates:
## Why is this PR necessary?
As part of our GDPR work, we need to change the way we collect marketing consent.
Trello card [here](https://trello.com/c/PFLEC9tg/1158-marketing-consent-on-membership)
cc @jfsoul

## What has changed?
1. Add a new consent checkbox to the checkout pages, this is to replace the consent checkbox on Identity which is going to be removed.
2. Change the way that consent is persisted to Identity:

## How?
Previously we have stored marketing consent by posting to the user/me endpoint with a boolean representing whether the user checked the consent checkbox. Although we will continue to do this for now, Identity will soon start ignoring this value in favour of a [new API](https://docs.google.com/document/d/1JDEpehzToi9aAMg4Fk7n_mnvQ_GOOf_kgOPSyWLBweA/)

### The new method
If user gives consent:
* We call Identity on the new consent-email endpoint to instruct them that the user has given consent to 'supporter' related emails. 
* Identity emails the user asking them to click a link to confirm their choice.
* NB we are no longer passing any marketing permission data to Salesforce.

If user does not give consent then no call to the consent emailer is made.

### Screenshots
**The new marketing box:**
screen shot 2018-01-10 at 17 26 54


frontend/app/services/IdentityService.scala
```diff
@@ -24,18 +26,17 @@ import views.support.IdentityUser
...
 case class IdentityServiceError(s: String) extends Throwable {
   override def getMessage: String = s
 }
 
-object IdentityService {
-  val DisregardResponseContent: (WSResponse => Either[String, Unit]) = resp => Right(Unit)
+object IdentityService extends LazyLogging {
+  val DisregardResponseContent: (WSResponse => Either[String, Unit]) = {
+    resp =>
+      logger.debug(s"Webservice returned code ${resp.status}, and body: ${resp.body}")
```
jacobwinch: Is this debug statement still needed?

rupertbates: I think it can be useful because otherwise this method completely throws away the response, even if it is an error.

PR #1768, opened 2018-01-12
Reduce the length of the marketing consent text

rupertbates:
## Why are you doing this?
Because the original marketing consent text we were given was ridiculously long and looked silly

## Screenshots
screen shot 2018-01-12 at 13 33 00


jacobwinch:
👍

PR #1770, opened 2018-01-15
Update Stripe Roundel

Ap0c:
## Why are you doing this?

Updating the stripe roundel to use the new roundel asset.

## Changes

- Updated the asset link for the roundel.

## Screenshots

**Before:**

roundel-old

**After:**

roundel-new


Ap0c:
Working in PROD.

PR #1773, opened 2018-01-17
Remove css lowercase transforms.

AWare:
## Why are you doing this?

To match the new brand.

## Changes
Removes text-transform: lowercase

## Screenshots
tk


frontend/app/views/bundle/bundlesLanding.scala.html
```diff
@@ -23,7 +23,7 @@
 
 	<section class="introduction">
 		<div class="introduction__content">
-			<h1 class="introduction__heading">support the guardian</h1>
```
Ap0c: I don't think you can get to this page any more, it's the pre-support bundles landing page.

PR #1776, opened 2018-01-30
Replace stale client-side libraries part one: Bonzo and Qwery

JustinPinner:
## Why are you doing this?
Don't judge me. Yet.

We found some old libraries that haven't been updated in ages, and thus have no straightforward upgrade path. While it's true that they aren't causing any specific problems at the moment there's a danger that at some point, a security flaw here or another upgrade there may cause us a headache, and we'll have to react to the breakage quickly.

Instead, what if we replaced those libraries now, with something that is currently still under active development, widely used and publicly scrutinised for problems? What if the first of those happened to be jQuery, to replace Bonzo and Qwery?

Tada! Here it is.

Initial tests run locally look like it's workable. There were a few minor tweaks needed to compensate for the change, but on the whole, not too dramatic it seems. I've also tested many things on CODE (see below) and nothing is obviously broken that I can see.

If you know of an alternative to jQuery that is likely to be around for the foreseeable future, is currently well maintained and is equally likely to be a more-or-less drop in replacement then bung the details in some comments, or forever remain complicit ;) 

## Changes
* Bonzo: you're outta here
* Qwery: dismissed
* jQuery: hello

## Signup / cancellation paths tested
Sign up as Friend -> (ok)
* upgrade to Supporter -> (ok)
* cancel membership -> (ok)

Sign up as Supporter -> (ok)
* upgrade to Partner (ok)
* cancel membership (ok)

Sign up as Partner -> (ok)
* upgrade to Patron -> (ok)
* downgrade to Supporter -> unable to change from Patron to Supporter (ok)
* cancel membership -> (ok)

Sign up as Patron -> (ok)
* downgrade to Partner -> Unable to change from Patron to Partner (ok)
* cancel membership -> (ok)

## Other
Page rendering -> (ok)
Search for masterclasses (filter list) -> (ok)
Form field validation/focus -> (ok)


AWare:
bonzo and qwery remain in use in frontend, so I'm not sure whether they're priority targets for removal.

JustinPinner:
@AWare Hi Alex, it's true they're still in frontend, but they are in line to be replaced/removed.

PR #1778, opened 2018-02-07
Remove deprecation warnings

davidfurey:
## Why are you doing this?

First step to upgrading to Play 2.6.  This also adds compile-time dependency injection since this was the easiest way to remove the deprecation warning about using a GlobalSettings object

## Changes
* Remove uses of Play.current
* Remove uses of WS object
* Remove uses of Akka agent
* Remove uses of scala.concurrent.ExecutionContext.Implicits.global
* Dependency inject all controllers (first run-time, later compile-time)
* Remove unused NOCSRF


frontend/app/services/TouchpointBackend.scala
```diff
@@ -126,6 +119,18 @@ object TouchpointBackend extends LazyLogging {
...
   }
 
+  case class Resolution(
+    backend: TouchpointBackend,
+    typ: BackendType,
+    validTestUserCredentialOpt: Option[TestUserCredentialType[_]]
+  )
+}
+
+class TouchpointBackendProvider(implicit val system: ActorSystem, implicit val executionContext: ExecutionContext, implicit val wsClient: WSClient) extends LazyLogging {
```
johnduffell: I called this TouchpointBackends in subs frontned

frontend/app/controllers/Bundle.scala
```diff
@@ -8,7 +8,7 @@ import play.api.mvc.Controller
 import services._
 import views.support.PageInfo
 
-trait Bundle extends Controller {
+class Bundle(touchpointBackend: TouchpointBackendProvider) extends Controller {
```
johnduffell: the naming is a little inconsistent, I can see why (the old class name matches)

frontend/app/controllers/CachedAssets.scala
```diff
@@ -4,7 +4,7 @@ import play.Logger
 import play.api.mvc.{Action, Controller}
 import play.api.libs.concurrent.Execution.Implicits.defaultContext
 
-object CachedAssets extends Controller {
+class CachedAssets() extends Controller {
```
johnduffell: just a heads up when I switched to play 2.6 this started 404ing all the assets in dev, so I had to change to DI for the Assets controller into this class.

frontend/app/wiring/AppComponents.scala
```diff
@@ -0,0 +1,96 @@
+package wiring
+
+import actions.TouchpointOAuthActions
+import play.api.routing.Router
+import play.api.libs.ws.ahc.AhcWSComponents
+import play.api.mvc.EssentialFilter
+import play.api.BuiltInComponentsFromContext
+import com.softwaremill.macwire.wire
```
johnduffell: I didn't use wire in subs frontend (or @pvighi  in members data api) as I'm not sure it buys us much for an extra dependency.  I'm guessing intellij would find and check everything still as well as in SBT?

frontend/conf/application.conf
```diff
@@ -1,6 +1,8 @@
 # Config for the 'Vanilla' Membership site
 include "copy"
 
+play.application.loader = "wiring.AppLoader"
```
johnduffell: I put the subs one in the root package and called it MyApplicationLoader (not sure that was ideal naming, but it came from the play exaple docs)

johnduffell:
great! I've flagged a few differences to how I did things in subs in case you prefer to be consistent with naming (or you could ask me to rename stuff!)
I also think you could have split it down a bit more, as you can do just the controllers without having to touch other things, then just add them in later.
I assume you've done testing of sign up and tier changes etc.
Other than that it's a great job and thanks for putting the time in! 👍 Ship it! 🚀

frontend/app/actions/TouchpointActionRefiners.scala
```diff
@@ -0,0 +1,118 @@
...
+import play.api.mvc._
+import views.support.MembershipCompat._
+
+import scalaz.{-\/, EitherT, OptionT, \/, \/-}
+import scalaz.std.scalaFuture._
+
+import scala.concurrent.{ExecutionContext, Future}
+import scalaz.{-\/, \/, \/-}
+
+class TouchpointActionRefiners(implicit val touchpointBackend: TouchpointBackendProvider, implicit private val ec: ExecutionContext) extends LazyLogging {
```
jacobwinch: Do we need the word 'Touchpoint' here?

frontend/app/actions/TouchpointActionRefiners.scala
```diff
@@ -0,0 +1,118 @@
...
+  private def getContributorRequest[A](request: AuthRequest[A]): Future[String \/ Option[SubReqWithContributor[A]]] = {
```
jacobwinch: I appreciate that this has just been moved from another file, but do we need action refiners related to contributions in this project anymore? I guess we should probably tidy this up in a separate PR though...

frontend/app/actions/TouchpointCommonActions.scala
```diff
@@ -0,0 +1,47 @@
...
+import actions.ActionRefiners.{freeSubscriptionRefiner, paidSubscriptionRefiner}
+import actions.Fallbacks.joinStaffMembership
+import com.gu.salesforce.PaidTier
+import play.api.mvc.Results.Forbidden
+import play.api.mvc.{ActionTransformer, Request}
+import services.{AuthenticationService, TouchpointBackendProvider}
+
+import scala.concurrent.Future
+
+class TouchpointCommonActions(touchpointBackendProvider: TouchpointBackendProvider, touchpointActionRefiners: TouchpointActionRefiners) {
```
jacobwinch: As above

frontend/app/services/checkout/identitystrategy/Strategy.scala
```diff
@@ -7,12 +7,11 @@ import play.api.mvc.{RequestHeader, Result}
 import services.AuthenticationService.authenticatedIdUserProvider
 import services.{IdentityApi, IdentityService}
 
-import scala.concurrent.Future
+import scala.concurrent.{ExecutionContext, Future}
```
jacobwinch: Is there a reason we chose to use Scala's execution context [over Play's](https://www.playframework.com/documentation/2.5.x/ThreadPools#Using-the-default-thread-pool)?

jacobwinch:
Great work - this must have been painful! 👍

frontend/app/services/checkout/identitystrategy/Strategy.scala
johnduffell: that's the import of the class definition, not the execution context implicit at runtime

frontend/conf/application.conf
davidfurey: I copied this from support-frontend, which copied it from somewhere else that I can't remember.

frontend/app/services/TouchpointBackend.scala
davidfurey: I will rename for consistency

frontend/app/controllers/CachedAssets.scala
davidfurey: Good to know, thanks

PR #1780, opened 2018-02-12
Play 2.6 upgrade

davidfurey:
## Why are you doing this?

GDPR

## Changes
* Upgrade to SBT 1.0
* Upgrade to Play 2.6
* Upgrade Cats and Scalaz


johnduffell:
> presumably the tests are failing because we're waiting for guardian/acquisition-event-producer#26 to be released?

I've just shipped it as v 3.0.0 🚀

frontend/app/controllers/CacheBustedAssets.scala
```diff
@@ -3,8 +3,8 @@ package controllers
 import play.api.mvc.{AnyContent, Action, Controller}
 import views.support.Asset
 
-class CacheBustedAssets() extends Controller{
+class CacheBustedAssets(asset: Assets) extends Controller {
   def at(path: String): Action[AnyContent] =  {
-    controllers.Assets.at("/public", Asset.map(path), true)
+    asset.at("/public", Asset.map(path), true)
```
johnduffell: missing an `s` (sorry it's the best I can do, not much else to comment on!)

frontend/conf/ahc.properties
```diff
@@ -0,0 +1,91 @@
+org.asynchttpclient.threadPoolName=AsyncHttpClient
```
johnduffell: what's all this and do we need it in subs too?

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=0.13.16
+sbt.version=1.0.2
```
johnduffell: might have been possible to do the sbt PR separately, just to keep the size down?  But I realise some things tangle together a lot.

johnduffell:
great, looks good, I wasn't sure what the whole shaded thing was, I don't know if I should add that to subs frontend too?  Other than that only nit picking - go for it! 👍 🚀

frontend/conf/ahc.properties
jacobwinch: I had the same question - there are more details here: https://github.com/guardian/membership-frontend/pull/1780/commits/c8f465673f7dfa512d11b3aa763f180dc3c6ed68

johnduffell: thanks! not sure if it's worth putting that in a file comment too

PR #1782, opened 2018-02-13
Workaround legacy hashing warning

davidfurey:
## Why are you doing this?

Having warnings that we don't need to worry about is noise that hides genuine warnings.  This PR masks Google's deprecation of md5

## Trello card: [Here](https://trello.com/c/OqOfINff/1229-upgrade-membership-frontend-to-play-26)

### DIFF
--- a/frontend/app/utils/LegacyHashing.java
+++ b/frontend/app/utils/LegacyHashing.java
@@ -1,0 +1,21 @@
+package utils;
+
+import com.google.common.hash.HashFunction;
+import com.google.common.hash.Hashing;
+
+/**
+ * Provides an md5 and sha1 hash function without producing deprecation warnings.
+ * From: https://github.com/google/guava/issues/2841
+ */
+
+public class LegacyHashing {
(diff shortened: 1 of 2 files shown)


frontend/app/utils/LegacyHashing.java
```diff
@@ -0,0 +1,21 @@
...
+
+import com.google.common.hash.HashFunction;
+import com.google.common.hash.Hashing;
+
+/**
+ * Provides an md5 and sha1 hash function without producing deprecation warnings.
+ * From: https://github.com/google/guava/issues/2841
+ */
+
+public class LegacyHashing {
```
pvighi: does this need to be a class with just static methods or can it just be an object?

davidfurey: I don't think you can suppress warnings on individual lines in Scala, so this file has to be Java

pvighi: Oh, somehow  I didn't even notice it was java.

PR #1785, opened 2018-02-14
Remove reqwest library

JustinPinner:
## Why are you doing this?
In pursuit of removing old, unsupported code libraries, we've identified Reqwest as one of those, having last been updated in December 2015, by our very own Oliver J Ash as it happens.

While posing no immediate risk to security or stability, membership-frontend is considered a legacy codebase now, so we're seeking to minimise the effort likely to be required to maintain it in the future.

We've already dropped jQuery in to replace Bonzo and Qwery (see https://github.com/guardian/membership-frontend/pull/1776) and now we're going to use jQuery's ajax support to replace Reqwest. Assuming the automated and subsequent manual tests all check out of course. They have locally, but we'll see... 🤞 

Why jQuery? Primarily because it still has an active user community and is therefore a good candidate for security weaknesses to be detected and fixed. And because it appears to offer the functionality we need, the tradeoff being a slightly larger code payload for browsers.

In the course of making this swap, I found some code that seems to be unused, and as it included an ajax call, figured we should get rid of it if we can't test it. If you happen to know otherwise, and more importantly _how to test it_, please leave some comments?

## Changes
* We always wrapped calls to Reqwest under our own ajax alias, so now we use that same alias to direct calls to jQuery.ajax instead. There were some minor differences between the Reqwest and jQ ajax requests so we take care of those in our wrapper.
* Removed some unused (TBC) credit card-related code that also made use of ajax requests. Nothing in the current test suite or upgrade paths appears to execute that code so it's gone, pending confirmation that everything still works. These have been removed:
  * `cardDetail.scala.html`
  * `form.hasPayment` in `form.js`
  *  `hasPayment` in `formUtil.js`
  *  `processing.js`
  * `form.hasPayment` in `listeners.js`

## Screenshots
**Tested on CODE**

I tested creating new supporter accounts with card and paypal, then upgraded these through each of the next level tiers. I didn't capture screens of all of those stages (they all worked though) but here's what I did capture; 

**Upgrade from Partner to Patron (Card)**
screen shot 2018-02-14 at 14 25 44 upgraded to patron card

**Upgrade from Supporter to Partner (PayPal)**
screen shot 2018-02-14 at 14 39 41 upgraded to partner paypal

**Upgraded from Partner to Patron (PayPal)**
screen shot 2018-02-14 at 14 42 36 upgraded to patron paypal


frontend/app/views/fragments/form/cardDetail.scala.html
```diff
@@ -1,179 +0,0 @@
...
-                </script>
-            </div>
-            <div class="bottom-credit-card-description--note">
-                <h3 class="credit-card-note__header">Cancellation</h3>
-                <p class="credit-card-note__text">You can cancel your membership online through your profile page.</p>
-            </div>
-        </div>
-
-    </div>
-</fieldset>
```
Ap0c: It's odd that this isn't in use. @AWare is it possible that this was the form used prior to the introduction of Stripe Checkout?

@JustinPinner with this file gone is it possible that [this listeners file](https://github.com/guardian/membership-frontend/blob/3e93a9826d3091c32059cf84e45fb472eea7c268/frontend/assets/javascripts/src/modules/form/payment/listeners.js) is also now redundant?

frontend/assets/javascripts/src/modules/form/payment/processing.js
```diff
@@ -1,119 +0,0 @@
-/*global Stripe */
-define([
-    'ajax',
-    'src/utils/helper',
-    'src/modules/form/helper/formUtil',
-    'src/modules/form/helper/serializer',
-    'src/modules/form/helper/loader',
-    'config/paymentErrorMessages',
```
Ap0c: It seems like this processing file is the only place [this error message module](https://github.com/guardian/membership-frontend/blob/791eb801143f174764dc6eccabfe3523f2e28d38/frontend/assets/javascripts/config/paymentErrorMessages.js) is imported, maybe we can throw that out too?

frontend/app/views/fragments/form/cardDetail.scala.html
jacobwinch: I would also guess pre-Stripe checkout - perhaps @AWare or @paulbrown1982 can confirm as they're probably the most familiar with this code?

frontend/assets/javascripts/src/modules/paidToPaid.es6
```diff
@@ -41,27 +41,28 @@ export function init() {
...
             type: 'json',
             url: window.location.pathname.replace('change', 'preview'),
-            data: {promoCode: PROMO_CODE.val()}
-        }).then(result => {
-            LOADER.removeClass('is-loading');
-
-            if (result.error) {
-                FEEDBACK_CONTAINER.html(promoError({errorMessage: result.error}));
-                return;
+            data: {promoCode: PROMO_CODE.val()},
```
jacobwinch: @paulbrown1982 - is this promo code stuff still needed?

frontend/app/views/fragments/form/cardDetail.scala.html
paulbrown1982: This was the old form. It should be deleted.

frontend/assets/javascripts/src/modules/paidToPaid.es6
paulbrown1982: Not in membership

frontend/assets/javascripts/src/modules/form/payment/processing.js
jacobwinch: It looks like the PaymentGatewayErrors have been duplicated here: https://github.com/guardian/membership-frontend/blob/8e6779579a2e5758c7ef103c77dd2b200a9bd18f/frontend/assets/javascripts/src/modules/form/payment/paymentError.es6

jacobwinch: And presumably the errors in the 'card_error' block are now unnecessary, as Stripe Checkout handles this first stage of card validation.

jacobwinch: So I agree that we can probably remove that dead code too.

JustinPinner:
Thanks for the comments everyone. I'm not intending to pull any more dead code out with this particular PR but I'll create some Trello cards somewhere to capture these additional observations. I'm on holiday tomorrow so will probably merge this on Monday morning next week, in case anyone else has any other comments to add between now and then.

PR #1786, opened 2018-02-15
Remove Sentry Tags with (soon to be) PII and stop logging bad requests incorrectly

jacobwinch:
## Why are you doing this?
This is the first in a series of PRs which aims to improve/sanitize our Sentry logging, in order to help us to attain GDPR compliance.

As a bonus, this PR also fixes a problem which meant that we were logging bad requests as internal server errors. 

## Changes
* Remove tags which contain ids, as we no longer want to send this information to Sentry
* Log bad requests as a warning (note: this means that they will no longer be sent to Sentry)

### DIFF
--- a/frontend/app/monitoring/ErrorHandler.scala
+++ b/frontend/app/monitoring/ErrorHandler.scala
@@ -2,19 +2,14 @@ package monitoring
 
 import java.lang.Long
 import java.lang.System.currentTimeMillis
-
-import com.gu.googleauth.UserIdentity
+import com.typesafe.scalalogging.StrictLogging
 import controllers.{Cached, NoCache}
-import monitoring.SentryLogging.{UserGoogleId, UserIdentityId}
-import org.slf4j.MDC
 import play.api._
 import play.api.http.DefaultHttpErrorHandler
 import play.api.mvc.Results._
 import play.api.mvc._
 import play.api.routing.Router
-import services.AuthenticationService
 import play.core.SourceMapper
-
 import scala.concurrent._
 
 class ErrorHandler(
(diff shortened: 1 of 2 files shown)


frontend/app/monitoring/ErrorHandler.scala
```diff
@@ -23,17 +18,7 @@ class ErrorHandler(
   sourceMapper: Option[SourceMapper],
   router: => Option[Router],
   implicit val executionContext: ExecutionContext
-) extends DefaultHttpErrorHandler(env, config, sourceMapper, router) {
-
-  override def logServerError(request: RequestHeader, usefulException: UsefulException) {
```
jacobwinch: I don't think we need this override now that we aren't sending these tags on...

frontend/app/monitoring/SentryLogging.scala
```diff
@@ -13,10 +13,7 @@ import scala.util.{Failure, Success, Try}
 
 object SentryLogging {
 
-  val UserIdentityId = "userIdentityId"
-  val UserGoogleId = "userGoogleId"
-  val PlayErrorId = "playErrorId"
-  val AllMDCTags = Seq(UserIdentityId, UserGoogleId,PlayErrorId)
+  val AllMDCTags = Seq()
```
lmath: Why not just not call setExtraTags?

jacobwinch: That's a very good point...

jacobwinch: 🤦‍♂️

PR #1789, opened 2018-02-16
Use SafeLogger

jacobwinch:
## Why are you doing this?
I am migrating this project over to using a new SafeLogger object (imported [here](https://github.com/guardian/membership-frontend/pull/1787)), with the ultimate aim of keeping PII out of Sentry.

Note that error messages won't actually be filtered out until the next PR is finished. The (final PR) will add the filters to the Sentry appender and override Play's default error handling.

## Changes
* Stop inheriting logging traits, and use SafeLogger object (from membership-common) instead
* Simplify Salesforce vs Members-Data-API comparison, and remove metrics

PR #1791, opened 2018-02-19
Only compare tier information when checking members-data-api response

jacobwinch:
## Why are you doing this?
This [change](https://github.com/guardian/membership-frontend/pull/1789/files/0e46c0cb47da9d0e7addda9803dd719e2fb457c0#diff-c978671788c98af3d5b98e03e2f6e252) is generating a lot of noise, because we have accidentally started logging errors when members-data-api doesn't return a membership number.


johnduffell:
yep good good! 👍 🎈 
I'm beginning to wonder whether ***** was the right word to use for secrets, should have used @%^&*% 😂

PR #1796, opened 2018-03-07
Remove bean and use jquery events instead

JustinPinner:
## Why are you doing this?
As part of our ongoing commitment to remove unsupported/outdated and/or insecure libraries, we're removing Bean on the grounds that it falls into the former category.

See also: https://github.com/guardian/membership-frontend/pull/1776
and: https://github.com/guardian/membership-frontend/pull/1785 

## Changes
* Remove references to the Bean library
* Switch over to jQuery's `.on()` and `.off()` methods instead
 
## Screenshots
beanfree

## Tested on CODE?
Yes. Have tested signing up from friend -> supporter -> partner -> patron and finally -> cancel: ALL OK.

Also, the build tests all passed too, obvs.

### DIFF
--- a/frontend/assets/javascripts/src/modules/form/stripe.es6
+++ b/frontend/assets/javascripts/src/modules/form/stripe.es6
@@ -1,11 +1,10 @@
 import * as payment from 'src/modules/payment';
-import bean from 'bean'
 import $ from '$'
 export function init() {
     let handler = window.StripeCheckout.configure(guardian.stripeCheckout);
     let success = false;
     const button = $('.js-stripe-checkout');
-    bean.on(window, 'popstate', handler.close);
+    $(document).on('popstate', handler.close);
     const amount = () => {
 
         let billingPeriod = guardian.membership.checkoutForm.billingPeriods[guardian.membership.checkoutForm.billingPeriod];
@@ -42,5 +41,5 @@ export function init() {
         e.preventDefault();
     };
 
-    bean.on(button[0], 'click', open);
+    $(button).on('click', open);
 }
--- a/frontend/assets/javascripts/src/modules/form/stripe.es6
+++ b/frontend/assets/javascripts/src/modules/form/stripe.es6
@@ -1,10 +1,10 @@
 import * as payment from 'src/modules/payment';
-import $ from '$'
+import $ from '$';
 export function init() {
     let handler = window.StripeCheckout.configure(guardian.stripeCheckout);
     let success = false;
     const button = $('.js-stripe-checkout');
-    $(document).on('popstate', handler.close);
+    window.addEventListener('popstate', handler.close);
     const amount = () => {
 
         let billingPeriod = guardian.membership.checkoutForm.billingPeriods[guardian.membership.checkoutForm.billingPeriod];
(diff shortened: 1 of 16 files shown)


paulbrown1982:
👍 Ender would be proud of his [Jeesh](https://en.wikipedia.org/wiki/List_of_Ender%27s_Game_characters)mate's tenure.

frontend/assets/javascripts/src/modules/form/stripe.es6
```diff
@@ -1,11 +1,10 @@
 import * as payment from 'src/modules/payment';
-import bean from 'bean'
 import $ from '$'
```
Ap0c: `;`?

frontend/assets/javascripts/src/modules/form/stripe.es6
```diff
@@ -1,11 +1,10 @@
...
 export function init() {
     let handler = window.StripeCheckout.configure(guardian.stripeCheckout);
     let success = false;
     const button = $('.js-stripe-checkout');
-    bean.on(window, 'popstate', handler.close);
+    $(document).on('popstate', handler.close);
```
Ap0c: What's the difference between listening for this event on the `window` vs the `document`?

frontend/assets/javascripts/src/modules/form/stripe.es6
```diff
@@ -42,5 +41,5 @@ export function init() {
         e.preventDefault();
     };
 
-    bean.on(button[0], 'click', open);
+    $(button).on('click', open);
```
Ap0c: Isn't `button` already a jquery object? Does this work if it's just:

```js
button.on('click', open);
```

frontend/assets/javascripts/src/modules/identityMenu/identityPopup.js
```diff
@@ -24,22 +24,22 @@ define([
     }
 
     function disableLink() {
-        bean.on(IDENTITY_MENU_CTA_ELEM, 'click', function(e) {
+        IDENTITY_MENU_CTA_ELEM.on('click', function(e) {
```
Ap0c: If this is the only place `bean` is used in this file, could we just do `$(IDENTITY_MENU_CTA_ELEM)` here, and leave the native browser methods intact everywhere else?

JustinPinner:
I've introduced a form validation error here somewhere, so I'll need to revisit on CODE prior to merging this.

Steps to reproduce the error;

1. Make sure you're signed out
2. Choose 'Become a supporter'
3. During the address-filling phase, tab over a required field without adding anything
4. Observe the error state in the form
5. Go back and correct the missing data
6. Tab through or click on the Continue button
7. The form states there are still errors

If you complete the form correctly, as the tests do, it'll be fine so it's only when you make a mistake that it's a problem. This behaviour is not present in PROD.

JustinPinner:
The problem described above is fixed. The solution was to flush the accordian-styled details form to reassess which fields require validation when the continue button is pressed. All the form validation tests still pass so it seems like a reasonable thing to do.

I've also manually re-tested on code the progression from supporter via partner to patron and then cancelling without any problems.

JustinPinner:
Damned PayPal tests - they fail often. Am in the process of checking manually anyway.

PR #1801, opened 2018-03-19
Less monitoring

davidfurey:
## Why are you doing this?

Spend less money on cloudwatch metrics by sending fewer metrics.


frontend/app/configuration/Config.scala
```diff
@@ -42,6 +42,8 @@ object Config {
 
   val idMember = "clientId" -> "members"
 
+  val sendJVMMetrics = Try { config.getBoolean("send-jvm-metrics") }.toOption.contains(true)
```
johnduffell: excellent idea giving it a switch

johnduffell:
was it expensive?
either way seems fair enough, I think a lot came from the boxes recycling issues we were having a few times in the past so it's possiible we'll need in future
👍 🚀

frontend/app/configuration/Config.scala
jacobwinch: 👍

davidfurey:
@johnduffell - in February cloudwatch made up 33% of our bill.  I don't know how to break it down by metric or service, but membership-frontend has the most metrics so it seemed like a good place to start.

johnduffell:
it's the requests that are expensive by the looks not the number of metrics by 3:1.  we're doing nearly 2 million a day, I wonder if members-data-api is pushing a value on every request?

johnduffell:
just checked and it looks like members data api is fine, you're right most of the samples are coming from membership-frontend! excellent work! 👍

johnduffell:
Although the frontend seems to be working fine, we're getting a lot of errors around 6000 per hour in kibana starting at that moment: "CloudWatch PutMetricDataRequest error: The parameter Namespace is required. The parameter MetricData.member.1.Dimensions.member.1.Value is required. (Service: AmazonCloudWatch; Status Code: 400; Error Code: MissingParameter; Request ID: 2051ddf9-2dfa-11e8-8fd3-211709018a51)}"

PR #1802, opened 2018-03-19
Revert "Reverted cloud formation changes"

davidfurey:
Reverts guardian/membership-frontend#1746

Do not merge until https://github.com/guardian/membership-frontend/pull/1757#pullrequestreview-105024820 is deployed


svillafe:
It is worth to check if we will have a similar NTP issue after we merge this: https://github.com/guardian/support-frontend/pull/341

davidfurey:
Good point.  I've changed membership-xenial to use chrony, so this shouldn't be a problem.  I'll confirm once I've merged.

PR #1804, opened 2018-03-22
Use a long timeout when fetching the catalog

jacobwinch:
## Why are you doing this?
A recent incident with members-data-api alerted us to a potential risk concerning the startup for this app. If the catalog cannot be loaded on startup, a failed future is kept, which causes errors downstream. This is a quick fix to decrease the likelihood of the request to get the catalog failing.

## Changes
* Use a 60 second timeout when loading the catalog.


johnduffell:
thanks! 👍 🚀 
this is important because if any instances recycle while the latency is high, we will be unable to replace them, possibly overnight.

PR #1806, opened 2018-03-29
Add extra logging to investigate user query

davidfurey:
## Why are you doing this?

To help investigate case 61178/00408687


johnduffell:
👍

PR #1807, opened 2018-04-10
Replace curl-amd with dotcom's loadScript

JustinPinner:
I _think_ that's everything done now. Enjoy!

## Why are you doing this?
In pursuit of removing old, unsupported code, we targeted curl-amd as a library that falls into this category.

While there's nothing (currently known about) wrong with it at the moment, if any GDPR-sensitive issues are discovered, we'd need to quickly remove and replace it. Also, as membership-frontend's journey into legacy status continues, we'd prefer to tackle it now.

To replace it, we've dropped in the same code used by dotcom to handle external script files.

This PR started out as a spike but to investigate the feasibility of changing each dependency properly, it turned into a working solution.

## Changes
* The `curl-amd` loader is gone
* Some `.js` files have become `.es6` files
* The `loadScript.es6` file is a direct copy from dotcom

## Tested on CODE
Yes. I've checked that the video player works, the PayPal and Stripe checkouts behave as expected and that the tracking calls we make to krux, bing(!), google and facebook all look sensible.

I've also checked the full upgrade path from supporter -> partner -> patron -> cancel using both PayPal and Stripe payment methods, and checked that the Zuora UAT records all look correct at each stage.

Events and masterclasses still work as expected, including discounts for partners and patrons, and the links through to Eventbrite seem to be fine too. 

## Screenshots
N/A (it all looks the same)


frontend/app/views/fragments/javaScriptLaterSteps.scala.html
```diff
@@ -4,15 +4,15 @@
...
+    var sideLoad = {
         paths: {
             stripe: 'https://js.stripe.com/v2/stripe-dss3',
             uet: '@Asset.at("javascripts/lib/uet/uet.js")',
-            paypal: 'https://www.paypalobjects.com/api/checkout.js',
-            stripeCheckout: 'https://checkout.stripe.com/checkout.js'
+            paypal: '//www.paypalobjects.com/api/checkout.js',
+            stripeCheckout: '//checkout.stripe.com/checkout.js',
+            krux: '//cdn.krxd.net/controltag',
+            youTubeApi: '//www.youtube.com/iframe_api?noext'
```
Ap0c: Why does stripe keep the protocol but the others don't?

frontend/assets/javascripts/src/modules/form.es6
```diff
@@ -51,61 +51,52 @@
...
-            if (form.hasStripeCheckout) {
-
-                curl('js!stripeCheckout')
-                    .then(function () {
-                        stripe.init();
-                    });
-            }
+import validation from 'src/modules/form/validation';
+import form from 'src/modules/form/helper/formUtil';
+// import payment from 'src/modules/form/payment'; // TODO Can this be deleted?
```
Ap0c: Because it appears to be unused in this file? That's weird...

PR #1812, opened 2018-04-25
Allow non-Members to go straight through to EventBrite rather than see a sign-up screen

paulbrown1982:
Currently only logged in Members can buy tickets to [Live Events](https://membership.theguardian.com/events). Partner and Patron members get a promo code when linking through for up to 2 free tickets.

This change lets any reader through to the standard EventBrite page, rather than bringing up a 'Become a member', with no discounted tickets that they can go on to buy through EventBrite. The logic does not change if they are logged in as a Partner/Patron.

## Why are you doing this?

To increase the number of tickets being sold.

cc @jacobwinch @johnduffell

### DIFF
--- a/frontend/app/controllers/Event.scala
+++ b/frontend/app/controllers/Event.scala
@@ -134,17 +134,12 @@ class Event(
   }
 
   def buy(id: String): Action[AnyContent] = eventbriteService.getEvent(id) match {
-      case Some(event@(_: GuLiveEvent)) =>
-        // Logged out readers and non-members will go via the onUnauthenticated route
-        // They will get a "SOLD OUT" EventBrite page if the even only has membership tickets.
-        BuyAction(id, onUnauthenticated = redirectAnonUserToEventbrite(event)(_)).async { implicit request =>
-          // Only logged in Members come here. If they are a Friend and the event is only for paid members
-          // we will redirect to the upgrade screen rather than to EventBrite with or without a promo code.
-          if (event.isBookableByTier(request.subscriber.subscription.plan.tier))
-            redirectSignedInMemberToEventbrite(event)
-          else suggestUserUpgrades
-        }
-      case Some(event@(_: MasterclassEvent)) =>
+      // One conditional now covers Live Events AND Masterclasses...
+
+      // Logged out readers and non-members will go via the onUnauthenticated route
+      // They, and any ineligible-tier Members will get a "SOLD OUT" EventBrite page if the event has membership-only tickets.
+      // Previously ineligible-tier Members would get a suggested upgrade page.
+      case Some(event@(_: RichEvent)) =>
         BuyAction(id, onUnauthenticated = redirectAnonUserToEventbrite(event)(_)).async { implicit request =>
           redirectSignedInMemberToEventbrite(event)
         }
@@ -180,7 +175,7 @@ class Event(
     Found(addEventBriteGACrossDomainParam(eventUrl)).withCookies(Cookie(eventCookie(event), "", Some(3600)))
   }
 
-
+  @deprecated("Guardian Live Events are not configured to use this anymore, and host their own thankyou page on EventBrite.")
   def thankyou(id: String, orderIdOpt: Option[String]) = SubscriptionAction.async { implicit request =>
     orderIdOpt.fold {
       val resultOpt = for {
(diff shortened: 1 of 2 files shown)


frontend/app/controllers/Event.scala
```diff
@@ -135,7 +135,11 @@ class Event(
 
   def buy(id: String): Action[AnyContent] = eventbriteService.getEvent(id) match {
       case Some(event@(_: GuLiveEvent)) =>
```
paulbrown1982: I'm just testing whether to collapse this down with the MasterClasses event to be something like:

`case Some(event@(_: LiveEvent)) =>`

Once I've found a advance-booking event I'll comment.

frontend/app/controllers/Event.scala
```diff
@@ -134,13 +134,12 @@ class Event(
...
-        }
-      case Some(event@(_: MasterclassEvent)) =>
+      // One conditional now covers Live Events AND Masterclasses...
+
+      // Logged out readers and non-members will go via the onUnauthenticated route
+      // They, and any ineligible-tier Members will get a "SOLD OUT" EventBrite page if the event has membership-only tickets.
+      // Previously ineligible-tier Members would get a suggested upgrade page.
+      case Some(event@(_: RichEvent)) =>
         BuyAction(id, onUnauthenticated = redirectAnonUserToEventbrite(event)(_)).async { implicit request =>
           redirectSignedInMemberToEventbrite(event)
```
jacobwinch: Are _redirectAnonUserToEventbrite_ and _redirectAnonUserToEventbrite_ still sensible names for these functions?

jacobwinch: Given your comment it feels like these should be _redirectEligibleSignedInUserToEventbrite_ and _redirectIneligibleUserToEventbrite_ (or something along those lines)?

frontend/app/controllers/Event.scala
```diff
@@ -134,13 +134,12 @@ class Event(
   }
 
   def buy(id: String): Action[AnyContent] = eventbriteService.getEvent(id) match {
-      case Some(event@(_: GuLiveEvent)) =>
-        BuyAction(id).async { implicit request =>
-          if (event.isBookableByTier(request.subscriber.subscription.plan.tier))
-            redirectSignedInMemberToEventbrite(event)
-          else suggestUserUpgrades
```
jacobwinch: are the suggestUserUpgrades and isBookableByTier functions still needed?

frontend/app/controllers/Event.scala
```diff
@@ -134,13 +134,12 @@ class Event(
...
-        }
-      case Some(event@(_: MasterclassEvent)) =>
+      // One conditional now covers Live Events AND Masterclasses...
+
+      // Logged out readers and non-members will go via the onUnauthenticated route
+      // They, and any ineligible-tier Members will get a "SOLD OUT" EventBrite page if the event has membership-only tickets.
```
jacobwinch: where is the logic which checks for ineligible tier happening?

paulbrown1982: Technically, it's in the comment of the description field on the eventbrite page. See the 'ticketing' val in Eventbrite.scala

paulbrown1982: And TicketSaleDates.scala

frontend/app/controllers/Event.scala
paulbrown1982: We can delete the former, the latter is used in DestinationService.scala for the deeplinked event purchase page.

frontend/app/controllers/Event.scala
jacobwinch: Sorry, I misread this, feel free to ignore these comments.

paulbrown1982: How wbout redirectMemberToEventBrite and redirectNonMemberToEventBrite ?

jacobwinch: Yes, perhaps that's a bit clearer 👍

frontend/app/controllers/Event.scala
jacobwinch: Yep, sorry I just pulled your branch and found that out (hence the edited comment!)

frontend/app/controllers/Event.scala
```diff
@@ -176,7 +172,7 @@ class Event(
     Found(addEventBriteGACrossDomainParam(eventUrl)).withCookies(Cookie(eventCookie(event), "", Some(3600)))
   }
 
-
+  @deprecated("Guardian Live Events are not configured to use this anymore, and host their own thankyou page on EventBrite.")
```
johnduffell, line 175: I noticed a message I think it was in chat or email saying we are not limiting the number of tickets people buy, which is affecting paid ticket sales.  I noticed this is because this ThankYou page was writeing in zuora that a free ticket had been used.  Now when we query zuora everyone always has a full complement of free tickets as this isn't called any more.

PR #1813, opened 2018-04-25
Use default CSRF settings and contentType whitelist

JustinPinner:
## Why are you doing this?
To paraphrase the Trello card;
> The upgrade of Play on membership-frontend left the CSRF restrictions the same as 2.4. It would be good to remove this more lax behaviour and go with the default Play 2.6 behaviour

By removing _all_ of the older config settings, loading the PayPal checkout window returned a `403 Forbidden` response due to the content type. Reapplying just the `contentType.blackList` settings, everything seems to work just fine in tests (locally and during the build) for PayPal and Stripe payments.

However, even better is to specify a `contentType.whiteList` to restrict the types of data we expect to handle. These are currently `application/json` and `text/html`

### DIFF
--- a/frontend/conf/application.conf
+++ b/frontend/conf/application.conf
@@ -7,19 +7,9 @@ play.http.session.secure=true
 
 play.http.errorHandler = "monitoring.ErrorHandler"
 
+play.filters.enabled += "play.filters.csrf.CSRFFilter"
+
 play.filters.csrf {
-    header {
-        bypassHeaders {
-            X-Requested-With = "*"
-            Csrf-Token = "nocheck"
-        }
-        protectHeaders = null
-    }
-    bypassCorsTrustedOrigins = false
-    method {
-        whiteList = []
-        blackList = ["POST"]
-    }
     contentType.blackList = ["application/x-www-form-urlencoded", "multipart/form-data", "text/plain"]
 }
 
--- a/frontend/conf/application.conf
+++ b/frontend/conf/application.conf
@@ -7,10 +7,8 @@ play.http.session.secure=true
 
 play.http.errorHandler = "monitoring.ErrorHandler"
 
-play.filters.enabled += "play.filters.csrf.CSRFFilter"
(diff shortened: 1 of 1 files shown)


rupertbates:
What is the content-type that PayPal uses? We could whitelist that rather than using the blacklist?

JustinPinner:
I'll try adding it and re-testing in the morning

JustinPinner:
Won't that whitelist _only_ application/json though?

frontend/conf/application.conf
```diff
@@ -7,19 +7,9 @@ play.http.session.secure=true
 
 play.http.errorHandler = "monitoring.ErrorHandler"
 
+play.filters.enabled += "play.filters.csrf.CSRFFilter"
```
davidfurey: Does this filter not default to on?

rupertbates:
Yes but if that's the only one thats causing a problem then that should do the trick. I think in principal it's better to whitelist the types you need and check everything else rather than only checking a set list.

frontend/conf/application.conf
JustinPinner: From 2.6 yes. Removed.

JustinPinner:
From https://www.playframework.com/documentation/2.6.x/resources/confs/filters-helpers/reference.conf

```    # Content type lists
    # If both white lists and black lists are empty, then all content types are checked.
    contentType {
      # If non empty, then requests will be checked if the content type is not in this list.
      whiteList = []

      # The black list is only used if the white list is empty.
      # Only check content types in this list.
      blackList = []
    }```

JustinPinner:
This appears to be fine on CODE @rupertbates and @davidfurey  

I manually tested the paths for supporter ➔ partner ➔ patron ➔ cancel, for stripe and supporter ➔ partner ➔ cancel for paypal. The tickets/discounts options for events and masterclasses all worked as expected, and the Zuora records were all fine too.

PR #1816, opened 2018-05-16
Prepare membership-frontend for SSM instead of SSH

JustinPinner:
## Why are you doing this?
We want to transition to using SSM instead of our existing SSH setup on our instances. This change adds the cloudformation policy to bring in our slightly less privileged `guardian-ec2-role-for-ssm` as used on support-frontend and (I think) contributions.

Once we know the new policy allows us to `ssm ssh --tags frontend,membership,PROD --profile membership` and then not have access to anything we shouldn't, we're good to go.

## Changes
* Adds a new `ManagedPolicyArns` to `MembershipAppRole`
* That's it!

## Screenshots
Not really screenshots, but...
```
ssm ssh --tags frontend,membership,CODE --profile membership
========= i-0a69e524194c182b9 =========
STDOUT:
```

so far so good, now to try

```
aws s3 ls s3://gu-support-logs

An error occurred (AccessDenied) when calling the ListObjects operation: Access Denied
```

Excellent!


jacobwinch:
Looking in more detail at how the existing role works (see https://amigo.gutools.co.uk/roles#s3-ssh-keys), it looks like we can also delete the following S3 permissions, which are presumably only used to keep GitHub ssh keys up to date: https://github.com/guardian/membership-frontend/blob/master/cloud-formation/membership-app.cf.yaml#L142-L151

JustinPinner:
Thanks @jacobwinch. That's part two of this task. 1: Make sure SSM works. 2: Remove SSH :)

JustinPinner:
That close and comment is too big. Even though it's not green.

johnduffell:
great look sgood! thanks. 👍 🔒

jacobwinch:
👍

PR #1829, opened 2018-06-29
Now points to support platform

svillafe:
## Why are you doing this?
To stop selling membership

## Changes
* UPdated the become a supporter link from the Nav

## Screenshots
N/A

### DIFF
--- a/frontend/app/model/Nav.scala
+++ b/frontend/app/model/Nav.scala
@@ -19,7 +19,7 @@ object Nav {
   }
 
   val primaryNavigation = List(
-    NavItem("become a supporter", "https://support.theguardian.com?INTCMP=gdnwb_copts_memco_internal_supporter_link", "Become a supporter"),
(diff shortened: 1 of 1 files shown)


svillafe:
Tested in CODE ✅

frontend/app/model/Nav.scala
```diff
@@ -19,7 +19,7 @@ object Nav {
   }
 
   val primaryNavigation = List(
-    NavItem("become a supporter", "/supporter?INTCMP=gdnwb_copts_memco_internal_supporter_link", "Become a supporter"),
+    NavItem("become a supporter", "https://support.theguardian.com?INTCMP=gdnwb_copts_memco_internal_supporter_link", "Become a supporter"),
```
joelochlann: shall we change replace the `INTCMP` with an `acquisitionData` param? https://support.theguardian.com/?acquisitionData=%7B%22source%22%3A%22GUARDIAN_WEB%22%2C%22componentId%22%3A%22become_a_supporter_link_on_membership_site%22%7D

svillafe: Updated 👍  thanks for generating it

PR #1838, opened 2018-07-20
Upgrade bcprov

jranks123:
Due to a Synk vulnerability, we are upgrading the `bcprov` library, which is a transitive dependency of `play-googleauth`. Identity are upgrading `play-googleauth`, but they are no longer compiling against scala `2.11`, so this repo will need to be upgraded to receive future security updates


johnduffell:
thanks for the PR
I have had a look at what the identity guys have been up to: https://github.com/guardian/identity-play-auth/pull/22
version 2.5 is out, so I'm going to update this PR to use it instead of patching

johnduffell:
scratch that - the new version is scala 2.12 only and it's not a trivial job to update (as we'd have to update a lot of libraries)

PR #1842, opened 2018-08-20
Make it easier to debug failures to load the catalog

jacobwinch:
## Why are you doing this?
https://trello.com/c/7m3LobBn/26-retro-action-improve-logging-when-we-fail-to-load-the-catalog

### DIFF
--- a/frontend/app/services/TouchpointBackend.scala
+++ b/frontend/app/services/TouchpointBackend.scala
@@ -20,12 +20,12 @@ import com.gu.zuora.soap.ClientWithFeatureSupplier
 import com.gu.zuora.{soap, ZuoraSoapService => ZuoraSoapServiceImpl}
 import com.netaporter.uri.Uri
 import com.gu.monitoring.SafeLogger
+import com.gu.monitoring.SafeLogger._
 import configuration.Config
 import model.FeatureChoice
 import play.api.libs.ws.WSClient
 import play.api.mvc.RequestHeader
 import utils.TestUsers.{TestUserCredentialType, isTestUser}
-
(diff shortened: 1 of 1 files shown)


frontend/app/services/TouchpointBackend.scala
```diff
@@ -70,7 +70,15 @@ object TouchpointBackend {
...
     val catalogRestClient: SimpleClient[Future] = new SimpleClient[Future](restBackendConfig, RequestRunners.configurableFutureRunner(60.seconds))
     val newCatalogService = new subsv2.services.CatalogService[Future](pids, FetchCatalog.fromZuoraApi(catalogRestClient), Await.result(_, 60.seconds), restBackendConfig.envName)
-    val futureCatalog: Future[CatalogMap] = newCatalogService.catalog.map(_.fold[CatalogMap](error => {println(s"error: ${error.list.toList.mkString}"); Map()}, _.map))
+
+    val futureCatalog: Future[CatalogMap] = newCatalogService.catalog
+      .map(_.fold[CatalogMap](error => {println(s"error: ${error.list.toList.mkString}"); Map()}, _.map))
+      .recover {
+        case error =>
+          SafeLogger.error(scrub"Failed to load catalog from Zuora due to: $error")
+          throw error
```
pvighi: this change is good, I just don't understand what the existing println was doing in there

johnduffell: looks like an "if not populated yet, println it and return empty" kind of hack.  Not sure if it was supposed to be shipped in that state though.

johnduffell:
thanks ! 👍

PR #1843, opened 2018-08-21
Update post-deployment tests

jacobwinch:
## Why are you doing this?
These tests are currently broken due to changes to the default Identity flow. (See a similar PR here: https://github.com/guardian/support-frontend/pull/789). 

## Changes
* Change Selenium tests to cope with two-step identity flow.
* Update test dependencies


Ap0c:
Nice! 🚀

PR #1850, opened 2018-10-01
membership frontend no longer sends emails via sqs

QuarpT:
## Why are you doing this?

The sqs trigger thank you email was unused. It seems this is now done by support workers. 
I came across this while working on the CMT migration, searching https://github.com/search?q=org%3Aguardian+SubscriberAttributes&type=Code

## Changes
* Remove unused thank you email sqs trigger


jacobwinch:
How does this app send thankyou emails @paulbrown1982? Do they come from Salesforce?

johnduffell:
yes thanks for tidying up! 👍 
And it seems that tier change (including from blank) emails come from SF, we discovered that when we did the SF contact merging and sent our thousands.  When the membership tier field is editied it happens automatically.

PR #1854, opened 2018-12-05
run yarn upgrade grunt

pvighi:
## Why are you doing this?
Our lock file was including a dependency that had ahigh severity vulnerability according to  Snyk

I just tested creating a supporter UAT in Chrome. 
I'm not sure a change like this will require testing in more different clients
@AWare @Ap0c @paulbrown1982 @twrichards


frontend/yarn.lock
```diff
@@ -8039,8 +8074,8 @@ undefsafe@^2.0.0:
     debug "^2.2.0"
 
 underscore.string@~3.3.4:
-  version "3.3.4"
```
pvighi: it looks like this is the only change we need but I don't know of way of getting yarn to only upgrade this dependency

twrichards: did you try committing only this line (and the one below it) to see if it builds in TeamCity - I personally think thats the less risky / lighter touch change.

pvighi: Yeah after talking to you I was going to do that but I read the top of the file that says in all caps "THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY". 
Then I talked to @Ap0c which got a different idea from the server side meeting where apparently this kind of issues were discussed.
So my plan was to tag all the client side people I can think of in this PR and see what the majority thinks.

svillafe:
I think if we run `yarn upgrade --latest` it will update the `package.json` too. I don't know if we would like to do that though.

johnduffell:
thanks for updating! no idea whether it will break things but it's worth a try! 👍

PR #1855, opened 2018-12-07
Upgrade to Scala 2.12

jacobwinch:
## Why are you doing this?

This project depends on a couple of libraries which have now dropped 2.11 support. This PR brings the project up to date, which allows us to pick up: guardian/membership-common#585

This is required so that this app can continue to load the catalog after some planned changes.

## Changes
* Upgrade to Scala 2.12
* Upgrade various libraries
* Remove retry wrappers, since dispatch was dropped by: https://github.com/guardian/membership-common/pull/584

### DIFF
--- a/frontend/app/services/IdentityService.scala
+++ b/frontend/app/services/IdentityService.scala
@@ -6,12 +6,8 @@ import cats.syntax.either._
 import com.gu.identity.play._
 import com.gu.identity.play.idapi.{CreateIdUser, UpdateIdUser, UserRegistrationResult}
 import com.gu.memsub.Address
-import com.gu.memsub.util.Timing
-import com.gu.monitoring.SafeLogger
 import configuration.Config
 import controllers.IdentityRequest
-import dispatch.Defaults.timer
-import dispatch._
 import forms.MemberForm._
 import com.gu.monitoring.SafeLogger
 import play.api.http.Status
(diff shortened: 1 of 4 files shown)


frontend/app/services/IdentityService.scala
```diff
@@ -103,9 +99,7 @@ class IdentityService(identityApiProvider: => IdentityApi, implicit val ec: Exec
   }).valueOr(_ => false)
 
   def getFullUserDetails(user: IdMinimalUser)(implicit identityRequest: IdentityRequest): Future[IdUser] =
-    retry.Backoff(max = 3, delay = 2.seconds, base = 2){ () =>
-      getUser(user).value
-    }.map(_.fold({ errorMessage =>
+    getUser(user).value.map(_.fold({ errorMessage =>
```
pvighi: do you know if there was any reason these calls had a retry ?

jacobwinch: Presumably both of these calls were often unreliable at some stage. Given that the retries were added some time ago, and this app is now much lower traffic, I'm prepared to remove these and then keep an eye on how things fare in production unless anyone strongly disagrees?

pvighi: I think it's fine but I would wait to merge this on Monday

jacobwinch: Agreed 👍

johnduffell: oh yes sorry I found these deps via snyk and clearly didn't actully push them out.  Thanks for doing that. I found by following the deps that we weren't actually using the vulnerable library, so I wasn't concerned about an actual security issue.

johnduffell:
oh yeThanks 👍 💯

jacobwinch:
^^^ This is due to the PayPal test (which is known to be unstable) failing... I'm going to remove that flakey test.

jacobwinch:
https://github.com/guardian/membership-frontend/pull/1856

PR #1856, opened 2018-12-10
Remove PayPal test

jacobwinch:
## Why are you doing this?

This test consistently fails even though the flow is stable. Investigating these failures is a waste of our time, and since we are intending to deprecate this old flow anyway, I don't think it's worth investing the time to make this more robust.

## Changes
* Delete unstable test


johnduffell:
sure, good idea, thanks 👍

PR #1873, opened 2019-02-27
Move Babel and Grunt to devDependencies

jacobwinch:
## Why are you doing this?

As far as I know these should be listed as devDependencies. Currently Snyk frequently reports vulnerabilities against these dependencies, even though they are only used within our secure build/dev enviornments (i.e. not at runtime). This should stop if we categorise them correctly.

https://gruntjs.com/getting-started#preparing-a-new-grunt-project
https://babeljs.io/en/setup#installation


johnduffell:
thanks Jacob! 👍 🥇

PR #1878, opened 2019-03-25
Use t3.small instances

jacobwinch:
## Why are you doing this?
We are moving to [t3 instances](https://aws.amazon.com/ec2/instance-types/t3/).

## Changes
* Add t3.small as an allowed instance type (and then update Cloudformation via the console in order to modify this app's launch configuration to actually use t3.smalls)

### DIFF
--- a/cloud-formation/membership-app.cf.yaml
+++ b/cloud-formation/membership-app.cf.yaml
@@ -32,10 +32,8 @@ Parameters:
     Description: EC2 instance type
     Type: String
     AllowedValues:
-    - t2.micro
     - t2.small
-    - t2.medium
-    - m3.medium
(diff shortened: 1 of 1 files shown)


cloud-formation/membership-app.cf.yaml
```diff
@@ -32,10 +32,8 @@ Parameters:
     Description: EC2 instance type
     Type: String
     AllowedValues:
-    - t2.micro
     - t2.small
```
jacobwinch: Leaving t2.small as an option for now, because it allows us to rollback quickly (without a PR).

jacobwinch:
I've tested the Cloudformation update steps with CODE.

PR #1880, opened 2019-04-09
Relax Masterclass display rules

jacobwinch:
## Why are you doing this?

Some Masterclasses are currently being hidden due to old Membership-related business rules. The Events team have confirmed that these old business rules are no longer valid.

## Changes
* Use standard def for events (remove override which adds the problematic filter)
* Remove test code related to unused filter

## Screenshots

Before:

image

After:

image

Masterclass `Telling stories with data: Introduction to hand-drawn data visualisation` is now present.


frontend/app/services/EventbriteService.scala
```diff
@@ -189,9 +183,6 @@ class MasterclassEventService(executionContext: ExecutionContext, actorSystem: A
 
   override val httpClient: FutureHttpClient = RequestRunners.futureRunner
 
-  override def events: Seq[RichEvent] =
```
jacobwinch: See: https://github.com/guardian/membership-frontend/blob/master/frontend/app/services/EventbriteService.scala#L72

PR #1883, opened 2019-04-18
use latest libthrift (binary substitutuion only)

johnduffell:
## Why are you doing this?

Snyk has poicked up an issue with libthrift https://app.snyk.io/vuln/SNYK-JAVA-ORGAPACHETHRIFT-173706
We are using it via capi client and the ophan event model, but upgrading either of those causes compile errors.  Since I'm not going to have time to update those before next week, I have just done the version patch.

This could cause some binary incompatibilities, but I managed to check out locally with supporter membership with the patch.  So i think it should be ok, but it should be kept an eye on when it goes out.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -40,13 +40,14 @@ object Dependencies {
     ExclusionRule(organization = "org.scalactic")
   )
   val googleAuth = "com.gu" %% "play-googleauth" % "0.7.6"
+  val libthrift = "org.apache.thrift" % "libthrift" % "0.12.0"
   //projects
 
   val frontendDependencies =  Seq(googleAuth, scalaUri, membershipCommon, enumPlay,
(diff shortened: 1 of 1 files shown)


johnduffell:
are you sure, the snyk link above says:
> Affecting org.apache.thrift:libthrift artifact, versions [0.9.2, 0.11.0)

johnduffell:
@regiskuckaertz snyk is specifically complaining about 0.10 still, so do you think I should go ahead with this PR or we need to update all the routes described in https://app.snyk.io/org/the-guardian-cuu/project/17832d7d-f280-403b-9d17-4065dc14d89a/ ?

regiskuckaertz:
I know, though the vulnerability is not severe. I guess we could try using v0.12 in this project, just for a try and if that works, then I can release an update of our libraries.

johnduffell:
thanks Regis, watch this space...!

johnduffell:
looks good on v0.12

regiskuckaertz:
Ooh thank you 🙏 I'll release an update to the libs

PR #1884, opened 2019-04-18
Bump capi to v14.2

regiskuckaertz:
## Why are you doing this?

We need to upgrade our thrift dependencies to the latest version used in Scrooge (don't ask), because it currently impairs our ability to freely evolve the CAPI model. Last time we tried, it broke frontend badly. So we are in the process of upgrading across our whole codebase. I have noticed that upgrading CAPI consumers posed no problem as long as we did not change the model. So that's what I'm doing with this PR.

### DIFF
--- a/frontend/app/controllers/MemberOnlyContent.scala
+++ b/frontend/app/controllers/MemberOnlyContent.scala
@@ -1,7 +1,7 @@
 package controllers
 
 import actions.CommonActions
-import com.gu.contentapi.client.GuardianContentApiError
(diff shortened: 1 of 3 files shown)


johnduffell:
I tried exactly that earlier but it wouldn't compile.  In the end I gave up and rolled out a hacky-patch.  But if you can get it to compile, all the better.

regiskuckaertz:
@rupertbates thanks Rupert. I don't have the rights to merge, so please feel free whenever time permits 🙏

PR #1892, opened 2019-07-18
updated package.json to fix synk

damien-shaw-guardian:
## Why are you doing this?
There are two high severity packages that have shown up in synk, I have updated package.json to fix these problems, using `npm audit fix` which automatically bumps versions which have security issues

## Changes
* update to package.json


johnduffell:
I have run through this test manually and I do get to the thank you page, so I'll kicked off the tests again

johnduffell:
I think webpack needs updating too, the updates you have made are only in the devDependencies, which are not used at run time.
I can see that the two issues are in webpack dependencies which is presumably used at runtime?  If not it can be moved to devDependencies (this was incorrect in the past in places)
https://app.snyk.io/org/the-guardian-cuu/project/eec84773-4792-421c-a1f0-a1e5521599d9/

PR #1896, opened 2019-08-02
Use identity-model instead of identity models in identity-play-auth

Mullefa:
We go again! 🙄

In preparation for deprecating identity-play-auth in preference of [identity-auth-play](https://github.com/guardian/identity/tree/master/identity-auth-play/src) (which authenticates users by making a call to identity API), introduce the identity-model library, since currently we are using duplicate user models rolled out in identity-play-auth.

PR #1897, opened 2019-08-06
Synchronous authentication via identity API

Mullefa:
Follow up to #1896; authenticate a user by calling out to identity API so that session invalidation is respected.

The request to identity API is made synchronously. The reasoning behind this is the same as subscriptions frontend ([#1274](https://github.com/guardian/subscriptions-frontend/pull/1274)).

### DIFF
--- a/frontend/app/model/Identity.scala
+++ b/frontend/app/model/Identity.scala
@@ -2,6 +2,8 @@ package model
 
 // Code in this file ported over from identity-play-auth which is deprecated.
 
+import java.time.{Clock, Duration}
+
 import com.gu.identity.model.cookies.{CookieDescription, CookieDescriptionList}
 import play.api.mvc.Cookie
 // import Writes type classes required to generate Writes instances for case classes defined in this file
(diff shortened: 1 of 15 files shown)


Mullefa:
Tested on CODE. If anyone that isn't @johnduffell fancies reviewing one of these auth PRs that'd be appreciated; otherwise we'll have to wait for him on Monday 😅

frontend/app/model/Identity.scala
```diff
@@ -56,3 +58,36 @@ object AccessCredentials {
 }
 case class IdMinimalUser(id: String, displayName: Option[String])
 case class AuthenticatedIdUser(credentials: AccessCredentials, minimalUser: IdMinimalUser)
+
+// identity-play-auth (which previously provided CookieBuilder) has been deprecated in favour of identity-auth-play:
```
johnduffell: 🙈 couldn't it have been called identity-play-auth2 then at least it's clear which is the new one?  with an arbitrary change it doesn't really help in the short term.

johnduffell:
thanks for updating this, looks good, go for it!
👍

frontend/app/model/Identity.scala
Mullefa: yep, fair point :/

Mullefa:
will do John

PR #1900, opened 2019-10-11
push some snyk updates to jackson

johnduffell:
## Why are you doing this?
There is another jackson databind issue!  We have comitted to fix them within 2 weeks of a fix being available.

Tested locally, checked out a supporter membership Ok.

@alexflav23 @twrichards

### DIFF
--- a/frontend/build.sbt
+++ b/frontend/build.sbt
@@ -24,7 +24,7 @@ organization := "com.gu"
 
 version := "1.0-SNAPSHOT"
 
-scalaVersion := "2.12.8"
+scalaVersion := "2.12.10"
 
 resolvers ++= Seq(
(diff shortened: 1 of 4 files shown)


frontend/build.sbt
```diff
@@ -24,7 +24,7 @@ organization := "com.gu"
 
 version := "1.0-SNAPSHOT"
 
-scalaVersion := "2.12.8"
+scalaVersion := "2.12.10"
```
johnduffell: just for general updatedness

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.2.8
+sbt.version=1.3.2
```
johnduffell: just for general updatedness

PR #1901, opened 2019-10-11
update sbt-dependency-graph to allow snyk monitor to work

johnduffell:
After updating sbt to 1.3.2, snyk monitor didn't work any more.  This causes snyk monitor not to realise the dependencies have changed.
I have bumped it to the latest, but actually there are still outstanding bugs.  At least this means this step will pass. See the issue on the plugin: https://github.com/jrudolph/sbt-dependency-graph/issues/178
@alexflav23 @twrichards


johnduffell:
looks like the post deploy tests have been failing for a couple of months, as they need java 8 and it's not installing via the command it uses any more.

PR #1915, opened 2020-02-28
move away from idapi api display name

QuarpT:
## Why are you doing this?

We are planning on removing displayName from the identity api

## Changes

displayName is now generated using the user's username, falling back to the first name or 'firstname secondname' where it seems appropriate

### DIFF
--- a/frontend/app/controllers/Joiner.scala
+++ b/frontend/app/controllers/Joiner.scala
@@ -32,7 +32,7 @@ import services.{GuardianContentService, _}
 import tracking.AcquisitionTracking
 import utils.RequestCountry._
 import utils.TestUsers.{NameEnteredInForm, PreSigninTestCookie, isTestUser}
-import utils.{Feature, ReferralData, TestUsers, TierChangeCookies}
+import utils.{Feature, ReferralData, TestUsers, TierChangeCookies, DisplayName}
 import views.support
 import views.support.MembershipCompat._
 import views.support.Pricing._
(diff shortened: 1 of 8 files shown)


frontend/app/utils/DisplayName.scala
```diff
@@ -0,0 +1,14 @@
+package utils
+
+import com.gu.identity.model.User
+
+object DisplayName {
```
QuarpT: I'll add tests for this

frontend/app/controllers/Joiner.scala
```diff
@@ -120,7 +120,7 @@ class Joiner(
       case Some(user) => for {
         fullUser <- identityService.getFullUserDetails(user.minimalUser)(IdentityRequest(request))
         primaryEmailAddress = fullUser.primaryEmailAddress
-        displayName = fullUser.publicFields.displayName
+        displayName = DisplayName(fullUser)
```
jfsoul: Why do we have `DisplayName.apply` and `DisplayName.fallbackFullname`?

QuarpT: Speaking to @dominickendrick and Rosie the plan was to display the Identity username to users, falling back to first name in place of display name (where this is preferable on rendered web pages).

However, for `updateSalesforceContactWithMembership` it seems strange just passing a first name to Salesforce. Sending a first and last name seem more appropriate, maybe we should go with this instead and never pass a username to Salesforce 🤔 

The closest logic to the old functionality would be username, falling back to "$firstName $secondName"

rupertbates: This is all effectively dead code since we switched the acquisition flows off so I wouldn't spend too long worrying about the perfect solution

QuarpT: @rupertbates  I'll leave as is then. Is there a plan for this code to be removed or the project to be decomissioned?

PR #1920, opened 2020-04-02
Update text and add link so people can express interest

lucymonie:
## Why are you doing this?
This is a small update to the COVID-19 text, allowing people to access a form where they can express interest in future events.

## Screenshots
localhost_9100_events(desktop)

### DIFF
--- a/frontend/app/views/fragments/event/headerBarCovid.scala.html
+++ b/frontend/app/views/fragments/event/headerBarCovid.scala.html
@@ -4,7 +4,8 @@
     <h2 class="header-bar__title">
         Given the growing concerns over public health and safety due to the developing Coronavirus (COVID-19)
         situation, we have suspended our events programme until 31 May 2020. For updates please follow us on
-        <a href="@twitterLink">Twitter</a>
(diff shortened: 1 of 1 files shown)


frontend/app/views/fragments/event/headerBarCovid.scala.html
```diff
@@ -4,7 +4,8 @@
     <h2 class="header-bar__title">
         Given the growing concerns over public health and safety due to the developing Coronavirus (COVID-19)
         situation, we have suspended our events programme until 31 May 2020. For updates please follow us on
-        <a href="@twitterLink">Twitter</a>
-        or <a href="https://www.theguardian.com/guardian-masterclasses/2015/jan/19/sign-up-to-the-guardian-masterclasses-newsletter">sign up to our newsletter</a>.
+        <a href="@twitterLink">Twitter</a>,
+        <a href="https://www.theguardian.com/guardian-masterclasses/2015/jan/19/sign-up-to-the-guardian-masterclasses-newsletter">sign up to our newsletter</a>,
+        or express your interest in future events <a href="https://guardiannewsampampmedia.formstack.com/forms/masterclasses__express_your_interest_in_this_course">here</a>.
```
jfsoul, line 9: Apologies if the card wasn't too clear on this, but I think they want this only on the masterclasses page - so we might have to make this part conditional or pass it into this component?

PR #1922, opened 2020-04-03
make nginx hit membership on http as well as https

johnduffell:
## Why are you doing this?
I was getting promo instead of membership when running with dev nginx locally.
This is because nginx defaults to the first server block if it can't match host AND proto.  And we only had https defined and I was hitting http.

https://github.com/guardian/dev-nginx/pull/17 has more info


johnduffell:
most useful command

`git commit --amend -m '[.....]'`

😂

PR #1923, opened 2020-04-06
update packages to snyk compliant version

johnduffell:
## Why are you doing this?
Snyk usual updates and version bumps to avoid security issues.

I have also done some other cleanup at the same time.

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -1,7 +1,7 @@
 // Comment to get more information during initialization
 logLevel := Level.Warn
 
-resolvers += "Typesafe repository" at "http://repo.typesafe.com/typesafe/releases/"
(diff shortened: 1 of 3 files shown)


project/plugins.sbt
```diff
@@ -1,7 +1,7 @@
 // Comment to get more information during initialization
 logLevel := Level.Warn
 
-resolvers += "Typesafe repository" at "http://repo.typesafe.com/typesafe/releases/"
+resolvers += "Typesafe repository" at "https://repo.typesafe.com/typesafe/releases/"
```
johnduffell, line 4: this was causing a warning because http repos could be hijacked and bad stuff injected.  If we go straight to https then that takes away the opportunity to be redirected somewhere bad.

PR #1924, opened 2020-04-07
use non-deprecated url for events feed

johnduffell:
## Why are you doing this?
Eventbrite are deprecating some aspects of their API.  We need to move over to the new url.

> GET /v3/users/me/owned_events/ - this is in the process of deprecation and we strongly encourage shifting to GET /v3/organizations/:organization_id/events/

## Trello card: [Here](https://trello.com/c/5vnK0iXf/2731-eventbrite-api-upgrade-for-membershiptheguardiancom)

Tested locally and via postman to the old and new endpoints
The events in http://mem.thegulocal.com/events/archive are being populated okay.

### DIFF
--- a/frontend/app/services/EventbriteService.scala
+++ b/frontend/app/services/EventbriteService.scala
@@ -43,7 +43,7 @@ abstract class EventbriteService(implicit val ec: ExecutionContext, system: Acto
   def eventsTaskFor(status: String, initialDelay: FiniteDuration, refreshTime: FiniteDuration): ScheduledTask[Seq[RichEvent]] =
     ScheduledTask[Seq[RichEvent]](s"Eventbrite $status events", Nil, initialDelay, refreshTime) {
       for {
-        events <- getAll[EBEvent]("users/me/owned_events/", List(
+        events <- getAll[EBEvent]("organizations/104062388105/events/", List(
           "status" -> status,
           "expand" -> EBEvent.expansions.mkString(",")))
         richEvents <- Future.traverse(events)(mkRichEvent)
--- a/frontend/app/services/EventbriteService.scala
+++ b/frontend/app/services/EventbriteService.scala
@@ -18,9 +18,25 @@ import utils.StringUtils._
 import scala.concurrent.{ExecutionContext, Future}
 import scala.concurrent.duration._
 
-abstract class EventbriteService(implicit val ec: ExecutionContext, system: ActorSystem) extends WebServiceHelper[EBObject, EBError] {
+case class OrganisationId(value: String) extends AnyVal
+case class EBAccount(apiToken: String, organisationId: OrganisationId)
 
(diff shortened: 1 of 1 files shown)


frontend/app/services/EventbriteService.scala
```diff
@@ -43,7 +43,7 @@ abstract class EventbriteService(implicit val ec: ExecutionContext, system: Acto
   def eventsTaskFor(status: String, initialDelay: FiniteDuration, refreshTime: FiniteDuration): ScheduledTask[Seq[RichEvent]] =
     ScheduledTask[Seq[RichEvent]](s"Eventbrite $status events", Nil, initialDelay, refreshTime) {
       for {
-        events <- getAll[EBEvent]("users/me/owned_events/", List(
+        events <- getAll[EBEvent]("organizations/104062388105/events/", List(
```
jfsoul, line 46: I think we have more than one organisation - there's one for masterclasses and one for live events. I.e. the difference between:
https://membership.theguardian.com/events AND
https://membership.theguardian.com/masterclasses

jfsoul: I believe I've got access to access these if you don't have one or the other of them.

johnduffell: oh yes I can see that, I just looked at the main api key, good spot 🏅

frontend/app/services/EventbriteService.scala
```diff
@@ -18,9 +18,25 @@ import utils.StringUtils._
...
+    OrganisationId("7266571413")
+  )
+}
+abstract class EventbriteService(ebAccount: EBAccount)(
+  implicit val ec: ExecutionContext,
+  system: ActorSystem
+) extends WebServiceHelper[EBObject, EBError] {
 
-  val apiToken: String
+  import ebAccount._
```
johnduffell, line 39: I decided to make the EBAccount self contained and group the id with the key, rather than making the service classes bigger. it did add a bit of boilerplate though in making it more explicit.

frontend/app/services/EventbriteService.scala
```diff
@@ -122,25 +138,22 @@ abstract class EventbriteService(implicit val ec: ExecutionContext, system: Acto
   def getOrder(id: String): Future[EBOrder] = get[EBOrder](s"orders/$id/", "expand" -> EBOrder.expansions.mkString(","))
 }
 
-abstract class LiveService(implicit ec: ExecutionContext, system: ActorSystem, contentApiService: GuardianContentService, gridService: GridService) extends EventbriteService {
```
johnduffell, line 125: this liveservice class looked like it was there for existing/old reasons so I got rid of it.  Although given @rupertbates comments last week about assuming things are needed, I'm not sure it's a good idea!  But this time I'm more confident it's not needed, last time I was quite uncertain.

PR #1929, opened 2020-04-17
get event description from new endpoint

johnduffell:
## Why are you doing this?

Folliwing https://github.com/guardian/membership-frontend/pull/1924
The eventbrite events created via the [structured event flow](https://www.eventbrite.com/platform/docs/event-description) (the new one) will return the summary only in the description field, and the full/structured description will only be available by making [a separate call](https://www.eventbrite.com/platform/api#/reference/event-description).

This PR adds the second call, combines the two objects into a new EventWithDescription model and updates the code to use that where it used the old description.

### Before
image
### after
image


frontend/app/controllers/Event.scala
```diff
@@ -49,7 +49,7 @@ class Event(
 
     override def invokeBlock[A](request: Request[A], block: (Request[A]) => Future[Result]): Future[Result] = {
       eventbriteService.getEvent(eventId).map { event =>
-        SafeLogger.info(s"Buy action invoked for event: ${event.id}")
+        SafeLogger.info(s"Buy action invoked for event: ${event.underlying.ebEvent.id}")
```
johnduffell, line 52: there was originally an implicit converion from RichEvent to Event, which I've removed.
https://github.com/guardian/membership-frontend/pull/1929/files#diff-3241cd55902f5775708482880f94420bL186
We have the extra layer because we have to group together the Description and Event objects.  This extra layer is because we expose the underlying wire-data model which has changed.
I thought about copying the new description into the old event description, but that seemed bodgetastic and I'd like to keep the general architecture that already exists where possible.
Another option would be to flesh out a couple more implicit conversions, but they are generally to be avoided e.g. https://dotty.epfl.ch/docs/reference/contextual/motivation.html

frontend/app/controllers/Event.scala
```diff
@@ -160,11 +160,11 @@ class Event(
   }
 
   def eventbriteRedirect(event: RichEvent, discountCodeOpt: Option[EBCode])(implicit req: RequestHeader) = {
-    val eventUrl = discountCodeOpt.fold(Uri.parse(event.url))(c => event.url ? ("discount" -> c.code))
+    val eventUrl = discountCodeOpt.fold(Uri.parse(event.underlying.ebEvent.url))(c => event.underlying.ebEvent.url ? ("discount" -> c.code))
     Found(addEventBriteGACrossDomainParam(eventUrl)).withCookies(Cookie(eventCookie(event), "", Some(3600)))
   }
 
-  @deprecated("Guardian Live Events are not configured to use this anymore, and host their own thankyou page on EventBrite.")
+  @deprecated("Guardian Live Events are not configured to use this anymore, and host their own thankyou page on EventBrite.", "")
```
johnduffell, line 167: the single argument deprecated annotation is deprecated - the second argument is to say which version it's deprecated since

frontend/app/model/Eventbrite.scala
```diff
@@ -220,25 +220,10 @@ object Eventbrite {
...
-    created: Instant,
-    venue: EBVenue,
-    capacity: Int,
-    ticket_classes: Seq[EBTicketClass],
-    status: String
-  ) extends EBObject {
-
-    val times = EventTimes(created, start)
-    val startAndEnd = new Interval(start, end)
+  case class EventWithDescription(ebEvent: EBEvent, ebDescription: EBDescription) {
```
johnduffell, line 223: EventWithDescription wraps the two new classes, and the methods are distributed as appropriate between the three classes that result, but not many changes

frontend/app/model/Eventbrite.scala
```diff
@@ -302,6 +272,54 @@ object Eventbrite {
...
+  ) extends EBObject {
+
+    val providerOpt = for {
+      m <- "<!-- provider: (\\S+) -->".r.findFirstMatchIn(description)
+      providerOpt <- EBEvent.availableProviders.find(_.id == m.group(1))
+    } yield providerOpt
+
+    val isPartnerEvent = providerOpt.isDefined
+
+    def cleanHtml: String = {
```
johnduffell, line 288: this was copied from cleanHtml elsewhere

frontend/app/model/RichEvent.scala
```diff
@@ -64,7 +64,7 @@ object RichEvent {
   }
 
   trait RichEvent {
-    val event: EBEvent
+    val underlying: EventWithDescription
```
johnduffell, line 67: we ended up with `event.event.event.id` in many places, so I changed this to `underlying` and the model event to `ebEvent`.  Any alternative suggestions would be taken though.

frontend/test/model/EBEventTest.scala
```diff
@@ -6,25 +6,45 @@ import model.EventbriteDeserializer._
...
+  def completedEvent = ebResponse.data.find(_.id == "13024577863").get.cheatyMigrate
+  def limitedAvailabilityEvent = ebResponse.data.find(_.id == "12718560557").get.cheatyMigrate
+
+  def ticketedEvent = ebLiveEvent
 
-  val ticketedEvent = ebLiveEvent;
+  def updateEventWithDescription(event: EventWithDescription, desc: String) = event.copy(ebDescription = event.ebDescription.copy(description = desc))
 
-  def updateEventWithDescription(event: EBEvent, desc: String) = event.copy(description = Some(EBRichText(desc,desc)))
+  "structured event" should {
```
johnduffell, line 34: this is the only test for the new event type, the old test sare all still there and designed to use the old data and fill in the new description field with what I think eventbrite does when it's migrating the old events
We need to add some new tests when we decide how to model the image url etc using the new description field.

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.3.2
+sbt.version=1.3.10
```
johnduffell, line 1: thought I might as well slip that one in there

frontend/test/utils/Implicits.scala
```diff
@@ -0,0 +1,14 @@
+package utils
+
+import model.Eventbrite.{EBDescription, EBEvent, EventWithDescription}
+
+object Implicits {
+
+  implicit class Cheat(event: EBEvent) {
+    // this is just a way to get legacy events from the test data into the new format
+    // the proper way would be to recreate in eventbrite, and run the actual api calls to get it
+    def cheatyMigrate: EventWithDescription =
```
jfsoul, line 10: I was a bit confused as to what this was doing because of the name - is there anything more descriptive we could use? `toEventWithSeparateDescription` (just off the top of my head)

johnduffell: I wanted to make it obvious this was migrating the old events to the new format in the way I assumed would happen, rather than actually getting them from the API in the correct format (as the test data was produced years ago and setting it up again would be a big manual process)  However you're right I should have gone back and thought about the name again to be more clear, I'll go in and tweak it.

PR #1938, opened 2020-06-09
Disable feature warnings

mario-galic:
Related PR: https://github.com/guardian/membership-common/pull/616

This is related to [`ZuoraSoapService#getFeatures`](https://github.com/guardian/membership-common/blob/436462503fed218800bdb0989c37b5aef24abf15/src/main/scala/com/gu/zuora/ZuoraSoapService.scala#L236) which seems to be [dead](https://github.com/search?q=org%3Aguardian+getFeatures&type=Code) and removing it would result in simplifications of membership-common, in particular weakening dependency on 

```
"com.typesafe.akka" %% "akka-agent" % "2.5.30"
```

which is deprecated.

https://github.com/guardian/membership-common/blob/436462503fed218800bdb0989c37b5aef24abf15/src/main/scala/com/gu/zuora/soap/ClientWithFeatureSupplier.scala#L28

```
class ClientWithFeatureSupplier(
    featureCodes: Set[Code],
    apiConfig: ZuoraSoapConfig,
    httpClient: FutureHttpClient,
    extendedHttpClient: FutureHttpClient,
    metrics: ZuoraMetrics = NoOpZuoraMetrics
)(implicit actorSystem: ActorSystem, ec: ExecutionContext)
  extends Client(apiConfig, httpClient, extendedHttpClient, metrics) {

  val featuresSupplier = new FutureSupplier[Seq[Feature]](getFeatures)
  actorSystem.scheduler.schedule(30.minutes, 30.minutes)(featuresSupplier.refresh())

  private def getFeatures: Future[Seq[Feature]] = {
    val featuresF = query[Feature](SimpleFilter("Status", "Active"))
    featuresF.foreach { features =>
      val diff = featureCodes &~ features.map(_.code).toSet
      if (diff.nonEmpty) {
        SafeLogger.error(scrub"Zuora ${apiConfig.envName} is missing the following product features: ${diff.mkString(", ")}. Please update configuration ASAP!")
      }
    }
    featuresF
  }
}
```

### DIFF
--- a/frontend/app/services/TouchpointBackend.scala
+++ b/frontend/app/services/TouchpointBackend.scala
@@ -63,7 +63,7 @@ object TouchpointBackend {
     // extendedRunner sets the configurable read timeout, which is used for the createSubscription call.
     val extendedRunner = RequestRunners.configurableFutureRunner(20.seconds)
 
-    val zuoraSoapClient = new ClientWithFeatureSupplier(FeatureChoice.codes, config.zuoraSoap, runner, extendedRunner)
+    val zuoraSoapClient = new ClientWithFeatureSupplier(Set.empty, config.zuoraSoap, runner, extendedRunner)
     val discounter = new Discounter(Config.discountRatePlanIds(config.zuoraEnvName))
     val zuoraSoapService = new ZuoraSoapServiceImpl(zuoraSoapClient)
     val zuoraRestService = new ZuoraRestService[Future]


frontend/app/services/TouchpointBackend.scala
```diff
@@ -63,7 +63,7 @@ object TouchpointBackend {
     // extendedRunner sets the configurable read timeout, which is used for the createSubscription call.
     val extendedRunner = RequestRunners.configurableFutureRunner(20.seconds)
 
-    val zuoraSoapClient = new ClientWithFeatureSupplier(FeatureChoice.codes, config.zuoraSoap, runner, extendedRunner)
+    val zuoraSoapClient = new ClientWithFeatureSupplier(Set.empty, config.zuoraSoap, runner, extendedRunner)
```
paulbrown1982, line 66: Legacy Patron Members could select either Books or Events benefits on checkout. These were stored as Features in the Zuora record. If they select Events, rather than Books, then the site will give them access to the discounted events.

paulbrown1982: As long as that logic still works after this then the change is safe.

mario-galic: This logic is implemented via https://github.com/guardian/membership-frontend/blob/master/frontend/app/model/FeatureChoice.scala so it is not using membership-common for it. When patrons want to renew, what is the process? Do CSRs direct them to the hidden checkout?

johnduffell: patrons don't really renew themselves, it happens automatically in zuora. At present they get unlimited free tickets due to a missing eventbrite integration as we sell tickets directly there now.  The check is based on the dates in zuora, so it automatically would reset every year anyway.
I don't think we should be acquiring new patrons, certainly not in any significant numbers.

johnduffell:
looks good thanks 👍

johnduffell:
I noticed there are other uses of `getFeatures` in this project - was this the first step and are you planning on bumping the version of membership-common to latest in this project once you are happy?

mario-galic:
@johnduffell Strange it did no show up on this search

https://github.com/search?q=org%3Aguardian+getFeatures&type=Code

I will revert.

mario-galic:
Any chance we could invest some effort into retiring this repo. It is forcing us to implement risky workarounds on production system which are actually brining value.

johnduffell:
I don't think so (certainly not from a techncial side) as events is a whole thing that is carrying on for the foreseeable future and as far as I know is also bringing value.  We are not doing anything other than essential maintenance, but we still have a large number of members and some get free tickets via their membership.
We are hoping to remove new members acquisition eventually (but not existing members) but we are focussing on subscriptions-frontend (print + digital) next quarter as we can progress that as a dev team.

mario-galic:
Perhaps decrease the surface by taking out only the parts that are used or bringing value.

PR #1941, opened 2020-06-11
actually bump scala uri, also fix EB rate limit and dates

johnduffell:
## Why are you doing this?

second attempt at scala uri bump, the old one was still being picked up from membership-common as the package names and maven artifact names are different, so it didn't evict the old one and kept using it.

first attempt here: https://github.com/guardian/membership-frontend/pull/1939

I also tweaked the dates a bit more based on further comments.
image

I have also reduced the refresh time for event brite to try to improve the rate limit errors.  It will refresh every 5 mins instead of every minute.

### DIFF
--- a/frontend/conf/PROD.public.conf
+++ b/frontend/conf/PROD.public.conf
@@ -52,3 +52,6 @@ google.adwords.joiner.conversion {
 }
 
 logstash.enabled=false
+
+eventbrite.api.refresh-time-seconds=299
+eventbrite.api.refresh-time-priority-events-seconds=119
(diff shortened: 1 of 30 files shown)


frontend/conf/PROD.public.conf
```diff
@@ -52,3 +52,6 @@ google.adwords.joiner.conversion {
 }
 
 logstash.enabled=false
+
+eventbrite.api.refresh-time-seconds=299
+eventbrite.api.refresh-time-priority-events-seconds=119
```
johnduffell, line 57: reduced these to try to prevent rate limit errors.  We can always bump further or reduce the size of the ASG to help further.

mattoh1:
👍

PR #1944, opened 2020-06-29
Remove appNexus tracking/analytics

GHaberis:
## Why are you doing this?

As part of the work to support CCPA we should make sure our third party "advertising" tags are behind a consent check if required, and if no longer required we should remove them.

The AppNexusPixel scala template is no longer required and should be removed.


GHaberis:
This is now going to remain but move behind a consent check

PR #1945, opened 2020-06-30
Show CCPA CMP to US users

GHaberis:
## Why are you doing this?

From July 1st the main Guardian domain will be collecting US users consent state in relation to CCPA. We will also be required to do this across on `membership.theguardian.com`.

@guardian/commercial-dev have been working on a library https://www.npmjs.com/package/@guardian/consent-management-platform that can be used to surface a CCPA compliant CMP to our US audience. If a user is in the US, we should initialise the CCPA CMP, otherwise we should continue using the current CMP.

The CCPA CMP will show on the user's first page view, at which point they have the opportunity to opt out of tracking, if they don't opt out at this point they will be considered to have opted in all subsequent page views and the CMP will not show again.

[**Trello Card**](https://trello.com/c/HleorMOx/3123-implement-ccpa-cmp-for-us-audience-on-membershiptheguardiancom)

## Changes

- `ccpa.js` checks the `/geocountry` endpoint, if the country code is "US" and `useCCPA` is true it resolves as `ccpaEnabled()` resolves as `true`.
- `main.js` now imports `@guardian/consent-management-platform` as `cmp`. If `ccpaEnabled()`  resolves `true` we call `cmp.init()` with `{ useCCPA: true }` when we initialise the page, this will show the CCPA CMP to that user if required. 
- We also now  pass a `ccpaEnabled` boolean to `consentBanner.init()`, if `ccpaEnabled` is `true` we hide the non-CCPA consent banner.
- `getTrackingConsent` in `thirdPartyTrackingConsentTest.js` is now asynchronous and returns a `Promise<ThirdPartyTrackingConsent> ` instead of `ThirdPartyTrackingConsent`. This is necessary because we have to use `onIabConsentNotification` to obtain CCPA `consentState` which is an asynchronous action.
- Hide the grey CMP with `display: none` by default and update to `display: block` if it needs to be shown.

## Todo

- Put CCPA behind a feature switch so it can easily be enabled/disabled.

## Screenshots

**CCPA CMP on Mobile...**

mem-us-cmp-mobile

**CCPA CMP on Desktop...**

mem-us-cmp-dt

### DIFF
--- a/frontend/assets/javascripts/src/main.js
+++ b/frontend/assets/javascripts/src/main.js
@@ -28,7 +28,9 @@ require([
     'src/modules/faq',
     'src/modules/landingBundles',
     'src/modules/bundlesLanding',
-    'src/modules/consentBanner'
+    'src/modules/consentBanner',
+    '../utils/cookie',
+    '@guardian/consent-management-platform'
 ], function(
     ajax,
     raven,
@@ -59,13 +61,29 @@ require([
     faq,
     landingBundles,
     bundlesLanding,
-    consentBanner
+    consentBanner,
+    cookie,
+    cmp
 ) {
     'use strict';
 
+    const countryId = cookie.getCookie('GU_country');
+
+    /**
+     * If in US initialise CMP as early as possible so
+     * subsequent call to onIabConsentNotification
+     * returns the correct consentState.
+    * */
+    if (countryId === 'US') {
+        cmp.init({
+            useCcpa: true,
+        });
+    } else {
+        consentBanner.init();
+    }
+
     ajax.init({page: {ajaxUrl: ''}});
     raven.init('https://d35a3ab8382a49889557d312e75b2179@sentry.io/1218929');
-
     analytics.init();
 
     // Global
@@ -109,6 +127,4 @@ require([
     memstatus.init();
 
     faq.init();
-
-    consentBanner.init();
 });
--- a/frontend/assets/javascripts/src/main.js
+++ b/frontend/assets/javascripts/src/main.js
@@ -67,22 +67,22 @@ require([
 ) {
     'use strict';
 
-    const ccpaEnabled = ccpa.ccpaEnabled();
+    ccpa.ccpaEnabled().then(useCcpa => {
+        /**
+         * If ccpaEnabled initialise CCPA CMP as early
+         * as possible so subsequent call so
+         * onIabConsentNotification returns the correct
+         * consentState.
+        * */
+        if (useCcpa) {
+            cmp.init({
+                useCcpa
+            });
+        }
 
-    /**
-     * If ccpaEnabled initialise CCPA CMP as early
-     * as possible so subsequent call so
-     * onIabConsentNotification returns the correct
-     * consentState.
-    * */
-    if (ccpaEnabled) {
-        cmp.init({
-            useCcpa: true,
-        });
-    }
-
-    // If ccpaEnabled is true consentBanner will hide non-CCPA banner
-    consentBanner.init(ccpaEnabled);
+        // If useCCPA is true consentBanner will hide non-CCPA banner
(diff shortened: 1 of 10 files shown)


frontend/assets/javascripts/src/main.js
```diff
@@ -59,13 +61,31 @@ require([
...
+    ccpa,
+    cmp
 ) {
     'use strict';
 
+    const ccpaEnabled = ccpa.ccpaEnabled();
+
+    /**
+     * If ccpaEnabled initialise CCPA CMP as early
+     * as possible so subsequent call so
```
ripecosta, line 74:
```suggestion
     * as possible so subsequent call to
```

ripecosta, line 74:
```suggestion
     * as possible so subsequent call to
```

frontend/assets/javascripts/src/modules/ccpa.js
```diff
@@ -0,0 +1,37 @@
...
+    if (!useCCPA) {
+        return Promise.resolve(false);
+    }
+
+    if (fetchCountry) {
+        return fetchCountry;
+    }
+
+    fetchCountry = fetch(GEOCOUNTRY_URL).then(response => {
+        if (response.ok) {
```
johnduffell, line 24: we probably need to make the URL return an InternalServerError rather than Ok if the header was not set.  Then we will get an alarm if this breaks.
We probably don't want to stop showing it altogether if the header disappears.

frontend/assets/javascripts/src/modules/ccpa.js
```diff
@@ -0,0 +1,37 @@
...
+            return response.text();
+        } else {
+            throw new Error('failed to get geocountry');
+        }
+    }).then(responseCountryCode => {
+        return responseCountryCode === US_COUNTRY_CODE;
+    }).catch(err => {
+        Raven.captureException(err);
+        return false;
```
johnduffell, line 33: will we notice if it is following this code path?  not sure if we want to return false when something goes wrong if we are not going to notice anything

GHaberis: I've included the `Raven.captureException(err);` call here so we log these errors in `Sentry`, so we'll have visibility there.

frontend/assets/javascripts/src/modules/analytics/thirdPartyTracking.js
```diff
@@ -9,13 +12,46 @@ const OptedOut = 'OptedOut';
...
 
 const getTrackingConsent = () => {
-    const cookieVal = getCookie(ConsentCookieName);
-    if (cookieVal && cookieVal.split('.')[0] === '1') { return OptedIn; }
-    if (cookieVal && cookieVal.split('.')[0] === '0') { return OptedOut; }
-    return Unset;
-};
+    return ccpaEnabled().then(useCcpa => {
+        if (useCcpa) {
+            return new Promise((resolve) => {
```
paulbrown1982, line 17: Did you add Promise to the polyfil.js for IE11?

paulbrown1982: Ah I found it.

frontend/assets/javascripts/src/modules/ccpa.js
```diff
@@ -0,0 +1,37 @@
...
+const GEOCOUNTRY_URL = '/geocountry';
+const US_COUNTRY_CODE = 'US';
+
+let fetchCountry;
+
+export const ccpaEnabled = () => {
+    const useCCPA = true; // set false to switch CCPA off
+
+    if (!useCCPA) {
+        return Promise.resolve(false);
```
mxdvl, line 16: You may want to use the same camelCase as elsewhere `useCcpa` ?

PR #1946, opened 2020-07-02
Set up geocountry route on membership-frontend

GHaberis:
## Why are you doing this?

This PR introduces a new `/geocountry` route to membership frontend, this is required so we can retrieve the users' country code, which is necessary for knowing which CMP to present them with.

### DIFF
--- a/frontend/build-tc.sh
+++ b/frontend/build-tc.sh
@@ -1,6 +1,12 @@
 #!/bin/bash
 echo ${BUILD_NUMBER:-DEV} > conf/build.txt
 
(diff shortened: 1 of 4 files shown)


frontend/build-tc.sh
```diff
@@ -1,6 +1,12 @@
 #!/bin/bash
 echo ${BUILD_NUMBER:-DEV} > conf/build.txt
 
+export NVM_DIR="$HOME/.nvm"
```
johnduffell, line 4: This moves the node install from teamcity to the script

PR #1956, opened 2020-08-07
remove join pages from membership

johnduffell:
## Why are you doing this?

Membership acquisition is not needed any more.  We prefer people to contribute or subscribe.


rupertbates:
I think we should probably redirect the `choose-tier` url and the various `tier/change` urls

frontend/conf/routes
```diff
@@ -20,6 +20,8 @@ GET         /feature/:feature                          controllers.FeatureOptIn.
 GET         /feature/:feature/:onOrOff                 controllers.FeatureOptIn.setFeatureOnOrOff(feature: Feature, onOrOff: OnOrOff)
 
 # Join
+GET         /join/*any                                controllers.Redirects.whySupportRedirectIgnore(any)
```
rupertbates, line 23: We could also disallow `/join` in robots.txt

johnduffell: Hadn't thought about that but what benefit are you expecting from robots.txt?
I assume search engines will follow the redirect, ideally we want people who google join membership to get to the support page.  I did a small amount of research just now and they seem to just say use a redirect.
 https://support.google.com/webmasters/answer/93633

rupertbates: That's a fair point, I always find that it's quite hard to get Google to remove paths from the index but as we're redirecting that should be fine. 👍

frontend/app/model/Faq.scala
```diff
@@ -38,41 +38,6 @@ object Faq {
   )
 
   val help = List(
-    Section("Guardian Membership: the basics", List(
```
johnduffell, line 41: image

PR #1973, opened 2020-09-11
Fixing GA to load when all consents have been agreed to

paulbrown1982:
## Why are you doing this?

Fixing GA to load when all consents have been agreed to. This is whilst we wait for GA to be properly set up as a vendor.

### DIFF
--- a/frontend/assets/javascripts/src/modules/analytics/ga.es6
+++ b/frontend/assets/javascripts/src/modules/analytics/ga.es6
@@ -90,6 +90,9 @@ export function wrappedGa(a,b,c){
 
 export function init() {
     let guardian = window.guardian;
+
+    console.log('Loading Google Analytics');
+
     create();
 
     wrappedGa('require', 'linker');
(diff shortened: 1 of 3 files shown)


frontend/assets/javascripts/src/modules/analytics/ga.es6
```diff
@@ -90,6 +90,9 @@ export function wrappedGa(a,b,c){
 
 export function init() {
     let guardian = window.guardian;
+
+    console.log('Loading Google Analytics');
```
GHaberis, line 94: Was this for debugging? Do you want to keep this in?

paulbrown1982: Yes I want to keep this.

GHaberis (approved):
👍

frontend/assets/javascripts/src/modules/analytics/ga.es6
paulbrown1982: Actually, I'll do it the same way as subs.

paulbrown1982:
This has been tested in CODE.

PR #1975, opened 2020-10-22
Remove tier change option

rupertbates:
## Why are you doing this?

We want to remove the option to change tiers from the membership site and instead direct people to support.

Ultimately we want to remove all code related to membership and the dependency on membership-common however that is a bit more involved because it is quite tangled up with the events code however this is another step in that direction.


frontend/conf/routes
```diff
@@ -83,15 +83,12 @@ GET         /preview-event/:id                         controllers.Event.preview
...
-POST        /tier/cancel/confirm                       controllers.TierController.redirectToNewCancellationFlow
-GET         /tier/cancel/paid/summary                  controllers.TierController.redirectToNewCancellationFlow
-GET         /tier/cancel/free/summary                  controllers.TierController.redirectToNewCancellationFlow
+GET         /tier/change                               controllers.Redirects.supportRedirect
+GET         /tier/cancel                               controllers.Redirects.redirectToNewCancellationFlow
+POST        /tier/cancel/confirm                       controllers.Redirects.redirectToNewCancellationFlow
+GET         /tier/cancel/paid/summary                  controllers.Redirects.redirectToNewCancellationFlow
+GET         /tier/cancel/free/summary                  controllers.Redirects.redirectToNewCancellationFlow
 GET         /tier/change/patron                        controllers.Info.patronsRedirect
-GET         /tier/change/:tier                         controllers.TierController.upgrade(tier: PaidTier)
```
johnduffell, line 92: will this 404 now rather than redirect?

johnduffell (approved):
great, wasn't too bad after all 👍

PR #1976, opened 2020-10-29
Remove more membership related code

rupertbates:
## Why are you doing this?
This PR removes a whole load more membership related routes and code from membership-frontend. Ultimately it would be great to remove all of the membership code as well as the dependency on membership-common as this would greatly reduce the amount of maintenance work required to patch security vulnerabilities.

Follows on from #1975


frontend/app/actions/Fallbacks.scala
```diff
@@ -18,14 +18,10 @@ object Fallbacks {
   def memberHome(implicit request: RequestHeader) =
     redirectTo(controllers.routes.FrontPage.welcome)
 
-  def notYetAMemberOn(implicit request: RequestHeader) =
-    redirectTo(controllers.routes.Joiner.tierChooser()).addingToSession("preJoinReturnUrl" -> request.uri)
+  def notYetAMemberOn(implicit request: RequestHeader) = Forbidden
```
johnduffell, line 21: could use Gone or NotFound
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/410

frontend/app/controllers/Healthcheck.scala
```diff
@@ -21,8 +21,6 @@ class Healthcheck(eventbriteService: EventbriteCollectiveServices, touchpointBac
   def tests = Seq(
     //new BoolTest("Events", () => eventbriteService.guardianLiveEventService.events.nonEmpty),
     new BoolTest("CloudWatch", () => CloudWatchHealth.hasPushedMetricSuccessfully),
-    new BoolTest("ZuoraPing", () => touchpointBackends.Normal.zuoraSoapClient.lastPingTimeWithin(2.minutes)),
-    new BoolTest("Salesforce", () => touchpointBackends.Normal.salesforceService.isAuthenticated)
```
johnduffell, line 25: great, this will improve uptime a lot! 👍

frontend/app/services/MemberService.scala
```diff
@@ -164,9 +162,6 @@ class MemberService(
       SafeLogger.error(scrub"Could not create Salesforce contact for user ${user.id}", e)
```
johnduffell, line 162: there's more code here ^^^^^^ by the looks for a future cleanup

frontend/conf/routes
```diff
@@ -22,22 +22,7 @@ GET         /feature/:feature/:onOrOff                 controllers.FeatureOptIn.
 # Join
 GET         /join/*any                                controllers.Redirects.whySupportRedirectIgnore(any)
 GET         /choose-tier                               controllers.Redirects.whySupportRedirect
-#####Dead routes starting with /join - needed because of reverse routing until we delete the code
```
johnduffell, line 25: 👏

johnduffell (approved):
image
nice work! 👏  looks great and I agree that remove the membership common dependency would be great in future.

frontend/app/actions/Fallbacks.scala
rupertbates: I think the thing is that it does still exist for members so it's not gone, but if you are not a member it is forbidden

PR #1977, opened 2020-11-05
track conversion on buy button click

GHaberis:
## Why are you doing this?

This PR adds GA conversion tracking to the membership site in order to track 'Book ticket' button clicks. The remarketing module is already behind a consent check, so these clicks will only be tracked if consent has been granted.

Trello: https://trello.com/c/V1kSIA1x/3426-track-ga-conversion-tag-on-membership-site-buy-button-clicks


lucymonie (approved):
LGTM 👍

PR #1978, opened 2020-11-11
Prepare for Australian CMP w/ JS Build Fixes 

GHaberis:
## What are you doing in this PR?

**Summary**
- Replace the polyfill script, the script previously in use was not polyfilling `Promise` on older browsers (eg. IE11) as expected, meaning the Webpack generated bundle failed to run. The new script is the one we use on `support-frontend` and polyfills as expected.
- Fix the build pipeline to transpile `.js` aswell as `.es6` module, previously any modern JS in a `.js` module wasn't transpiled, which would again cause issues on older browsers (eg. IE11)
- Update babel dependencies and configuration as per babel's recommended format https://babeljs.io/docs/en/babel-plugin-transform-runtime.
- Move from `lodash-amd` to `lodash`. Once I'd fixed the polyfill and transpilation issues I found `lodash-amd` modules would cause a syntax error in IE11 whenever they were imported. Replacing with `lodash` resolved this issue.
- Update the `@guardian/consent-management-platform` dependency `^6.6.1` in preparation for adopting the new means of calling `cmp.init` (using `country` over `isInUsa`) and therefore enabling the Australian CMP. The launch date for this has been pushed back but we can continue to use the old `country` interface for now.

**These changes have been tested on `CODE` using IE11 and Chrome and the JS runs without error. The bundle size has increased slightly from 1.7kb to 1.9kb**

**Details**

This PR upgrades the `@guardian/consent-management-platform` dependency `^6.6.1` and updates the way we call `cmp.init()`.

To use the latest version of `@guardian/consent-management-platform` I needed to update the webpack config to put all `@guardian` scope npm packages through the babel pipeline, further details of why are listed in this PR https://github.com/guardian/frontend/pull/23011 which made  the equivalent change on `frontend`. This also required bringing all `babel` dependencies up to date. I've also had to switch from using `.babelrc` to `babel.config.json`, this is babel's recommended format https://babeljs.io/docs/en/babel-plugin-transform-runtime

I've also made a change so we transpile all `.js` files, previously we were only transpiling `.es6` files, this means we can use modern js in the `.js` files, currently if we they contain modern js this isn't transpiled. This required the inclusion of the `@babel/plugin-transform-modules-amd` dependency. 

**TODO: Next step would be to see if we can change the file extension of these `.es6` files to `.js`. We can also probably remove a lot of these scripts now the acquisition flows and tier change flow are deprecated.**.

[**Trello Card**](https://trello.com/c/Mzl9SDju/3443-upgrade-cmp-library-to-v661-on-membership-frontend)

## Why are you doing this?

There’s a new opt-out consent framework for Australia. It’s likely to be launched Thursday 12/11/2020. 

The new initialisation requires passing in the ISO country code as a string instead of the `isInUsa` boolean, we should therefore update to the latest version of the `@guardian/consent-management-platform` (`v6.6.1`).

### DIFF
--- a/frontend/app/views/fragments/javaScriptFirstSteps.scala.html
+++ b/frontend/app/views/fragments/javaScriptFirstSteps.scala.html
@@ -13,8 +13,7 @@
 <!--[if IE 9]>
 <script src="@Asset.at("javascripts/lib/polyfills.min.js")"></script>
 <![endif]-->
-<!-- <script src="https://polyfill.guim.co.uk/v2/polyfill.min.js?features=Promise,fetch"></script> -->
-<script src="https://polyfill.io/v3/polyfill.min.js?features=default,Array.prototype.find,fetch,Array.prototype.includes,Number.isInteger,Object.values,Object.entries"></script>
+<script src="https://polyfill.guim.co.uk/v2/polyfill.min.js?features=Promise,fetch"></script>
 <script id="gu">
     var guardian = {
         isModernBrowser: (
--- a/frontend/app/views/fragments/javaScriptFirstSteps.scala.html
+++ b/frontend/app/views/fragments/javaScriptFirstSteps.scala.html
@@ -13,7 +13,7 @@
 <!--[if IE 9]>
 <script src="@Asset.at("javascripts/lib/polyfills.min.js")"></script>
 <![endif]-->
-<script src="https://polyfill.io/v3/polyfill.min.js?features=default,Array.prototype.find,fetch,Array.prototype.includes,Number.isInteger,Object.values,Object.entries"></script>
+<script src="https://assets.guim.co.uk/polyfill.io/v3/polyfill.min.js?features=default,Array.prototype.find,fetch,Array.prototype.includes,Number.isInteger,Object.values,Object.entries"></script>
 <script id="gu">
     var guardian = {
         isModernBrowser: (
--- a/frontend/app/views/fragments/javaScriptFirstSteps.scala.html
+++ b/frontend/app/views/fragments/javaScriptFirstSteps.scala.html
@@ -13,7 +13,8 @@
 <!--[if IE 9]>
 <script src="@Asset.at("javascripts/lib/polyfills.min.js")"></script>
 <![endif]-->
-<script src="https://polyfill.guim.co.uk/v2/polyfill.min.js?features=Promise,fetch"></script>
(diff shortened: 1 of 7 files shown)


GHaberis:
@guardian/commercial-dev preparation PR for the Australian CMP.

rupertbates (approved):
👍  Good work picking all that apart!

frontend/app/views/fragments/javaScriptFirstSteps.scala.html
```diff
@@ -13,7 +13,7 @@
 <!--[if IE 9]>
 <script src="@Asset.at("javascripts/lib/polyfills.min.js")"></script>
 <![endif]-->
-<script src="https://polyfill.guim.co.uk/v2/polyfill.min.js?features=Promise,fetch"></script>
+<script src="https://polyfill.io/v3/polyfill.min.js?features=default,Array.prototype.find,fetch,Array.prototype.includes,Number.isInteger,Object.values,Object.entries"></script>
```
sndrs, line 16: dotcom uses https://assets.guim.co.uk/polyfill.io, don't know if you want to use that too?
```suggestion
<script src="https://assets.guim.co.uk/polyfill.io/v3/polyfill.min.js?features=default,Array.prototype.find,fetch,Array.prototype.includes,Number.isInteger,Object.values,Object.entries"></script>
```

GHaberis: Yeah I'm happy to do thus 👍  it's probably worth us changing this on support-frontend too so we're in sync with DCR/Frontend.

frontend/assets/javascripts/src/main.js
```diff
@@ -71,6 +71,10 @@ require([
             throw new Error('failed to get geocountry');
         }
     }).then(responseCountryCode => {
+        /**
+         * TODO: Enable Australian CMP by using
+         * property country instead of isInUsa
+        */
```
mxdvl, line 77: Is this going to happen in this PR, or a subsequent one?

mxdvl (approved):
Except from the fact the new CMP interface with the country code is not implemented, this looks good.

frontend/assets/javascripts/src/main.js
GHaberis: The plan is to do this in a follow-up PR. From the conversations we've had, and testing I've carried out updating the library but continuing to use the old `isInUsa` property causes no issues, so can be merged now.

PR #1981, opened 2020-11-16
Enable Australian CMP

GHaberis:
## Why are you doing this?

Follow up PR to https://github.com/guardian/membership-frontend/pull/1978 that enables the Australian CMP using new `country` property when calling `cmp.init()`


mxdvl (approved):
Great!

PR #1982, opened 2020-11-19
Honour consent state in Australia

GHaberis:
## Why are you doing this?

In https://github.com/guardian/membership-frontend/pull/1981 we enabled the new Australian CMP. The state returned when in Australian mode by the `@guardian/consent-management-platform` library `onConsentChange` now looks like:

```js
{
  aus: {
    personalisedAdvertising: true
  }
}
```

We therefore need to update the consent checks to handle `state` if it looks like this.


mxdvl (approved):
LGTM

PR #1983, opened 2020-11-25
Ensuring Identity ID is not passed to GA

paulbrown1982:
## Why are you doing this?
The Identity ID is no longer needed in GA and the Support site stopped sending Identity IDs to GA ages ago. 

As part of our data minimisation work, poor @NathanielBennett has been sending off repeated deletion requests to Google whilst this site has been adding new ones back in! Hopefully that will end now.


GHaberis (approved):
👍

PR #1986, opened 2020-12-15
[CMP] Add pubData for Sourcepoint Reporting

mxdvl:
## Why are you doing this?

In order to join consent signals with browsers, we need to send the page view id to Sourcepoint.

## Changes
* Imported `ophan.viewId`
* Added the `pubData` object to `cmp.init`
* Added `pageViewId` and `platform` params

### DIFF
--- a/frontend/assets/javascripts/src/main.js
+++ b/frontend/assets/javascripts/src/main.js
@@ -61,7 +61,7 @@ require([
     landingBundles,
     bundlesLanding,
     cmp,
-    ophan
+    ophan,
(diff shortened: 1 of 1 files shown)


ripecosta (approved):
👍  LGTM but might be best if one of the code owners leaves a review.

GHaberis (approved):
Looks good to me

mxdvl:
I don’t have merging rights, so if any of you is happy to merge, that would be 💯

PR #1988, opened 2021-02-04
More GA custom dimensions to remove

paulbrown1982:
## What does this change?

Personal identifiers and dimensions with infinite cardinality (Ophan page view ID, Ophan Browser ID, Membership Number etc) should not be being set as GA Custom Dimensions. This change removes them as they are not used in any GA reports and will soon be made inactive historical values will be deleted. See https://docs.google.com/spreadsheets/d/1MmWHNeeiQE_dzekImIP9Tv4beLx_8JzWx3rOtCp4PGg/edit#gid=1460779756

## How to test

1) Browse the site, no Ophan identifiers should be being set anymore.
1) Browse the checkout in test user mode, accepting tracking in the CMP. Make a test membership. On the thank-you page no more Membership Number will be sent.

## How can we measure success?

No more of these identifiers should be being sent to GA.

## Have we considered potential risks?

Risk of continuing to collect these custom dimensions is higher than impact on them no longer being in reports in GA. Risks are low as this site is mainly used by Customer Service representatives.

## Images

N/A

### DIFF
--- a/frontend/assets/javascripts/src/modules/analytics/ga.es6
+++ b/frontend/assets/javascripts/src/modules/analytics/ga.es6
@@ -6,14 +6,14 @@ const tracker = 'membershipPropertyTracker';
 const dimensions = {
     signedIn: 'dimension1', // User
     signedOut: 'dimension2', // User
-    ophanPageViewId: 'dimension3', // Hit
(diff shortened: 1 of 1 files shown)


GHaberis (approved):
Looks good

PR #1991, opened 2021-03-05
Patch in necessary permissions for wazuh

jorgeazevedo:
## What are you doing in this PR?

* Open ports 1514 and 1515 for outbound traffic
* Give permission to the instance to call describe-tags,
   so the agent can identify itself in a human readable way with 
   the wazuh server.

## Why are you doing this?

There's currently a major review of the guardian's security practices under the cyber security programme. One of the projects in this programme is to install a vulnerability scanner called Wazuh in guardian ec2 instances.

The scanner has already been installed in the ami image, but it's attempt to connect with the central server is being block by the instance security group. Hopefully this will fix it, otherwise I'll have to review network ACLs!

## Testing

Works in code ✅

### DIFF
--- a/cloud-formation/membership-app.cf.yaml
+++ b/cloud-formation/membership-app.cf.yaml
@@ -259,6 +259,7 @@ Resources:
   WazuhSecurityGroup:
     Type: AWS::EC2::SecurityGroup
     Properties:
+      GroupDescription: Allow outbound traffic from wazuh agent to manager
       VpcId:
         Ref: VpcId
       SecurityGroupEgress:
--- a/cloud-formation/membership-app.cf.yaml
+++ b/cloud-formation/membership-app.cf.yaml
@@ -93,6 +93,7 @@ Resources:
         Ref: ImageId
       SecurityGroups:
       - Ref: InstanceSecurityGroup
+      - Ref: WazuhSecurityGroup
       InstanceType: t4g.small
       IamInstanceProfile:
         Ref: MembershipAppInstanceProfile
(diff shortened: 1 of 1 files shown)


cloud-formation/membership-app.cf.yaml
```diff
@@ -93,6 +93,7 @@ Resources:
         Ref: ImageId
       SecurityGroups:
       - Ref: InstanceSecurityGroup
+      - Ref: WazuhSecurityGroup
```
johnduffell, line 96: weirdly even though you put them that way, they appear the other way in the instance list. not a major issue but I did notice when checking our reservations
image

cloud-formation/membership-app.cf.yaml
```diff
@@ -240,6 +256,17 @@ Resources:
...
+    Type: AWS::EC2::SecurityGroup
+    Properties:
+      GroupDescription: Allow outbound traffic from wazuh agent to manager
+      VpcId:
+        Ref: VpcId
+      SecurityGroupEgress:
+      - IpProtocol: tcp
+        FromPort: 1514
+        ToPort: 1515
+        CidrIp: 0.0.0.0/0
```
johnduffell, line 269: presumably there's no sensible limit on the IP addresses?

johnduffell (approved):
thanks for doing this 👍

cloud-formation/membership-app.cf.yaml
jorgeazevedo: Ah, that's weird, shall I flip them then?

cloud-formation/membership-app.cf.yaml
jorgeazevedo: Hm that's a good question. We have dns set up for the service so I'm not 100% sure, but even if it is static it's considered private information so infosec advised us against publishing it in public repos. In the amigo recipe for example, we fetch it from SSM.

cloud-formation/membership-app.cf.yaml
johnduffell: no it's not an issue, I just noticed in the list because of that

PR #1993, opened 2021-03-25
Use SSM tunnel for SSH access

i-hardy:
Co-authored with @rupertbates and @lucymonie 

## Why are you doing this?

This updates the Cloudformation template to remove SSH access via Github keys and switch to using SSM
https://github.com/guardian/ssm-scala#enabling-ssm-tunnel

### DIFF
--- a/cloud-formation/membership-app.cf.yaml
+++ b/cloud-formation/membership-app.cf.yaml
@@ -242,7 +242,7 @@ Resources:
   InstanceSecurityGroup:
     Type: AWS::EC2::SecurityGroup
     Properties:
-      GroupDescription: Enable HTTP access on the configured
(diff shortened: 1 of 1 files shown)


cloud-formation/membership-app.cf.yaml
```diff
@@ -233,7 +242,7 @@ Resources:
   InstanceSecurityGroup:
     Type: AWS::EC2::SecurityGroup
     Properties:
-      GroupDescription: Open up ssh access and enable HTTP access on the configured
+      GroupDescription: Enable HTTP access on the configured
```
rupertbates, line 245: Could revert this description change now that we're not removing the port 22 access?

rupertbates (approved):
👍 just one suggestion

PR #1996, opened 2021-04-09
Bump play, scala, sbt, plus misc fixes

johnduffell:
We need tog et off play 2.6 because it is unsupported (and has a vulnerability)

This PR bumps play as well as scala and sbt.


frontend/app/views/error500.scala.html
```diff
@@ -8,9 +8,10 @@
                     Sorry, there was an error on this page. We're aware of the problem. If this error persists, please contact
                     support on <a href="mailto:membershipsupport@@theguardian.com">membershipsupport@@theguardian.com</a>.
                 </p>
-                @ex match { case err: play.api.UsefulException =>
+                @ex match { case err: play.api.UsefulException => {
                     <p>Please provide the following error code to help identify the issue.</p>
                     <pre>Error code: @err.id</pre>
+                    }
```
johnduffell, line 14: this stopped compiling and this was the solution https://stackoverflow.com/questions/9748258/match-case-in-scala-template-doesnt-work-in-play2

frontend/test/acceptance/JoinSupporterSpec.scala
```diff
@@ -1,122 +0,0 @@
-package acceptance
```
johnduffell, line 1: no point in testing joining as that functionality was already removed!
Deleting the whole acceptance directory seemed to break the tests (they hang with a configuration issue) so I left it for now.

project/Dependencies.scala
```diff
@@ -22,7 +22,7 @@ object Dependencies {
   val scalaz = "org.scalaz" %% "scalaz-core" % "7.2.7"
   val selenium = "org.seleniumhq.selenium" % "selenium-java" % "3.14.0" % "test"
   val seleniumManager = "io.github.bonigarcia" % "webdrivermanager" % "2.2.5" % "test"
-  val specs2Extra = "org.specs2" %% "specs2-matcher-extra" % "3.8.6" % "test"
+  val specs2Extra = "org.specs2" %% "specs2-matcher-extra" % "4.5.1" % "test"
```
johnduffell, line 25: this has to match the version in play to avoid a org.something.Json class not found exception
https://mvnrepository.com/artifact/com.typesafe.play/play-specs2_2.12/2.7.9
image

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.3.13
+sbt.version=1.4.9
```
johnduffell, line 1: sbt 1.5.0 is the latest, but requires play 2.8.8 which was out yesterday.  Otherwise you get a class not found for NSC due to twirl compiler using an internal api of the scala compiler (sbt 1.5.0 uses scala 2.12.13)

tomrf1 (approved):
great!

PR #1999, opened 2021-04-20
remove acquisition event producer/bintray dependency

johnduffell:
## Why are you doing this?

This PR removes the dependency on bintray and acquisition event producer.

This is because bintray is going away, also we don't do acquisition via this route any more.

https://trello.com/c/accJjX0f/3665-bintray-migration-membership-frontend


rupertbates (approved):
👍

PR #2000, opened 2021-04-29
change to g200 copy

johnduffell:
This PR changes the copy as requested by the events team.

This can be reverted after the June 17th.

PR #2002, opened 2021-05-06
Sort out separate headers for masterclasses and events

rupertbates:
## Why are you doing this?
#2000 added special header wording for the 20th anniversary. 

Unfortunately masterclasses do not want that heading on their page as it doesn't really make sense in that context.

This PR reverts the masterclasses header to the original text and tweaks the wording slightly on the events page.

## Screenshots
Screen Shot 2021-05-06 at 15 01 53
Screen Shot 2021-05-06 at 15 01 34


johnduffell (approved):
👍

PR #2008, opened 2021-05-10
Remove the blaze server override as it is causing class not found errors

rupertbates:
## Why are you doing this?

#2005 Upgraded various dependencies, one of which was `http4s-blaze-server`. However this is causing `java.lang.NoClassDefFoundError` errors in production when running user authentication.


lucymonie (approved):
LGTM 👍🏻

PR #2012, opened 2021-06-21
Upgrade to Play 2.8

rupertbates:
## Why are you doing this?

Upgrade Play to 2.8.8

### DIFF
--- a/frontend/app/actions/CommonActions.scala
+++ b/frontend/app/actions/CommonActions.scala
@@ -121,9 +121,9 @@ abstract class OAuthActions(parser: BodyParser[AnyContent], implicit val executi
 
   import commonActions.NoCacheAction
   //routes
-  private val loginTarget = routes.OAuth.loginAction()
-  override val defaultRedirectTarget = routes.FrontPage.welcome()
-  override val failureRedirectTarget = routes.OAuth.login()
+  private val loginTarget = routes.OAuth.loginAction
+  override val defaultRedirectTarget = routes.FrontPage.welcome
+  override val failureRedirectTarget = routes.OAuth.login
 
 
   lazy val groupChecker = Config.googleGroupChecker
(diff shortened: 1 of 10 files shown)


frontend/app/actions/CommonActions.scala
```diff
@@ -121,9 +121,9 @@ abstract class OAuthActions(parser: BodyParser[AnyContent], implicit val executi
 
   import commonActions.NoCacheAction
   //routes
-  private val loginTarget = routes.OAuth.loginAction()
```
johnduffell, line 124: I think this change should be done the other way aaround - i.e. you woudl need to change it in the routes file instead. 
This is because it's using some eta expansion thing that's deprecated in scala 2.13 and will be removed in scala 3

johnduffell: I know this because I fixed them all in support-frontend, then noticed all the scala 2.3 warnings, so fixed them, which then broke all these again.

johnduffell: see https://github.com/guardian/support-frontend/pull/3122/files#r653618017

johnduffell (approved):
looks good although I'm not clear why the PlayLogger thing waas needed?
Also a suggestion as I think the reverse routes changes might cause eta expansion warnings in scala 2.13
other that that great work, we're ahead on play for once! 👏

rupertbates:
> looks good although I'm not clear why the PlayLogger thing waas needed?
> Also a suggestion as I think the reverse routes changes might cause eta expansion warnings in scala 2.13
> other that that great work, we're ahead on play for once! 👏

PlayLogger was needed because the static methods have disappeared off the Play Logger class. I could have just inlined it in the file I guess because it was only used there but when I made the change I thought it was more widely used.

johnduffell:
yeah see what you mean, maybe inline it if you're making other changes to this PR, but it's not really a biggie.

frontend/app/actions/CommonActions.scala
rupertbates: Thanks, I've updated as you suggest

rupertbates:
> yeah see what you mean, maybe inline it if you're making other changes to this PR, but it's not really a biggie.

Have inlined it.

PR #2013, opened 2021-06-21
Identity auth play upgrade

rupertbates:
## Why are you doing this?

There are a couple of high severity vulnerabilities which are fixed by this upgrade
Screen Shot 2021-06-21 at 14 08 24


project/Dependencies.scala
```diff
@@ -9,7 +9,7 @@ object Dependencies {
   //libraries
   val sentryRavenLogback = "io.sentry" % "sentry-logback" % "1.7.5"
   val scalaUri = "io.lemonlabs" %% "scala-uri" % "2.2.2"
-  val identityAuthPlay = "com.gu.identity" %% "identity-auth-play" % "3.239-C2"
+  val identityAuthPlay = "com.gu.identity" %% "identity-auth-play" % "3.248"
```
johnduffell, line 12: C2 sounds funny for the old one - maybe that was the play 2.7 backwards compatible version!

johnduffell (approved):
billing address is no longer stored in identity so we don't use that any more I assume
Looks good 👍 thanks

also I changed the base branch to make it easier to review, github is quite good at correcting it when you merge the first branch now. but feel free to change back.

project/Dependencies.scala
rupertbates: Yeah could be

PR #2019, opened 2021-07-05
only download the required file for deploy

johnduffell:
The deploys are failing because we do a recursive download.   However we only actually need a single file.
image

```
aws --region $region s3 cp --recursive s3://membership-dist/${stack}/${stage}/frontend/ /dist
fatal error: An error occurred (AccessDenied) when calling the ListObjectsV2 operation: Access Denied
```
So I have changed it so it only download the file it needs, meaning we don't need to extend the permissions.

### DIFF
--- a/frontend/instance/frontend/init-instance.sh
+++ b/frontend/instance/frontend/init-instance.sh
@@ -5,7 +5,7 @@
 CONF_DIR=/etc/frontend
 # download dist
 mkdir /dist
-aws --region $region s3 cp --recursive s3://membership-dist/${stack}/${stage}/frontend/ /dist
+aws --region $region s3 cp s3://membership-dist/${stack}/${stage}/frontend/frontend_1.0-SNAPSHOT_all.deb /dist
 # download private for this stage
 mkdir /etc/gu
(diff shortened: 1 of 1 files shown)


frontend/instance/frontend/init-instance.sh
```diff
@@ -5,7 +5,7 @@
 CONF_DIR=/etc/frontend
 # download dist
 mkdir /dist
-aws --region $region s3 cp --recursive s3://membership-dist/${stack}/${stage}/frontend/ /dist
+aws --region $region s3 cp s3://membership-dist/${stack}/${stage}/frontend/frontend_1.0-SNAPSHOT_all.deb /dist
```
rupertbates, line 8: This file is all we need is it?

rupertbates (approved):
👍 thanks for looking into this.

frontend/instance/frontend/init-instance.sh
johnduffell: the screenshot in the description is the whole contents of that directory in the bucket, so there can't be anything else (this .sh script is downloaded in the userdata)
if it fails, it won't be any worse than before, so let's see! 🤞

PR #2020, opened 2021-07-07
remove the G200 header and revert to requested copy

johnduffell:
This PR changes the previously requested events page header to the new requested header copy.
image


rupertbates (approved):
👍 Quite a lot of header text, but if that's what they want...

johnduffell:
thanks.  they did question whether there is a word limit, but there isn't from a technical point of view!

PR #2022, opened 2021-08-10
Update change password link in top pop-out menu

rebecca-thompson:
## Why are you doing this?
This updates the change password link in the top pop-out menu to point to the correct password reset page. The page the link is currently pointing at will shortly be removed to reduce duplication and ensure we are adhering to security best practice.


ripecosta (approved):
👍

johnduffell (approved):
thanks for updating this 👍

PR #2024, opened 2021-08-19
Snyk: Apply version resolution to lodash dependency

GHaberis:
## Why are you doing this?

This PR applies a Yarn version [resolution](https://classic.yarnpkg.com/en/docs/selective-version-resolutions/) for `lodash`, in response to a vulnerability flagged by Snyk. 

The resolution instructs that any dependency using `lodash` should use the version outlined in the resolution  "^4.17.21" . `emotion/core` was the dependency that imported the version of `lodash` with a vulnerability. This means the vulnerability should be addressed without having to upgrade `emotion/core`.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -79,5 +79,8 @@
     "webpack": "^4.35.0",
     "webpack-dev-server": "^3.7.2"
   },
+  "resolutions": {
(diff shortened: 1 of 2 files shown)


johnduffell (approved):
👍

PR #2029, opened 2021-11-18
Security updates including update to scala 2.13

johnduffell:
## Why are you doing this?

There were some vulns showing, this PR gets us on the latest membership-common and hence scala 2.13

Also bumped everything needed to close up all the vulns reported by snyk, that have a fixed version available.

I also removed most of the code that relates to Partners having a limited number of free tickets.  This is because the limit has not been applied since 2018, due to it being broken, and we are unlikely to fix it and it has been known about for at 2 years plus. https://github.com/guardian/membership-frontend/pull/1812#discussion_r413166002


frontend/app/actions/ActionRefiners.scala
```diff
@@ -33,7 +33,7 @@ object ActionRefiners {
 
   type SubReqWithPaid[A] = SubscriptionRequest[A] with PaidSubscriber
   type SubReqWithFree[A] = SubscriptionRequest[A] with FreeSubscriber
-  type SubReqWithSub[A] = SubscriptionRequest[A] with Subscriber
+  type SubReqWithSub[A] = SubscriptionRequest[A] with actions.Subscriber
```
johnduffell, line 36: scala 2.13 has changed the resolution rules from 2.12.
We have the same class name for two classes.  One is in the same package as this, one is in the controllers package which has a wildcard import.
Scala 2.12 preferred the local one, wheras scala 2.13 prefers the imported one.

In reality we should avoid this situation, but I preferred this solution to aliasing or renaming the class.

frontend/app/actions/ActionRefiners.scala
```diff
@@ -93,10 +93,10 @@ class ActionRefiners(authenticationService: AuthenticationService, parser: BodyP
 // this is helping us stack future/either/option
 object OptionEither {
 
-  type FutureEither[X] = EitherT[Future, String, X]
+  type FutureEither[X] = EitherT[String, Future, X]
```
johnduffell, line 96: scalaz has swapped the order of type parameters for EitherT

frontend/app/actions/ActionRefiners.scala
```diff
@@ -105,6 +105,6 @@ object OptionEither {
     apply(Future.successful(x.map[Option[A]](Some.apply)))
 
   def liftFutureEither[A](x: Option[A]): OptionT[FutureEither, A] =
-    apply(Future.successful(\/.right(x)))
+    apply(Future.successful(\/.r[String](x)))
```
johnduffell, line 108: scalaz is no longer covariant on `\/` so we have to make sure the right type parameter is used before passing into `apply`

frontend/app/actions/Fallbacks.scala
```diff
@@ -16,7 +16,7 @@ object Fallbacks {
     TemporaryRedirect(controllers.routes.Outages.maintenanceMessage.absoluteURL(secure=true))
 
   def memberHome(implicit request: RequestHeader) =
-    redirectTo(controllers.routes.FrontPage.welcome)
+    redirectTo(controllers.routes.FrontPage.welcome())
```
johnduffell, line 19: in scala 2.13 calling zero arg functions without using parenthasis is deprecated, so I added them in.

frontend/app/actions/TouchpointActionRefiners.scala
```diff
@@ -26,18 +26,6 @@ class TouchpointActionRefiners(authenticationService: AuthenticationService, tou
   implicit private val ec = executionContext
   implicit private val tpbs = touchpointBackends
 
-  private def getContributorRequest[A](request: AuthRequest[A]): Future[String \/ Option[SubReqWithContributor[A]]] = {
```
johnduffell, line 29: I removed a fair bit of code related to checking if a partner has used their free ticket allocation.  This is because it has been broken for a long time and is a known issue.

frontend/app/model/RichEvent.scala
```diff
@@ -56,7 +56,7 @@ object RichEvent {
 
   def getCitiesWithCount(events: Seq[RichEvent]): Seq[(String, Int)] = {
     val cities = events.flatMap(_.underlying.ebEvent.venue.address.flatMap(_.city))
-    cities.groupBy(identity).mapValues(_.size).toSeq.sortBy{ case (name, size) => name }
+    cities.groupBy(identity).view.mapValues(_.size).toSeq.sortBy{ case (name, size) => name }
```
johnduffell, line 59: it's deprecated in scala 2.13 to use mapValues without doing .view first.

frontend/app/monitoring/MemberMetrics.scala
```diff
@@ -12,35 +14,35 @@ class MemberMetrics(val backendEnv: String) extends TouchpointBackendMetrics {
     put(s"attempted-sign-ups-${tier.name}")
   }
 
-  def putSignUp(tier: Tier) {
+  def putSignUp(tier: Tier) = {
```
johnduffell, line 17: "procedure syntax" has been deprecated for a while

frontend/app/monitoring/SentryLogging.scala
```diff
@@ -13,7 +13,7 @@ import io.sentry.logback.SentryAppender
 import org.slf4j.Logger.ROOT_LOGGER_NAME
 import org.slf4j.LoggerFactory
 
-import scala.collection.JavaConverters._
+import scala.jdk.CollectionConverters._
```
johnduffell, line 16: this is the new way to do asJava on things

frontend/app/services/GuardianContentService.scala
```diff
@@ -1,21 +1,18 @@
...
 import com.gu.contentapi.client.GuardianContentClient
+import com.gu.contentapi.client.model.v1._
+import com.gu.contentapi.client.model.{ContentApiError, ItemQuery, SearchQuery}
 import com.gu.memsub.util.ScheduledTask
-import configuration.Config
-import org.joda.time.DateTime
 import com.gu.monitoring.SafeLogger
 import com.gu.monitoring.SafeLogger._
-import play.api.libs.iteratee.{Enumerator, Iteratee}
+import configuration.Config
```
johnduffell, line 10: https://github.com/playframework/play-iteratees
is end of life and has no scala 2.13 release

I have rewritten the code in this file that uses it in order to use recursion instead.  This is a recursive flatMap so it's not tail recursive, but it should be ok for now.  I would be possible to make it tail recursive I'm sure, but I have't done so.

frontend/app/services/GuardianContentService.scala
```diff
@@ -66,13 +59,13 @@ class GuardianContentService(actorSystem: ActorSystem, executionContext: Executi
   private val contentApiPeriod = 30.minutes
 
   val masterclassContentTask = ScheduledTask[Seq[MasterclassData]](
-    "GuardianContentService - Masterclass content", Nil, 2.seconds, contentApiPeriod)(masterclasses)
+    "GuardianContentService - Masterclass content", Nil, 2.seconds, contentApiPeriod)(masterclasses())
 
   val contentTask = ScheduledTask[Seq[Content]](
-    "GuardianContentService - Content with Eventbrite reference", Nil, 1.millis, contentApiPeriod)(eventbrite)
+    "GuardianContentService - Content with Eventbrite reference", Nil, 1.millis, contentApiPeriod)(eventbrite())
```
johnduffell, line 65: these are now calling the new recursive function with default args, rather than the old enumerator.  This is still going to be reevaluated on every call, as it should (otherwise the events list would only update on deploy!)

frontend/app/services/IdentityService.scala
```diff
@@ -162,7 +162,7 @@ class IdentityApi(val wsClient: WSClient, implicit val ec: ExecutionContext) {
   def getUserPasswordExists(headers:List[(String, String)], parameters: List[(String, String)]) : Future[Boolean] = {
     val endpoint = "user/password-exists"
     val url = s"${Config.idApiUrl}/$endpoint"
-    wsClient.url(url).withHttpHeaders(headers: _*).withQueryStringParameters(parameters: _*).withRequestTimeout(1000 milli).get().map { response =>
+    wsClient.url(url).withHttpHeaders(headers: _*).withQueryStringParameters(parameters: _*).withRequestTimeout(1.second).get().map { response =>
```
johnduffell, line 165: the postfix notation needs a language import adding, so I changed to the slightly more conventional way.  Also I decided 1 second is clearer than 1000 ms

frontend/app/services/MasterclassDataExtractor.scala
```diff
@@ -15,7 +15,7 @@ object MasterclassDataExtractor {
 
   def extractEventbriteInformation(content: Content): Seq[MasterclassData] = {
     val eventbriteIdsFromRefs =
-      content.references.filter(_.`type` == "eventbrite").map(_.id.stripPrefix("eventbrite/"))
+      content.references.toSeq.filter(_.`type` == "eventbrite").map(_.id.stripPrefix("eventbrite/"))
 
     val eventbriteIds = (eventbriteIdsFromRefs ++ scrapeEventbriteIdsFrom(content)).distinct
```
johnduffell, line 21: the MembersService diff below is collapsed due to being "large" but actually it's An important change, I deleted most of the code as it was unused in the end.

frontend/app/services/TouchpointBackend.scala
```diff
@@ -85,21 +85,7 @@ object TouchpointBackend {
...
-      auStripeService = stripeAUMembershipService,
-      subscriptionService = newSubsService,
-      catalogService = newCatalogService,
-      paymentService = paymentService,
-      discounter = discounter,
-      discountIds = Config.discountRatePlanIds(config.zuoraEnvName),
-      invoiceIdsByCountry = Config.invoiceTemplateOverrides(config.zuoraEnvName),
-      ec = executionContext
-    )
+    val memberService = new MemberService()(executionContext)
```
johnduffell, line 88: member service (above file) has been simplified to the point of not really having much code of use, therefore almost no dependencies.

frontend/build.sbt
```diff
@@ -24,22 +24,22 @@ organization := "com.gu"
...
-scalaVersion := "2.12.13"
+scalaVersion := "2.13.7"
 
 resolvers ++= Seq(
     "Guardian Github Releases" at "https://guardian.github.io/maven/repo-releases",
     Resolver.sonatypeRepo("releases")
 )
 
-sources in (Compile,doc) := Seq.empty
+Compile / doc / sources := Seq.empty
```
johnduffell, line 34: `in Test` style syntax is loudly deprecated now in favour of the `Test / ` syntax

frontend/build.sbt
```diff
@@ -111,4 +111,6 @@ libraryDependencies ++= frontendDependencies.map {
      .exclude("org.slf4j", "slf4j-simple") // snatches SLF4J logging from Logback, our chosen logging system
 }
 
+dependencyOverrides += "com.fasterxml.jackson.core" % "jackson-databind" % "2.11.4"
```
johnduffell, line 114: this is needed to prevent it using 2.12 which causes play's akka to fall over at runtime objecting that it needs 2.11.x

frontend/conf/routes
```diff
@@ -61,18 +61,17 @@ GET         /masterclasses/:tag/:subTag                controllers.WhatsOn.maste
 GET         /event/:id                                 controllers.Event.details(id)
 GET         /event/:id/embed                           controllers.Event.embedData(id)
 GET         /event/:id/buy                             controllers.Event.buy(id)
-GET         /event/:id/thankyou                        controllers.Event.thankyou(id, oid: Option[String])
```
johnduffell, line 64: this is no longer used, I grepped 5 days of fastly logs and only bots had hit it

frontend/test/services/GridServiceTest.scala
```diff
@@ -28,7 +28,7 @@ class GridServiceTest(implicit ev: ExecutionEnv) extends Specification {
       val exports = gridResponse.data.exports.get
       val requestedCrop = "0_130_1703_1022"
       val gridService = new GridService(ev.executionContext)
-      val export = gridService.findExport(exports, requestedCrop)
+      val `export` = gridService.findExport(`exports`, requestedCrop)
```
johnduffell, line 31: export and exports is a warning as it is a reserved word in scala 3

frontend/app/actions/ActionRefiners.scala
tomrf1: how helpful 😆 
so this is now the opposite of cats' `EitherT`

frontend/app/services/GuardianContentService.scala
```diff
@@ -26,37 +23,33 @@ class GuardianContentService(actorSystem: ActorSystem, executionContext: Executi
...
-        } yield {
-          val pagination = ContentAPIPagination(response.currentPage, response.pages)
-          Some(pagination.nextPageOpt, response.results)
-
+  private def eventbrite(nextPage: Int = 1, accumulatedEventbriteData: Seq[Content] = Nil): Future[Seq[Content]] = {
+    for {
+      response <- eventbriteQuery(nextPage)
+      rest <- {
+        val pagination = ContentAPIPagination(response.currentPage, response.pages)
+        val combined = response.results.toSeq.reverse ++ accumulatedEventbriteData // linked list - prepend the smaller amount
```
tomrf1, line 31: would Vector be a better choice, to avoid all this reversing? E.g.

```
  private def eventbrite(nextPage: Int = 1): Future[Vector[Content]] = {
    for {
      response <- eventbriteQuery(nextPage)
      content <- {
        val pagination = ContentAPIPagination(response.currentPage, response.pages)
        val pageResults = response.results.toVector
        pagination.nextPageOpt match {
          case Some(nextPage) => eventbrite(nextPage).map(nextResults => pageResults ++ nextResults)
          case None => Future.successful(pageResults)
        }
      }
    } yield content
  }
```

tomrf1 (approved):
Great work! 👍 

Just one comment about the `GuardianContentService.scala` logic

frontend/app/services/GuardianContentService.scala
johnduffell: good point, it wasn't tail recursive anyway, due to the flatMap, and Vector simplifies the logic.  I have updated this and the one below according to your suggestion and it works ok locally.

PR #2030, opened 2021-12-13
update to version 1.5.6 of SBT

johnduffell:
## Why are you doing this?

we want to move to 1.5.6 just in case of the log4j vulnerability


TomPretty (approved):
👍

PR #2111, opened 2022-05-10
Update identity-auth-play to 3.254

rupertbates:
## Why are you doing this?

Upgrading identity-auth-play to mitigate a critical vulnerability. More details here https://github.com/guardian/identity/pull/2106


kelvin-chappell (approved):
👍

PR #2120, opened 2022-05-20
minimist / dependency updates

GHaberis:
## Why are you doing this?

Aims to fix this Dependabot alert https://github.com/guardian/membership-frontend/security/dependabot/90 . I've added a resolution as `minimist` is a transitive dependency, and can't be updated through updates to our dependencies. This forces any reference to `minimist` to resolve to `^1.2.6`.

I've tested the build (`yarn compile`) and validation (`yarn test`) steps and everything is working as expected, and I've tested the site locally and all is OK.

A couple of other changes include:
- Moved `load-grunt-tasks` and `time-grunt` to `devDependencies`.
- Removed the unused dependency `mkdirp`.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -30,9 +30,7 @@
     "gumshoejs": "^3.5.0",
     "jquery": "^3.5.1",
     "lazysizes": "^5.2.2",
-    "load-grunt-tasks": "^4.0.0",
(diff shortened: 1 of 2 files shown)


michaelbjacobson (approved):
Looks good to me 👍🏼

PR #2312, opened 2022-10-26
Remove misleading ticket info from Events pages

GHaberis:
## Why are you doing this?

The request was to remove the following sections from Guardian Live event pages. The templates that contain these elements are shared with Masterclass event pages, and these sections need to remain on those pages. I've therefore wrapped the elements in conditional checks `event.metadata.identifier == "masterclasses"`

Screenshot 2022-10-26 at 15 16 54

I've also made an additional change to only render the location map element if an address or venue name is available. Almost all events are virtual now so these are often `None` which leads a useless element being rendered...

Screenshot 2022-10-25 at 17 13 05

[Trello](https://trello.com/c/sB1KQPPA/756-remove-confusing-copy-from-events-pages)

### Desktop: Before

Screenshot 2022-10-26 at 17 47 07

### Desktop: After

Screenshot 2022-10-26 at 16 03 58

### Tablet: Before

Screenshot 2022-10-26 at 17 46 26

### Tablet: After

Screenshot 2022-10-26 at 17 45 48

### Mobile: Before

Screenshot 2022-10-26 at 17 46 53

### Mobile: After

Screenshot 2022-10-26 at 16 04 29

### DIFF
--- a/frontend/app/views/fragments/event/stats.scala.html
+++ b/frontend/app/views/fragments/event/stats.scala.html
@@ -15,7 +15,7 @@
 </div>
 
 @* Event locations *@
-@if(event.underlying.ebEvent.venue.name && event.underlying.ebEvent.venue.address) {
+@if(event.underlying.ebEvent.venue.addressDefaultLine.isDefined || event.underlying.ebEvent.venue.googleMapsLink.isDefined) {
     <div class="stat-item">
         <div class="stat-item__first">
         @fragments.inlineIcon("location", List("icon-inline--medium", "icon-inline--top", "icon-inline--neutral"))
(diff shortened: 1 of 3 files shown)


frontend/app/views/fragments/event/stats.scala.html
```diff
@@ -15,24 +15,26 @@
 </div>
 
 @* Event locations *@
-<div class="stat-item">
-    <div class="stat-item__first">
+@if(event.underlying.ebEvent.venue.name && event.underlying.ebEvent.venue.address) {
```
rupertbates, line 18: I didn't know that Twirl templates had truthiness!

GHaberis: How would you have typically made this check for not `None`? I tried this and it worked, but happy to go with a more idiomatic approach.

GHaberis: I've actually updated this conditional to use `isDefined` and I've made it an `||` rather than an `&&` as if either of these properties are defined we can render this location information.

PR #2323, opened 2022-11-04
Updated the event schema markup

LAKSHMIRPILLAI:
## Why are you doing this?

To update event schema markup to event pages on the membership site, to enable Guardian events to show in the Google event carousel listings and dramatically increase organic traffic to our live events.

## Changes
* Added url as a new property under LocationSchema to add “[VirtualLocation](https://schema.org/VirtualLocation)” as a Location Type
* Changes to locationOpt method to check for the locationType and pass values to Schema accordingly .Also modified it to change the url links for “IsSoldOut” cases as well.If events are  “Sold Out” , the users must be able to see the waiting list url
* Added eventStatus and eventAttendanceMode as  new properties to  Event Schema.

### DIFF
--- a/frontend/app/model/Schema.scala
+++ b/frontend/app/model/Schema.scala
@@ -54,16 +54,14 @@ object EventSchema {
   }
 
   private def locationOpt(event: RichEvent): Option[LocationSchema] = {
-    if (event.underlying.ebEvent.venue.name.isEmpty) {
-      if (!event.underlying.isSoldOut)
-        Some(LocationSchema(None,Some(event.underlying.ebEvent.memUrl),None,None,"VirtualLocation"))
-      else
-        Some(LocationSchema(None, Some(Config.eventbriteWaitlistUrl(event.underlying.ebEvent)), None, None,"VirtualLocation"))
+    event.underlying.ebEvent.venue.name match {
+      case None => if (!event.underlying.isSoldOut)
+          Some(LocationSchema(None, Some(event.underlying.ebEvent.memUrl), None, None, "VirtualLocation"))
+        else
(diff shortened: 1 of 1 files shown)


frontend/app/model/Schema.scala
```diff
@@ -2,52 +2,68 @@ package model
...
+    if (event.underlying.ebEvent.venue.name.isEmpty) {
+      if (!event.underlying.isSoldOut)
+        Some(LocationSchema(None,Some(event.underlying.ebEvent.memUrl),None,None,"VirtualLocation"))
+      else
+        Some(LocationSchema(None, Some(Config.eventbriteWaitlistUrl(event.underlying.ebEvent)), None, None,"VirtualLocation"))
     }
+    else
+      event.underlying.ebEvent.venue.name.map { name =>
+        LocationSchema(Some(name),None, event.underlying.ebEvent.venue.addressLine, event.underlying.ebEvent.venue.googleMapsLink,"Place")
+      }
```
rupertbates, line 66:
```suggestion
    event.underlying.ebEvent.venue.name match {
      case None => if (!event.underlying.isSoldOut)
          Some(LocationSchema(None, Some(event.underlying.ebEvent.memUrl), None, None, "VirtualLocation"))
        else
          Some(LocationSchema(None, Some(Config.eventbriteWaitlistUrl(event.underlying.ebEvent)), None, None, "VirtualLocation"))
      case Some(name) =>
        Some(LocationSchema(Some(name), None, event.underlying.ebEvent.venue.addressLine, event.underlying.ebEvent.venue.googleMapsLink, "Place"))
    }
```
This is a more standard way to write this code in Scala, it means we get the `name` parameter where it exists without having to map over it again.

rupertbates (approved):
👍 looks good, just one suggestion about using a match statement instead of an if

PR #2341, opened 2022-11-18
Event schema updated for organizer,validFrom and breadcrumbs

LAKSHMIRPILLAI:
## Why are you doing this?
To update event schema markup on the membership site  to add  BreadcrumbList and fix non-critical issues  like ‘validFrom’ and 'organizer' in order to optimise it for events to appear in Google search results

### DIFF
--- a/frontend/app/model/Schema.scala
+++ b/frontend/app/model/Schema.scala
@@ -25,8 +25,17 @@ case class OfferSchema(
  category: String,
  price: String,
  priceCurrency: String,
+ validFrom:String,
  availability: Option[String],
  `@type`: String = "Offer")
+object OrganizerSchema {
+  implicit val writesSchema = Json.writes[OrganizerSchema]
+}
+
+case class OrganizerSchema(
+                        url: String,
+                        name: String,
+                        `@type`: String = "Organization")
 
 case class EventSchema(
  name: String,
@@ -39,6 +48,7 @@ case class EventSchema(
  eventStatus:String,
  location: Option[LocationSchema],
  offers: Option[OfferSchema],
+ organizer:Option[OrganizerSchema],
  `@context`: String = "http://schema.org",
  `@type`: String = "Event")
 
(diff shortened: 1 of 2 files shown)


frontend/app/model/Schema.scala
```diff
@@ -25,8 +25,17 @@ case class OfferSchema(
...
  price: String,
  priceCurrency: String,
+ validFrom:String,
  availability: Option[String],
  `@type`: String = "Offer")
+object OrganizerSchema {
+  implicit val writesSchema = Json.writes[OrganizerSchema]
+}
+
+case class OrganizerSchema(
```
GHaberis, line 35: The spacing/formatting seems a little off compared to the other case classes in this file, eg. the props of the class are overly indented. It'd be better for readability if it was the same  as the rest of the module.

frontend/app/model/Schema.scala
```diff
@@ -39,6 +48,7 @@ case class EventSchema(
  eventStatus:String,
  location: Option[LocationSchema],
  offers: Option[OfferSchema],
+ organizer:Option[OrganizerSchema],
```
GHaberis, line 51: Just another formatting/readability point, but there should be a space between the `organizer:` and `Option[OrganizerSchema]` on this line.

GHaberis: I'd also add a space in the same way after the `eventAttendanceMode` and `eventStatus` keys.

frontend/app/model/Schema.scala
```diff
@@ -25,8 +25,17 @@ case class OfferSchema(
  category: String,
  price: String,
  priceCurrency: String,
+ validFrom:String,
```
GHaberis, line 28: I'd add a space between `validFrom:` and `String` on this line, for readability/formatting consistency.

frontend/app/views/main.scala.html
```diff
@@ -67,10 +67,38 @@
...
+                "id":"https://membership.theguardian.com/events"
+                }]},{
+                "@@type": "ListItem",
+                "position": 3,
+                "name":"@Html(schema.name)",
+                "item":[
                 @Html(Json.toJson(schema).toString)
+                ]
+                }]
+                }
```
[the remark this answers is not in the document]
LAKSHMIRPILLAI: It seems "https://membership.theguardian.com/masterclasses" is only for the  Masterclass pages where all ,masterclass events are listed.If we access each individual event , the url changes to https://membership.theguardian.com/event/ which would be the same for events as well as masterclasses .So I have taken out the 2nd breadcrumb

frontend/app/model/Schema.scala
LAKSHMIRPILLAI: Reformatted for readability

GHaberis (approved):
Looks good 👍

PR #2351, opened 2022-11-30
Modify breadcrumb list of event schema 

LAKSHMIRPILLAI:
## Why are you doing this?

The event schema markup on the membership site to be updated with 'id' in  BreadcrumbList  in order to optimise it for events to appear in Google search results.The breadcrumb list to be updated similar to what we have in pages of https://patrons.theguardian.com/events/ 

## Changes
* Taking events out of the Breadcrumb list
* Adding id to Breadcrumb list for individual events and masterclasses

After the change (Breadcrumbs from Membership site)
image

After the change (Events from Membership site) 
image

Breadcrumbs from Patrons (for example)
image

### DIFF
--- a/frontend/app/views/main.scala.html
+++ b/frontend/app/views/main.scala.html
@@ -83,11 +83,15 @@
                 "@@type": "ListItem",
                 "position": 2,
                 "name":"@Html(schema.name)",
-                "item":[
-                @Html(Json.toJson(schema).toString)
-                ]
+                "item":[{
+                "@@type": "WebPage",
+                "id":"@Html(schema.url)"
+                }]
                 }]
                 }
+            </script>
+            <script type="application/ld+json">
+            @Html(Json.toJson(schema).toString)
             </script>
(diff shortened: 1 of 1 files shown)


frontend/app/views/main.scala.html
```diff
@@ -83,12 +83,16 @@
...
-                "item":[
-                @Html(Json.toJson(schema).toString)
-                ]
+                "item":[{
+                "@@type": "WebPage",
+                "id":"@Html(schema.url)"
+                }]
                 }]
                 }
             </script>
```
GHaberis, line 92: One comment, it'd help with readability if the indentation for the json were consistent, eg. The opening `{` on line 70 is indented differently to the closing `}` on line 91.

LAKSHMIRPILLAI: Adjusted indentation for the json

GHaberis (approved):
👍

PR #2384, opened 2023-01-20
Routing change to force the correct path prefix for Event Details page

paulbrown1982:
## Why are you doing this?
Making a small routing change to force the correct path prefix (/masterclass/ or /live/) when showing an Event details page. This makes it much easier for a human to identify that pageview counts are for masterclass or live event traffic at a glance in Google Analytics etc

Before: 

A Masterclass URL would look like:

https://events.theguardian.com/event/kickstart-your-writing-year-an-online-retreat-with-guardian-masterclasses-474635275887 

Now, the above path redirects to:

https://events.theguardian.com/masterclass/kickstart-your-writing-year-an-online-retreat-with-guardian-masterclasses-474635275887

This should improve the Behaviour > Site Content listing in Google Analytics, which often shows a confusing mix of Masterclass and Live Event pages:

Screenshot 2023-01-20 at 09 35 24

This change will let the team setup partitioned views.

## Trello card: [Here](https://trello.com)
https://trello.com/c/6Bt9vjVh

## Testing

- [x] Tested on CODE  (https://membership.code.dev-theguardian.com/)

cc @GHaberis

### DIFF
--- a/frontend/app/controllers/Event.scala
+++ b/frontend/app/controllers/Event.scala
@@ -59,14 +59,51 @@ class Event(
   private def BuyAction(id: String, onUnauthenticated: RequestHeader => Result) = NoCacheAction andThen recordBuyIntention(id) andThen
     authenticated(onUnauthenticated = onUnauthenticated) andThen subscriptionRefiner(onNonMember = onUnauthenticated)
 
+  def detailsMasterclass(slug: String) = CachedAction { implicit request =>
+    detailsResult(slug)((event: RichEvent) => {
+      if (event.isInstanceOf[MasterclassEvent]) {
+        eventDetail(event)
+      } else {
+        // the url prefix or slug are not canonical so 302 to the correct one
+        Redirect(event.detailsUrl)
+      }
+    });
+  }
+
+  def detailsLive(slug: String) = CachedAction { implicit request =>
+    detailsResult(slug)((event: RichEvent) => {
+      if (event.isInstanceOf[LiveEvent]) {
+        eventDetail(event)
+      } else {
+        // the url prefix or slug are not canonical so 302 to the correct one
+        Redirect(event.detailsUrl)
+      }
+    });
+  }
+
   def details(slug: String) = CachedAction { implicit request =>
+    detailsResult(slug)((event: RichEvent) => {
+      if (event.isInstanceOf[LiveEvent] || event.isInstanceOf[MasterclassEvent]) {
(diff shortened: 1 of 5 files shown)


frontend/app/views/fragments/event/item.scala.html
```diff
@@ -34,7 +34,7 @@
     </div>
 }
 
-<a href="@routes.Event.details(event.underlying.ebEvent.slug).url" class="@RenderClasses(Map(
+<a href="@event.detailsUrl" class="@RenderClasses(Map(
```
paulbrown1982, line 37: This should have been using the property from the class anyway, rather than distributing the logic to do a slug-based lookup.

frontend/app/views/fragments/event/itemList.scala.html
```diff
@@ -1,6 +1,6 @@
 @(event: model.RichEvent.RichEvent, classes: Seq[String] = Seq.empty)
 
-<a href="@routes.Event.details(event.underlying.ebEvent.slug)"
+<a href="@event.detailsUrl"
```
paulbrown1982, line 3: This should have been using the property from the class anyway, rather than distributing the logic to do a slug-based lookup.

frontend/conf/routes
```diff
@@ -59,6 +59,8 @@ GET         /masterclasses/:tag/:subTag                controllers.WhatsOn.maste
 
 # Event Details
 GET         /event/:id                                 controllers.Event.details(id)
```
paulbrown1982, line 61: I left the old /event/ URL alone, but changed the logic in the method to redirect if it determines the event is definitely a Live Event or Masterclass - I don't think there is likely to be any non-specific RichEvents, but if there is this is the path to render them.

frontend/app/controllers/Event.scala
```diff
@@ -59,14 +59,51 @@ class Event(
...
+      } else {
+        // the url prefix or slug are not canonical so 302 to the correct one
+        Redirect(event.detailsUrl)
+      }
+    });
+  }
+
   def details(slug: String) = CachedAction { implicit request =>
+    detailsResult(slug)((event: RichEvent) => {
+      if (event.isInstanceOf[LiveEvent] || event.isInstanceOf[MasterclassEvent]) {
```
paulbrown1982, line 86: This is the logic to decide whether to redirect `/event/` to `/masterclass/` or `/live/`.

rupertbates (approved):
Looks good 👍

PR #2485, opened 2023-05-31
Update dependencies to remove ansi-regex vulnerability

undergroundquizscene:
- Trello card: https://trello.com/c/gB7yeZG1/1284-membership-frontend-ansi-regex-snyk-high

Snyk identified a transitive dependency on ansi-regex@2.1.1 as a high-severity vulnerability, introduced through time-grunt and autoprefixer-core.

This PR removes time-grunt as [there’s no newer version](https://www.npmjs.com/package/time-grunt#deprecated), and doesn’t add a replacement as it doesn’t seem super important to have execution times listed.

It also updates autoprefixer to the newest version that works (replacing autoprefixer-core, which [merged with it in version 6.0.1](https://github.com/ai/autoprefixer-core#autoprefixer-core-was-deprecated-useautoprefixerinstead)).

I tried the newest version (10.4.14) of autoprefixer, but it depends on postcss version 8, which I don’t think is supported by grunt-postcss: the newest version of grunt-postcss is still 0.9.0, which is now 6 years old.

It looks like the time-grunt maintainer is correct that grunt is not really maintained anymore and that we will need to move away from it.

There are also some formatting changes in a second commit, as well as an update to the browserslist and an .nvmrc: more detail in the messages on those commits.


undergroundquizscene:
[membership-frontend: ansi-regex (Snyk: HIGH)](https://trello.com/c/gB7yeZG1/1284-membership-frontend-ansi-regex-snyk-high)

LAKSHMIRPILLAI (approved):

PR #2489, opened 2023-06-22
Migrate membership-frontend from TeamCity to Github Actions.

LAKSHMIRPILLAI:
## Why are you doing this?

Migrate [membership-frontend](https://teamcity.gutools.co.uk/buildConfiguration/memsub_membership_frontend) from TeamCity to Github Actions.

## Trello card: [Here](https://trello.com/c/hgzNr4Kx/1399-teamcity-migration-membership-frontend)

DevX is planning to retire TeamCity by the end of September 2023. All workflows will need to be migrated to Github Actions

The TeamCity agent count will be continually decreasing from June onwards, reaching 0 by the end of September. So the build queues and therefore the time it takes to be able to deploy might increase, if the job count does not decrease.
‌
You should not use TeamCity for any new applications from now on.
Our main build service is now GitHub Actions. [GitHub Actions documentation - GitHub Docs](https://docs.github.com/en/actions) 

We will all benefit from this move by being able to keep our CI builds as code, version controlled alongside its codebase, and having a reduced number of tools and access, all bundled under one vendor, with better analytics through CloudQuery and Grafana.
‌
Furthermore, this change will help improve our resilience to [supply chain threats](https://slsa.dev/spec/v1.0/threats-overview) particularly in the build process.
‌
More detailed reasoning can be found in this proposal [Using GitHub Actions for CI](https://docs.google.com/document/d/13AoZwSpK_Dy_KFcORoPhiiLNuZZYy5MW0PBt6vNUSeM)


.github/workflows/membership-frontend-build.yml
```diff
@@ -0,0 +1,112 @@
...
+                fi
+              working-directory: frontend
+
+            - name: Install Dependencies
+              run: |
+                  export YARN_VERSION="1.12.3"
+                  export YARN_LOCATION="tools/${YARN_VERSION}"
+                  ${YARN_LOCATION}/yarn-v${YARN_VERSION}/bin/yarn cache clean
+                  ${YARN_LOCATION}/yarn-v${YARN_VERSION}/bin/yarn install
+              working-directory: frontend
```
undergroundquizscene, line 64: Yarn v1.22.19 [is installed by default on the runners](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md). Could we check whether this works with that version of yarn? If it does, we can also remove some of this code.
```suggestion
            - name: Install Dependencies
              run: |
                  yarn cache clean
                  yarn install
              working-directory: frontend
```

.github/workflows/membership-frontend-build.yml
```diff
@@ -0,0 +1,112 @@
...
+                  key: sbt
+
+                # ---- Build and Test membership-frontend---- #
+            - name: Test - membership-frontend
+              run: |
+                  sbt clean test riffRaffNotifyTeamcity
+
+            - name:  Run script/ci.sh
+              run: ./script/ci.sh
+
```
undergroundquizscene, line 99: This script doesn’t appear to do anything anymore – could we remove it?
```suggestion
```

undergroundquizscene (commented):
Have you deployed this to CODE and verified the frontend comes up ok?

.github/workflows/membership-frontend-build.yml
LAKSHMIRPILLAI: This worked with the  default yarn version on the runners.So have changed. as per the suggestion

undergroundquizscene (approved):

PR #2511, opened 2023-07-21
Remove Football Weekly Live tour events

paulbrown1982:
## Why are you doing this?
The Senior producer of Guardian Live has asked that the Football Weekly Live tour events do not show on the site. I can't see a configurable way to remove them other than hardcoding in this way.


GHaberis (approved):
Looks good

PR #2531, opened 2023-11-13
Reduce the number of Eventbrite API calls by slowing the refresh schedule for the archive

paulbrown1982:
## Why are you doing this?
Reduce the number of Eventbrite API calls by slowing the refresh schedule for the archive. This will hopefully give space to let other ETL jobs complete.


tomrf1 (approved):

johnduffell (approved):
sounds good, there must be so many archived events now.  Perhaps another option is to reduce how far back it goes.

